PR #1, opened 2015-04-13
Replace all 'url' with 'URL'.

lexrus:
We'd better take care about case sensitive.


onevcat:
Thanks for it. But I reverted the pod project file. It seems you are using an earlier version of CocoaPods, which changes the pod file.

PR #4, opened 2015-04-14
Fix some typo

mrahmiao:
Fixed some typo.


onevcat:
Thanks for it. It would be great if Xcode has spelling checking. :(

PR #11, opened 2015-04-19
Add not nil judgment for placeholderImage before set it to UIImageView

TonnyTao:
In some case, we don’t want set image = nil for UIImageView before downloading image. We just want to keep old image for it.


onevcat:
Hi, thanks for the p-r.

But if you always check the nil before set, how about if you actually want to set the image to nil? If you need to keep the original image, pass the imageView.image as the placeholder image will be OK.

PR #39, opened 2015-04-29
Add cacheType param to CompletionHandler

yonaskolb:
This adds a cacheType:CacheType parameter to the CompletionHandler. It also adds a .None case to the CacheType enum.
This allows for checking where the image came from, to for example, decide whether to animate a transition to this image (ie, if the cacheType == .None then fade in image as it's just been download, otherwise do nothing)

This also aligns with the SDWebImage completion handler, making switching frameworks easier


onevcat:
Hi, @yonaskolb 

Wow, great work. Thank you very much for the p-r.

LGTM, but could I request some modification before it could be merged?

First, currently `ImageDownloaderCompletionHandler` is a typealias of `CompletionHandler`, which is not necessary in fact. Downloader should know nothing about the cache (and the cacheType as well), so I suggest to refactor the `ImageDownloaderCompletionHandler` to remain without cacheType. The `cacheType` should be only returned from manager or cache class. By doing so, we can ensure the decoupling of downloader and cache.

And would you please to issue another p-r with this to branch 1.2.1 instead of master?

Thank you!

yonaskolb:
Ok, I've separated out the ImageDownloadCompletionHandler without the cacheType, and made a PR to 1.2.1

onevcat:
Merged. Thanks!

PR #42, opened 2015-04-30
[WIP] 1.3.0

onevcat:
Since there is API modification. I think it should be 1.3.0 instead of 1.2.1.

- [x] Add handling for status code 304
- [x] Add cacheType in completion handler.
- [x] Documentation
- [x] Wiki for how to handle 304

---

- [x] Update pod spec


alexdd55:
let me just mention, that this is great work. thank you. :)

PR #68, opened 2015-06-19
Add queue resulting queue to KingfisherManager.Option, default to mai…

krzyzanowskim:
Add queue resulting queue to KingfisherManager.Option, default to main queue

Thanks to this `retrieveImageWithURL()` is more configurable, in particular can be called synchronously easily if necessary.


onevcat:
Seems cool. Thank you!

PR #72, opened 2015-06-30
Add an indicator when loading images.

alexliubj:


onevcat:
Hi, @alexliubj 

Thanks for the pull-request, it looks great.

But instead of making the indicator a default action and creating & adding the indicator view every time loading an image, would I ask for some change before it can be merged?

1. Make an option on it, to create and show the indicator only when the option is true. I suggest to add an associated object (like `kf_showIndicatorWhenLoading`) to `UIImageView` to hold this option. So the users could set it whenever they need.

2. Rely on the option mentioned in 1 to create the indicator view. It is addition cost to create and draw a view, so please avoid doing it every time. Please also do not remove it once added. Compare to adding/removing it, just hide it would be a better tradeoff.

Kingfisher/UIImageView+Kingfisher.swift
```diff
@@ -110,6 +111,21 @@ public extension UIImageView {
                          progressBlock: DownloadProgressBlock?,
                      completionHandler: CompletionHandler?) -> RetrieveImageTask
     {
+        var _indicator = UIActivityIndicatorView(activityIndicatorStyle:.Gray)
```
vfn: Indicator can be a constant. I would also remove the `_` prefix as it doesn't match the current coding style. Any thoughts @onevcat 
```swift
let indicator = UIActivityIndicatorView(activityIndicatorStyle:.Gray)
```

onevcat: @vfn Sure. I guess I will improve this p-r later since there is no response from @alexliubj

alexliubj: Let me check it ASAP. Sorry for replying to late, I was super busy last week.

alexliubj:
Updated code, please help me check again.

Kingfisher/UIImageView+Kingfisher.swift
```diff
@@ -104,12 +106,30 @@ public extension UIImageView {
...
     */
+    
     public func kf_setImageWithURL(URL: NSURL,
                       placeholderImage: UIImage?,
                            optionsInfo: KingfisherOptionsInfo?,
                          progressBlock: DownloadProgressBlock?,
                      completionHandler: CompletionHandler?) -> RetrieveImageTask
     {
+        
+        var indicator = UIActivityIndicatorView(activityIndicatorStyle:.Gray)
```
vfn: Indicator should not be created twice. I believe you only won't to create it inside the if below

Kingfisher/UIImageView+Kingfisher.swift
```diff
@@ -104,12 +106,30 @@ public extension UIImageView {
...
+        if ( showIndicatorWhenLoading == true)
+        {
```
vfn: Curly brackets should be in the same line as the if, to follow the same coding style.

Kingfisher/UIImageView+Kingfisher.swift
```diff
@@ -104,12 +106,30 @@ public extension UIImageView {
...
+    
     public func kf_setImageWithURL(URL: NSURL,
                       placeholderImage: UIImage?,
                            optionsInfo: KingfisherOptionsInfo?,
                          progressBlock: DownloadProgressBlock?,
                      completionHandler: CompletionHandler?) -> RetrieveImageTask
     {
+        
+        var indicator = UIActivityIndicatorView(activityIndicatorStyle:.Gray)
+        if ( showIndicatorWhenLoading == true)
```
vfn: `if showIndicatorWhenLoading {` should be enough

Kingfisher/UIImageView+Kingfisher.swift
```diff
@@ -104,12 +106,30 @@ public extension UIImageView {
...
+        {
+            indicator = UIActivityIndicatorView(activityIndicatorStyle:.Gray)
+            indicator.center = self.center
+            indicator.autoresizingMask = UIViewAutoresizing.FlexibleLeftMargin|UIViewAutoresizing.FlexibleRightMargin|UIViewAutoresizing.FlexibleBottomMargin|UIViewAutoresizing.FlexibleTopMargin
```
vfn:
```swift
indicator.autoresizingMask = .FlexibleLeftMargin | .FlexibleRightMargin | .FlexibleBottomMargin | .FlexibleTopMargin
```

Kingfisher-Demo/ViewController.swift
```diff
@@ -57,12 +57,12 @@ extension ViewController: UICollectionViewDataSource {
     
     override func collectionView(collectionView: UICollectionView, cellForItemAtIndexPath indexPath: NSIndexPath) -> UICollectionViewCell {
         let cell = collectionView.dequeueReusableCellWithReuseIdentifier("collectionViewCell", forIndexPath: indexPath) as! CollectionViewCell
+        cell.cellImageView.kf_showIndicatorWhenLoading = true
```
vfn: Why don't you add the activity indicator flag as a boolean option that's part of `KingfisherOptionsInfo`

Kingfisher/UIImageView+Kingfisher.swift
```diff
@@ -134,6 +159,7 @@ public extension UIImageView {
 
 // MARK: - Associated Object
 private var lastURLkey: Void?
+private var showIndicatorWhenLoading: Bool?
```
vfn: This will be set globally to all UIImageView. Meaning that if you set it to true to 1 UIImage, It will be true to all the other instances. One more reason to have the the activity indicator as a item in the options.

vfn:
The addition of an activity indicator would need to allow the user to define what activity indicator to use. I believe it's easier to create and add the activity indicator yourself before making the request and then remove it once the download finishes.

Leaving it out of the library, makes the code simple and cleaner.

onevcat:
@vfn Thanks for the review!

Since it is popular to show an indicator when loading, I guess it would be a good feature to add a default indicator for those "lazy guys" who just need an out of box solution. Surely we will make it an option so if the default indicator doesn't match the app's UI, framework users can always add their customized indicator.

The reason of using a boolean (`kf_showIndicatorWhenLoading`) is that this is a `UIImageView` attachment instead of relating with the Kingfisher core. Currently, the image view extension knows nothing about the options detail and only passes it to Kingfisher manager. I guess it would be better to keep this isolation.

I will handle this p-r, try to modify current code to a better quality and open another pull request later.

PR #76, opened 2015-07-09
Breaking retain cycles

vfn:
Capture self `UIimageView` weakly inside completion block


vfn:
@onevcat let me know if you need anything to be changed on this PR

onevcat:
@vfn It's cool. How could it happen that I didn't notice the retain cycle in task before. 

Thank you very much. It works perfectly. I am preparing a new release now. :)

vfn:
No problems!

I'm looking forward to the new version

Issue #70, opened 2015-06-30
Retina Images in Cache


mheicke:
Hi,
I think there's a problem with retrieving cached images on retina displays. In the diskImageForKey function in the ImageCache extension, the image is retrieved by using
```swift
    let image = UIImage(data: data)
```
So the scale of the device is not considered as it is in the URLSession(session: NSURLSession, task: NSURLSessionTask, didCompleteWithError error: NSError?) function in ImageDownloader.swift.

I think it should be 
```swift
    let image = UIImage(data: data, scale: UIScreen.mainScreen().scale)
```

Maybe it would be nice, if the scaling would be optional, nevertheless I think it should be the same in ImageCache and ImageDownloader.

Keep up with the good work,
matt

onevcat:
Sure, you are right. I will fix it soon.

onevcat:
Resolved in #77

PR #77, opened 2015-07-09
Scale factor and default options

vfn:
I have started this PR with the intention of fixing https://github.com/onevcat/Kingfisher/issues/70, although I found a couple more things that could be improved.

I'm happy to split this PR if required.


onevcat:
Perfect! Thank you!

PR #81, opened 2015-07-14
Improve #72

onevcat:
This is a pull request improves #72 , which added an indicator view to the UIImage to show a progress when loading.


onevcat:
@alexliubj @vfn Would you please check current implementation on it?

`imageView.kf_indicator` could be used to set basic style or color as well. And the indicator will not be removed until `kf_showIndicatorWhenLoading` explicitly set to false, to improve the performance a bit.

PR #88, opened 2015-08-05
Fixes an issue where images weren't appropriately rotated

### LINKED ISSUE #87: Wrong image orientation - Images get rotated after saved to disk
UIImage instances that are created from JPEGs have [`imageOrientation`](https://developer.apple.com/library/ios/documentation/UIKit/Reference/UIImage_Class/#//apple_ref/occ/instp/UIImage/imageOrientation) set accordingly.

When an image is later saved to the disk in JPEG format, the orientation is persisted too in EXIF information. Therefore, when it's loaded later on from disk, it has the correct orientation.

However, Kingfisher is using PNG serialization rather than JPEG that doesn't support the rotation flag (metadata about orientation is not persisted). source: http://stackoverflow.com/a/4868914/1161723

tomaskraina:
when saving to a PNG file and imageOrientation flag was lost.

This fixes #87


onevcat:
LGTM. And I guess I can implement to save the image as JPG format later if it is a JPG file.

PR #91, opened 2015-08-07
fix compile error on xcode 7 beta 5

cezheng:
stringByAppendingPathComponent seems to have become a NSString exclusive method on Xcode 7 beta 5, this is just a quick fix.


onevcat:
LGTM. Thanks!

PR #96, opened 2015-08-17
Add default value to APIs

hazytint:


onevcat:
Hi, @hazytint 
Thanks for the pull request, it seems great.

Different from Objective-C, Swift has default parameters. So in fact we could just leave only one of these APIs (the full one). However, there is a reason for leaving several versions of the same API. The auto completion for methods containing default parameters in Xcode 6.x is really suck. You will always get the full version of the method, even when you need some parameters being default. That means you have to remove some of them manually often, which is a waste of time. So I decided to have different versions of the same API for you to choose from the auto complete popup list.

But it seems things changes in Xcode 7. Now in Xcode 7 the auto completion of such methods gets improved. So I will reconsider to add default values to APIs after the new version of Xcode released later.

PR #108, opened 2015-09-18
Fixed issue with image when placeholder was not set.

borut-t:
This resolves imageView blinking when changing images fast.

### DIFF
--- a/Kingfisher/UIImageView+Kingfisher.swift
+++ b/Kingfisher/UIImageView+Kingfisher.swift
@@ -184,7 +184,10 @@ public extension UIImageView {
             indicator?.startAnimating()
         }
         
-        image = placeholderImage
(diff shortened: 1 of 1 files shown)


onevcat:
Hi, @borut-t 
Thanks for the p-r. But it will cause a problem when reusing the image view in a cell. The image in a reused cell will be the original one instead of `nil` while loading, it makes it looks strange when scrolling a table view or collection view.

If you need to keep the earlier image while loading, you could just pass that image as a placeholder image:

```swift
imageView.kf_setImageWithURL(url, placeholderImage: imageView.image)
```

onevcat:
I am closing it.

borut-t:
I am not sure if this is ok. Perhaps it would be better if would be vice-versa. So by default image is not set before loading new image. Anyway, it's your call.

Thanks!

PR #122, opened 2015-10-18
add support for gif

liurenqiu520:
just when data is gif, use the originalData for save


onevcat:
Cool. Thanks for it. I guess I will add fully GIF support for it later since it is a common request.

#122

alexdd55:
:+1:

PR #125, opened 2015-10-21
Fix a retain cycle

chunkyguy:
There is a retain cycle because we are capturing self strongly within a closure, this is causing the memory to not released. This can easily be fixed by capturing the self as weak.

### DIFF
--- a/Kingfisher/ImageCache.swift
+++ b/Kingfisher/ImageCache.swift
@@ -279,35 +279,37 @@ extension ImageCache {
                 completionHandler(image, .Memory)
             }
         } else {
-            block = dispatch_block_create(DISPATCH_BLOCK_INHERIT_QOS_CLASS) {
-                //Begin to load image from disk
-                dispatch_async(self.ioQueue, { () -> Void in
-                    
-                    if let image = self.retrieveImageInDiskCacheForKey(key, scale: options.scale) {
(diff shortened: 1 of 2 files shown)


onevcat:
Thanks for it.

That's true! 
Although there is no direct retain cycle here between `self` and `block`, the generated `block` task might be held outside by framework users and when this happens the `cache` is also held unintended. It will not be a big problem if you are using the default cache of Kingfisher, since it is a singleton. However, it is a good catch for implementing cache logic by using a standalone cache.

Thank you a lot.

onevcat:
@chunkyguy 

But there is another concern for it. Think about the code below:

```swift
let cache = ImageCache(name: "MyCache")
cache.retrieveImageForKey("Hi", options: KingfisherManager.OptionsNone) { (image, type) -> () in
    print("Finished")
}
```

You will never get the "Finished" print out since the cache is not held and released before the closure(task) could complete.

So I guess we could better to capture `self` temporarily in the closure with another variable, and set it to `nil` after the completion block get called. Maybe something like this:

```swift
var sSelf: ImageCache! = self
block = dispatch_block_create(DISPATCH_BLOCK_INHERIT_QOS_CLASS) {
    dispatch_async(sSelf.ioQueue, { () -> Void in
        // Do Work
        completionHandler(image, .Disk)
         sSelf = nil
    }
}               
```

By doing so, the completion block could always be called with some result, as well as the task will not hold the cache (self).

How do you think about it? Or do you have any better suggestion for it?

PR #126, opened 2015-10-21
Fixed broken Objective-C bridge

### LINKED ISSUE #124: Annoying warning about broken Objective-C bridge
```Umbrella header for module 'Kingfisher' does not include header 'Kingfisher.h'```

This only occurs if using code from master branch. Version in CocoaPods doesn't have that warning.

yurijmi:
Really strange stuff going on with CocoaPods. Fixes #124


yurijmi:
Travis build failed due to broken tests. Not my fault :/

onevcat:
Thanks for it.

It was a race condition in test only appears in CI server. I tried to give it a fix before but without luck. I should take a look back at it sometimes later. :(

Another run of testing could fix it. Let me do it.

PR #133, opened 2015-10-25
Implemented #132

### LINKED ISSUE #132: Ability to block animation of gifs
Create a parameter in kf_animatedImageWithGIFData (and parent functions) to use just the first frame of gif.

yurijmi:
Fixes #132 
May have some issues with tests. Sorry, can't update that :(

TODO:
- [x] Create an option to store image animated
- [x] Fix an error with storing image without animation EVEN that's not asked
- [ ] Fix optimization issues
- [ ] Fix unexpectedly found nil crashes

### DIFF
--- a/Kingfisher/ImageCache.swift
+++ b/Kingfisher/ImageCache.swift
@@ -157,6 +157,7 @@ public extension ImageCache {
                                    If `nil` is supplied, the image data will be saved as a normalized PNG file.
     - parameter key:               Key for the image.
     - parameter toDisk:            Whether this image should be cached to disk or not. If false, the image will be only cached in memory.
+    - parameter animated:          Store image animated or not. Only applies to GIFs.
     - parameter completionHandler: Called when store operation completes.
     */
     public func storeImage(image: UIImage, originalData: NSData? = nil, forKey key: String, toDisk: Bool, animated: Bool = true, completionHandler: (() -> ())?) {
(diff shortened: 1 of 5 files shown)


yurijmi:
This pull request is ready, but:

* There's some issues regarding performance if user sets options for animated images on not animated ones.
* There's an issue? then download for GIF with NoAnimation option is started and you queue another one without that option. The issue is that you get the non animated one on the last query and the first query is canceled. This is not really an issue for Kingfisher, but for implementation of some apps.
* The tests needs to be updated

This is the list of the things that I can't fix by myself. Except the first one, I don't have to worry about it in my app but someone might.

onevcat:
@yurijmi Thanks for the pull-request. But may I know the main purpose of it?

If you just want to display the first frame of a GIF instead of an animating one, you can set the image again in the completion handler of Kingfisher's image view extension like this:

```swift
imageView.kf_setImageWithURL(some_url_to_a_gif) { 
    (image, error, cacheType, imageURL) -> () in
    cell.cellImageView.image = image?.images?[0]
}
```

So the improvement of it might be the performance. Indeed it will boost the decode time and memory usage for users who only want to use the first frame of GIF. However, the whole GIF image will be downloaded even you only need to display the first frame. Compared to the downloading time, I guess the decoding time would not be an overhead since it is implemented in C and just some operation on raw data (Remember Kingfisher is doing image decoding in a background progress as well, so there is no chance to block UI thread).

By introducing these two options, it is making the framework confusing and more complex. And the things you stored also might be not the same/cannot be restored to what you downloaded, which is also a potential problem.

So I guess this p-r will not be merged and we could find a better way to handle it later.

yurijmi:
I need to recode that since it's purpose not really clear. Basically this code is for handling caches if you don't want animated images. You can't save not animated GIFs to cache using "vanilla" approach.

PR #141, opened 2015-10-30
Updated fade in section in the README.md

tsimmi:
The fade in example in the documentation was wrong: [.Transition(ImageTransition.Fade(1))
This is correct: [.Transition: ImageTransition.Fade(1)]


onevcat:
Thanks for it, but It is the new syntax for version 1.7.

PR #157, opened 2015-11-16
ImageCache - custom cache path

samvanderhyden:
Added an optional path parameter to init. This is really useful for me as I want a cache in documents dir for user generated photos, and also to make the ImageCache backwards compatible from an existing cache.


onevcat:
Great! Thanks a lot for contributing.

PR #197, opened 2016-01-05
delete redundant code

tripleCC:


Kingfisher/ImageCache.swift
```diff
@@ -608,11 +604,11 @@ public extension ImageCache {
                     }
             }
             
-            dispatch_async(dispatch_get_main_queue(), { () -> Void in
-                if let completionHandler = completionHandler {
+            if let completionHandler = completionHandler {
+                dispatch_async(dispatch_get_main_queue(), { () -> Void in
                     completionHandler(size: diskCacheSize)
-                }
-            })
```
onevcat: I guess we could just simplify the API from an optional `completionHandler` to a non-option one. There would be meaningless if no callback supplied. Since that, we could get rid of the optional chaining here. Would you please file another commit for it? Thanks!

onevcat:
LGTM. Would you please send a fix for the comment as well?

And please ignore the broken tests. It is some race condition in CI environment and I am fixing it now.

onevcat:
@tripleCC 

LGTM. Would you please send a fix for the comment as well?

And please ignore the broken tests. It is some race condition in CI environment and I am fixing it now.

tripleCC:
I guess there are more API can be simplified from an optional completionHandler to a non-option one, such as retrieveImageForKey:options:completionHandler:。 How do you think about it?

onevcat:
Yes, you are right.

tripleCC:
It seems that writing ` let (options, targetCache, downloader) = parseOptionsInfo(optionsInfo) `will not cause a complier error any more in retrieveImageWithResource:optionsInfo:progressBlock:completionHandler:

onevcat:
The "options" is being rewritten now for [milestone 2.0.0](https://github.com/onevcat/Kingfisher/milestones/Version%202.0.0), so you could just leave it there. They will disappear soon.

tripleCC:
Cool!

Issue #202, opened 2016-01-08
ImageCache storeImage not using data from download


avalanched:
Is there a reason why storeImage computes its data from the UIImage instead of using the nsdata retrieved from the download? for jpg images although quality is at 100% it is compressed again.
(which also results in way bigger images than the ones that could be received from the download)

I'm wondering, are there any performance reasons?
Or could this be a usefull PR?

onevcat:
It is more likely a historical reason. There is no original data input for this API. Currently, it could ensure the data is valid and could represent an image. However, this is not a must in my opinion since we should  be able to trust the data from sever here. Another reason is that the `ImageCache` could be used separately.  This could make sure even if the user did not pass in an `originalData`, we could also store the image correctly.

A PR on it is warmly welcomed. We should respect the `originalData` and store it directly. If it is lack of the data, we should convert the input image to data and store the result data instead.

PR #208, opened 2016-01-14
use originalData in favour of image

avalanched:
This is a followup pull request on #202.
If originalData is available it will be used in favour of the supplied image to store to the cache.

This will omit compressing the image in case of a JPEG.


onevcat:
Cool. Thanks!

PR #249, opened 2016-02-24
Provide a simple prefetcher implementation.

krider2010:
This functionality is available in `SDWebImage` which influenced this
library, and was something we needed in migrating over.

It is a simple implementation, using default managers etc. but could be
expanded upon pretty easily to use different managers, downloaders, and
queues etc.

A utility method to check for cached images (useful when deciding
whether to prefetch or not) has also been added to `KingfisherManager`.

All new code is tested.


Sources/KingfisherManager.swift
```diff
@@ -159,6 +163,21 @@ public class KingfisherManager {
...
+    /**
+     Determine if a cached image exists for the given image, as keyed by the URL. It will return true if the
+     image is found either in memory or on disk. Essentially as long as there is a cache of the image somewhere
+     true is returned.
+     
+     - parameter url: The image URL.
+     
+     - returns: True if the image is cached, false otherwise.
+     */
+    public func cachedImageExistsforURL(url: NSURL) -> Bool {
```
onevcat: Although it could be useful and a quick way to check an image cached or not, I don't think it's a good idea to have a method for it here. 

I believe it should only in the `ImageCache`. Furthermore, since `cache` of `KingfisherManager` could be set by the users, it is not clear and might be confusing which cache is used for the checking.

Generally speaking, I cannot agree to add this here. We could use the same methods in `ImageCache` directly. Would you mind to remove this before this p-r could be merged?

Sources/ImagePrefetcher.swift
```diff
@@ -0,0 +1,119 @@
...
+        cancelPrefetching()
+        
+        prefetchURLs = urls
+        
+        guard urls.count > 0 else {
+            CompletionHandler?()
+            return
+        }
+        
+        for (var i = 0; i < maxConcurrentDownloads && requestedCount < urls.count; i++) {
```
onevcat: The C style loop will be deprecated and removed in a later Swift version. So we would better to rewrite it in a Swift way.

Sources/ImagePrefetcher.swift
```diff
@@ -0,0 +1,119 @@
...
+            startPrefetching(i, progressBlock: progressBlock, completionHandler: completionHandler)
+        }
+    }
+   
+    internal func cancelPrefetching() {
```
onevcat: The cancel logic contains some gutter here. Consider about the situation that you call `prefetchURLs` before the previous `prefetchURLs` download finishing.

Because there is no cancelling now for the download process, there is a chance that the `completionHandler` of `downloadAndCacheImageWithURL` for previous fetch is called wrongly and `finishedCount` gets added by mistake.

onevcat:
Hi, @krider2010 

Thanks for the pull request. I just took a look at it and I believe it is a cool feature that could boost the speed.

Except the comment above for the implementation detail, I have several concern and want to have a discussion.

1. The `KingfisherManager` type is now used as a "glue" between `ImageDownloader` and `ImageCache`. Framework users/image extension could use `KingfisherManager` to download and cache images without knowing the detail of downloader and cache. However, I don't think it is the situation for `ImagePrefetcher`. The `prefetcher` property is even not used in `KingfisherManager`. It would be better to 
2. Now `ImagePrefetcher` is relying `KingfisherManager` and caching the downloaded images to the manager's cache. It is probably not a good choice. There is no chance for a user who wants to cache the prefetched images to another cache besides the manager's one.

How do you think about these?

Sources/KingfisherManager.swift
krider2010: I'll happily move it to `ImageCache` but I feel it is important for our use case with prefetching as we do not care what the cache is doing. We let the library do that!

Sources/ImagePrefetcher.swift
krider2010: I'll clear it up.

krider2010:
@onevcat can you confirm what it would be better to do in point 1. ? It got cut off!

As for point 2. I did mention above "It is a simple implementation, using default managers etc. but could be expanded upon pretty easily to use different managers, downloaders, and queues etc." but I had no time to do this for our needs. I think it is a perfectly acceptable initial/default case. The solution here is to expand it more which I can look at doing.

onevcat:
Oops, sorry for the cut off. I was considering whether we should contain a prefetcher in the manager as a member. In fact at least we are not using it at all in the manager.

But yes you are right. It is already a good start point for it now. I will consider how to improve/refactor this to a better way later before it could be released. Now please just fix that obvious things and I will merge it when everything looks good!

Thanks in advance!

krider2010:
@onevcat let me know what you think of the latest changes :)

onevcat:
Great. Thanks for this.

PR #252, opened 2016-02-26
Added user interaction to animation

ivanbruel:
When using Kingfisher with any type of animation, it was just breaking user interaction and not allowing users to scroll through lists on interact with the `UIImageView`s themselves.

This was mostly due to the fact that `.AllowUserInteraction` was not an option on the `UIView.transitionWithView` call.


onevcat:
@ivanbruel Thanks!

PR #256, opened 2016-02-28
Make cachePathForKey public [resolves #255]

### LINKED ISSUE #255: How to find local cache path for disk
So I see that you can find the path by using `ImageCache("cacheName").diskCachePath` but I'm confused about how I can find the entire `String` path of an image that has been cached to disk. So rather than the actual `UIImage` which is store as `Image`; I'm actually looking for the `String` path

I've tried `ImageCache("cacheName").diskCachePath/stringForKey` and can't seem to access the entire file path after knowing that it has actually been saved to disk. Any help would be greatly appreciated.

jasonnoahchoi:
Understandably, the method was private; however, this commit solves the problem of anyone that requires access to the actual file path of the cached image. This is especially useful for projects that are building iOS apps that are showing content with `UIWebView`s and lacks access to Internet connection.

Ref: #255

### DIFF
--- a/Sources/ImageCache.swift
+++ b/Sources/ImageCache.swift
@@ -604,6 +604,18 @@ extension ImageCache {
             })
         })
     }
+    
+    /**
+    Get the cache path for the key.
+    It is useful for projects with UIWebView or anyone that needs access to the local file path
+    
+    i.e. <img src='key'>
+    */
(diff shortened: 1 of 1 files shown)


onevcat:
Thanks, it makes sense.

PR #257, opened 2016-02-29
Disable background decoding for images from the in memory cache.

wedrive-rb26:
- this improves the performance of image loading for in-memory cached
  images.

We observe a flicker when returing to an image view that renders
multiple images using `kf_setImageWithResource`.  We traced through the
calls and found a place where Kingfisher attempts to re-decode an image
that has already been decoded and stored in the in memory cache.

We think that it is safe to pull out an image from the in memory cache and not
attempt to decode again.


onevcat:
Yes, you are right.

PR #266, opened 2016-03-14
Change pod version from 2.1 to bump version

BilalReffas:


onevcat:
Hi,

Thanks for the p-r. But I think it is no need to specify a minor version in dependency.

PR #268, opened 2016-03-22
Some changes for converting to Swift 2.2

rain2540:
1.Update the style of selector in "ImageCache.swift"
2.Update a method name in "String+MD5.swift"


onevcat:
@rain2540 Thanks for it. I will check & merge it later!

Issue #187, opened 2016-01-02
How to start/stop/pause GIFs


victorBaro:
After setting a GIF URL like so: `gifImageView.kf_setImageWithURL(gifURL)` how can you pause it?

ImageView does not respond anymore to `stopAnimating()` and `startAnimating()` functions.

onevcat:
`Kingfisher` is using the `image` property directly for setting a GIF image to image view. So the animated image is not a sequence of images. It is not be played by `startAnimating` API, so you could not stop it by `stopAnimating` as well.

A workaround for it is adding a wrap to convert the single animated image to an array of images and then play it with `startAnimating` in the image view set method completion handler. Like this:

```swift
imageView.kf_setImageWithURL(URL, placeholderImage: nil, optionsInfo: nil, progressBlock: nil,
    completionHandler: {
        image, error, cacheType, imageURL in
        if let image = image {
            imageView.animationImages = image.images
            imageView.animationDuration = image.duration
            imageView.animationRepeatCount = 0
            imageView.image = image.images?.last
            imageView.startAnimating()
        }
    })
```

Then, you can call `stopAnimating` on the image view to stop it when you need:

```swift
imageView.stopAnimating()
```

I hope it could help.

PR #271, opened 2016-03-24
supporting stop and start animation of gif image

xiaoxiaocainiao:
i change the default demo of source project, is that ok ?

### DIFF
--- a/Demo/Kingfisher-Demo/ViewController.swift
+++ b/Demo/Kingfisher-Demo/ViewController.swift
@@ -68,7 +68,16 @@ extension ViewController {
         
         cell.cellImageView.kf_showIndicatorWhenLoading = true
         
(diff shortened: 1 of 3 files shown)


onevcat:
Thanks for considering to contribute. 

But does this make sense for you? I think it could give a way to stop GIF without changing current implementation: https://github.com/onevcat/Kingfisher/issues/187 . I cannot see much improvement so I am afraid it will not be merged yet.

Anyway, thank you!

xiaoxiaocainiao:
giphy is very popular. i just want to extension UIImageView class, add a methord for it,  with this methord , u can switch the starting animation and  stop animation more simple.  i think   [#187](https://github.com/onevcat/Kingfisher/issues/187) is a  simpler solution to do this.  thank you~

PR #278, opened 2016-03-31
Allow images other than JPEG, PNG or GIF (.ICO, as example)

vmartinelli:
Allow images with other formats than JPEG, PNG or GIF to be converted from downloaded image data (.ICO, as example)


onevcat:
Cool, thanks for it.

PR #287, opened 2016-04-15
implement NSButton extension for OSX

jackyzh:
I am not sure if there's a way to combine UIButton with NSButton. So I didn't change UIButton+extension, for now, I just create a new file to implement NSButton as it doesn't have backgroundImage property but alternateImage instead.


onevcat:
@jackyzh Sorry for the late reply!

Thanks for the pull request. It seems cool at first glance. I will take a look deeper later. Since the APIs of `UIButton` and `NSButton` are quite different, I guess it should be no problem to keep them separated. It will not be hard to combine them into one too, if we'd like to do so later.

Cheers.

Issue #269, opened 2016-03-23
Is there a version for using in App Extension ?


macbaszii:
I couldn't compile code because of the error in ImageCache.swift is using UIApplication.SharedApplication() 

`'sharedApplication()' is unavailable: User View Controller based solution where appropriate instead.`

yonaskolb:
Having this issue too

macbaszii:
@yonaskolb I changed to use SDWebImage for an extension.

onevcat:
Fixed in #290

PR #290, opened 2016-04-21
Add support for App Extension target

tripleCC:
Inspired by [SDWebImage](https://github.com/rs/SDWebImage/issues/1082), using runtime to judge      whether sharedApplication() is available in current target #269 .
It also can be done with the compile flags


onevcat:
Brilliant! Thanks for this.

PR #291, opened 2016-04-22
exclude NSButton extension for iOS, tvOS and watchOS

jackyzh:
And move NSButton into Sources folder. Sorry for didn't put it into right place in last PR.


onevcat:
Thanks for it. I also removed some unnecessary reference of NSButton extension in #292

PR #292, opened 2016-04-22
Exclude NSButton extension

onevcat:


jackyzh:
I think you can remove some codes of os version check in UIButton extension file as well.

onevcat:
It makes sense! Thanks for pointing it out.

PR #299, opened 2016-04-26
Simplify some code using default arguments

ejmartin504:
Was interested in how ImageCache worked, and started poking around the source. Noticed two instances where code could be simplified by using default arguments. I'm not sure if this was intentionally left this way, but it appears to me that there is no compelling reason to have two separate method signatures for these aspects of ImageCache. Let me know either way!

### DIFF
--- a/Sources/ImageCache.swift
+++ b/Sources/ImageCache.swift
@@ -144,34 +144,18 @@ public class ImageCache {
 // MARK: - Store & Remove
 extension ImageCache {
     /**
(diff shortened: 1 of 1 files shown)


onevcat:
It was a workaround for the lack of autocompletion for methods with default parameters in Swift 1.x and Xcode 6.x. Now in Swift 2 this part works better so there is no reason to keep them!

Thanks for it. 👍

PR #306, opened 2016-05-01
Fix 'Cannot form weak reference to instance of class Kingfisher.AnimatedImageView'

xspyhack:
This problem occured when you are using `replace segue` to show the view controller which have Kingfisher.AnimatedImageView kind property.


onevcat:
Oops. Although I don't know why, but the code looks good to me.

Thanks for it.

xspyhack:
Cannot form weak reference to instance (0x14ddb6e30) of class Kingfisher.AnimatedImageView. It is possible that this object was over-released, or is in the process of deallocation.

It seems when `displayLink ` object getting assigned is being deallocated. `stopAnimating()` or ` isAnimating()` is being called within `deinit`, so that it will crash if `setImage` haven't be call ( so `displayLink ` still nil ) before `deinit`.

onevcat:
Cool. It makes sense. Thanks for the explanation.

PR #311, opened 2016-05-04
a cleaner way of dealing with background task

biku:
If you are willing to use a APPLICATION_EXTENSION_API_ONLY = NO flag in your Build Settings, I think this is much cleaner approach to background task handling


onevcat:
It is intended to support to use Kingfisher in app extension, so we cannot use `sharedApplication` here.

PR #312, opened 2016-05-05
Fix name min length violation by SwiftLint

pixyzehn:
Fix the following error by SwiftLint and set the `pure` function's scope private.

`error: Variable Name Min Length Rule Violation: Variable name should be 2 characters or more: currently 1 characters (variable_name_min_length)`

If you'd like to allow the violation, please add `- variable_name_min_length` in `.swiftlint.yml` as a disabled_rules.

### DIFF
--- a/Sources/AnimatedImageView.swift
+++ b/Sources/AnimatedImageView.swift
@@ -370,6 +370,6 @@ extension Array {
     }
 }
 
-func pure<T>(a: T) -> [T] {
-    return [a]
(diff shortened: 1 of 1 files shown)


onevcat:
Hi, @pixyzehn

Thanks for the p-r. Would you mind to send another pull request but change the parameter name from `type` to `value`? It should not be a type in fact.

Thank you!

pixyzehn:
@onevcat 
Sure. I'll send another one.

PR #313, opened 2016-05-05
Avoid variable name min length rule violation by SwiftLint

pixyzehn:
I replaced `type` with `value`. Thank you for pointing it out.

This PR is related to https://github.com/onevcat/Kingfisher/pull/312.


onevcat:
👍 

Thanks!

PR #323, opened 2016-05-19
Fix spelling

dapenggao:
Fix some cases where "download" is misspelled. This should not be a breaking change, because all names modified are non-public.


onevcat:
Embarrassing! Thanks for it.

PR #334, opened 2016-06-09
Removed Implicit Unwrapping of CacheType

gscalzo:
First of all thank you for your fantastic Library!
We notified some crashes in production cause by the fact that the cache type was implicitly unwrapped and if was `nil` if the image is not cached


onevcat:
It makes sense! Thanks for this.

Issue #338, opened 2016-06-15
Swift 3 compatibility


GigabyteTheOne:
It would be great to have a Swift 3 compatibility for this lib

PR #340, opened 2016-06-15
Swift 3 compatibility

GigabyteTheOne:
PR for #338


onevcat:
I believe it is a good start point. We will do some more detail work on compatibility for following Swift 3 API guidelines. Thanks for it.

Issue #245, opened 2016-02-17
Return statusCode from ImageDownloader callback if failures occur


stephenvisser:
I'd like to differentiate between 404's (image doesn't exist) and auth / server exceptions.

One option:

`NSError(domain: KingfisherErrorDomain, code: KingfisherError.NotModified.rawValue, userInfo: ["statusCode": ...])`

mrstone:
+1 This would be a great addition

onevcat:
Basically implemented in #346

PR #346, opened 2016-06-20
Cancel donwloading when status code is invalid

koji-inoue:
I'm sorry for the previous one.

This is mentioned at #245

Currently, when server respond with 404 and image content type, this will cache the image and not give us any information.
I think it's not good. 
This change will make kingfisher to avoid to cache image when server respond with invalid status code like 404.
When status code is not 200 or 201, callback handler include error which has status code and localized description.

Thank you for your great library.


onevcat:
Hi, @koji-inoue 

Thanks for this p-r. It looks good to me at a first glance! I will take a detail look to it later.

Thank you.

PR #352, opened 2016-06-29
Swiftlint 

leoMehlig:
Disables swiftlint rules that fail with swift 3.
type_name and conditional_binding_cascade are bugs/errors in swift lint

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -8,3 +8,6 @@ disabled_rules:
   - type_body_length
   - valid_docs
   - function_parameter_count
+  - type_name
+  - force_try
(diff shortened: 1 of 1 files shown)


onevcat:
@leoMehlig Thanks for the p-r.

SwiftLint is keeping introducing new rules that prevent compiling. I found sometimes the newly introduced rules make no sense and it might cause users a bit confused since the error, I decided to remove SwiftLint from this project. The master branch is now already SwiftLint-free, and I will do it for the swift3 branch later too. So I just close this p-r without merging it.

Anyway, thank you for the contributing!

PR #362, opened 2016-07-20
Future WebP support, base on swift3

looseyi:


onevcat:
Thanks for this cool p-r and your great effort on it. 
In fact I am now considering whether Kingfisher should support WebP or not.

WebP is currently not widely supported or used, especially on Apple's platforms. I believe the costing of introducing a new framework as dependency and disabling bit-code support is not negligible for quite a few users. The core concept of Kingfisher is lightweight, with only necessary part being included.

I will leave this pull request open for a while before I could have a final decision later.

Anyway, good job on it. Thanks!

looseyi:
👌

onevcat:
Will not added in Kingfisher. At lease for now.

Sadmansamee:
It will be very hellfull and nice if there was WebP Support

onevcat:
From Kingfisher version 3, you could implement an `ImageProcessor` and convert any format (including webp) to `UIImage`. We will not introduce a webp framework dependency into Kingfisher, but you could implement it quite easily now. See [this](https://github.com/onevcat/Kingfisher/wiki/Cheat-Sheet#processor) for more.

onevcat:
P.S. There is a WebP framework as plugin for Kingfisher now. Check https://github.com/yeatse/KingfisherWebP

PR #367, opened 2016-07-22
Swift 2.3 compatibility

macecchi:
* Setsthe targets Swift version to 2.3
* Convert to Swift 2.3 syntax
* Update the CI image to use Xcode 8


macecchi:
The CI build failed because Travis sucks at building Xcode projects. If you want to trigger the tests again they should work. Or you can check the build from my fork, which passes: https://travis-ci.org/macecchi/Kingfisher/builds/146659059

onevcat:
:] Thanks for the great work!

PR #383, opened 2016-08-04
Fix gif frame duration calculation to match that of webkit

MichaelTzach:
Fix gif frame duration calculation to match that of webkit
This bug caused wrong animation speeds for some gif when not using AnimateImage

### DIFF
--- a/Sources/Image.swift
+++ b/Sources/Image.swift
@@ -238,7 +238,7 @@ extension Image {
                 
                 let unclampedDelayTime = gifInfo[kCGImagePropertyGIFUnclampedDelayTime as String]
                 let delayTime = gifInfo[kCGImagePropertyGIFDelayTime as String]
-                var duration = unclampedDelayTime ?? delayTime
+                let duration = unclampedDelayTime ?? delayTime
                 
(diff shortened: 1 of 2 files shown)


onevcat:
@MichaelTzach 

Hi, thanks for the p-r.

May I know what does the `gifDefaultFrameDuration` (0.100) and the `frameDuration.doubleValue > 0.011` means? I am not sure I could understand what happened here. Could you give some reference for this modification?

And do not worry about the CI. It should not be your issue, but a thread stall due to limited resource for open source project. I will take a look at it later.

MichaelTzach:
@onevcat 

Hey,
Thanks for the quick response.
We had a bug with gif playback when using the images downloaded using Kingfisher but without using AnimatedImage. I saw that when calculating a frame's length you used the kCGImagePropertyGIFDelayTime property.
When searching for the answer I came across this: http://stackoverflow.com/a/17824564/6690172 which explains the way webkit calculates the gif animation duration.

This is the way webkit does it:
1. UnclampedDelayTime
2. DelayTime
3. 100 milliseconds
* frames under 11 milliseconds are not allowed (in the past they were used to display blinking ads)

onevcat:
Make sense!

Thanks.

PR #385, opened 2016-08-04
Cache only option

victorg1991:
Could be some cases where you need to rely only in the cache and not make http requests.
I added a CacheOnly option that allows it

I hope you like it 
Any comments would be appreciated :)

Regards

### DIFF
--- a/Sources/ImageDownloader.swift
+++ b/Sources/ImageDownloader.swift
@@ -80,12 +80,14 @@ The error code.
 
 - BadData: The downloaded data is not an image or the data is corrupted.
 - NotModified: The remote server responsed a 304 code. No image data downloaded.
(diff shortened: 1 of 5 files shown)


onevcat:
Hi,

@victorg1991 Thanks for the p-r. It looks good, expect these two concerns for me. If you would like to fix them, I am glad to merge it.

1. The option name `CacheOnly` is a bit confusing. It could mean download it and do not cache it, which is not correct. Could you modify it to something more clear? Maybe `OnlyFromCache` would be better?
2. Please use space instead of tab as indent.

victorg1991:
Totally agree, I will change it and submit it again :)

Thanks

victorg1991:
@onevcat  I hope you like it now 😉

PR #386, opened 2016-08-05
Update README.md

broccolii:


onevcat:
Thanks!

And I think I will remove the old pod file style later.

PR #391, opened 2016-08-09
Swift 2.3 support

onevcat:


onevcat:
This will not merge but only for state review.

Swift 3 will be the default branch for Kingfisher later.

onevcat:
Added back compatible for Swift 2.2. So we could merge it now.

PR #396, opened 2016-08-15
backward compability to xcode7 added

pmierzejewski:


onevcat:
@pmierzejewski Thanks for the contribution. Good job!

However, Kingfisher will drop support for Swift 2.2 and Xcode 7 once Xcode 8 be released. And the master branch will be the Swift 3 version. So this p-r will not be merged (mainly because personal preferences. I do not like to introduce these kind of version definition condition into the code base).

Anyway, thank you! 👍

PR #399, opened 2016-08-16
Add customCreateImagesFromData Support

DianQK:
Hi~ @onevcat 

I think we can add some custom option.

Maybe like `customCreateImageFromData`, it's useful(maybe WebP).

Then, we still keep Kingfisher slim. 🎉

### DIFF
--- a/Sources/Image.swift
+++ b/Sources/Image.swift
@@ -42,6 +42,10 @@ private var animatedImageDataKey: Void?
 
 import ImageIO
 
+public struct ImageOption {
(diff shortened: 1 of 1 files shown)


onevcat:
It's a good idea to delegate the image decoding out, but we should have a better implementation for it (may be a decoding delegate instead of a static property like this).

I opened an issue for this and will give an implementation for this later. This p-r will be closed since that.

Thank you for the suggestion!

onevcat:
A better solution is implemented in Kingfisher 3.

PR #371, opened 2016-07-26
Swift 3 support

onevcat:

PR #402, opened 2016-08-16
Migrate to Swift 3 beta 6

andreyz:
Might help with migrating to Xcode 8 Beta 6


onevcat:
@andreyz Thanks for the great work. But it is already done here: https://github.com/onevcat/Kingfisher/pull/371/commits/15f911402fff8cb2d806eb6e103da264de803eb9

Issue #316, opened 2016-05-10
Fatal Error in prefetcher


mihaibulic:
I use the ImagePrefetcher to cache photos before I need them, so I frequently start and stop 2 prefetchers with several photos.

I occasionally run into a problem:  Fatal Error: Can't form Range with end < start
at this line in ImagePrefetcher.downloadAndCacheResource(resource: Resource):

`let pendingResources = self.prefetchResources[self.requestedCount..<self.prefetchResources.count]`

Is it that stop() was called on the prefetcher and a downloadTask is trying to stop itself, but it's actually already done (e.g requestedCount == prefetchResources.count)?

(also thanks for making this library, aside from this crash, it's great)

onevcat:
@mihaibulic Thanks for letting me know it.

May I know the number of images you are trying to fetch as well as the `maxConcurrentDownloads` you are using? It might be help for me to track what happens.

Thanks!

mihaibulic:
I have 2 separate prefetchers, never prefetching more than 3 images.  I don't change maxConcurrentDownload from the default

mihaibulic:
@onevcat, any word on this?

onevcat:
Hi, @mihaibulic 

I checked it and failed to reproduce your issue.

For you case, since you were only prefetching 3 images, the requested count should be always the same as prefetchResources.count. It will produce an empty slice and should not lead a crash.

I am now not sure what could lead the requested count be larger than the total resource count. Could you reproduce this error in a good way, or even better supply a simple demo so I could dig it deeper?

mihaibulic:
Hey @onevcat,

unfortunately it's not deterministic.  The only thing I have noticed is that once it starts to occur, it continues to occur.

Here is what I'm (basically) doing:

```
private let PREFIX = "https://storage.googleapis.com/spree-photo-storage/photos/1080x1920-"
private let SUFFEX = ".jpg"
private let IDS = [
    "0000a635-b38c-4521-bded-fdc173d884b2",
    "000130c9-884f-489e-a5f2-65e7ec1066fc",
    "000140b2-5f09-4845-90c2-0e2a56109243"
]

private var currentPrefetched : ImagePrefetcher?

/**
This function is generally called every 30 seconds, but may be called back-to-back
*/
private func prefetch(photoIds: [String]) {
    currentPrefetcher?.stop()
    var prefetchUrls : [NSURL] = []

    for id in photoIds {
        if let urlNotNil = generateURL(photoId) {
[...]

PR #422, opened 2016-09-02
Fix for Issue 316. Crash: 'Fatal Error: Can't form Range with end < start

danhollywells:
requestedCount incremented beyond the prefetchedResorces count. See Issue #316  

Fix: eliminate requestedCount

Because the start() method captures the initial X Resources by index and ships them off to startPrefetchingResource() and that methods recursive behavior, you can end up in a state where all the items have been processed but the initial X resources will get double processed because they were captured in the for loop in the start() method and again in the recursion of the startPrefetchingResource(). So if during the prefetch you get into this state and you call stop() you will end up with an unbalanced Range when sorting out the failedResources.

### DIFF
--- a/Sources/ImagePrefetcher.swift
+++ b/Sources/ImagePrefetcher.swift
@@ -60,18 +60,18 @@ public class ImagePrefetcher {
     
     private var tasks = [NSURL: RetrieveImageDownloadTask]()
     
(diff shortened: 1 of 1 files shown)


onevcat:
LGTM! Thanks for pointing this out!

Issue #315, opened 2016-05-09
Placeholder image GIF support?


andrewabdalla:
Is it possible to use a GIF as the placeholder image while loading a photo?

I've have tried while using AnimatedImageView() but I only see the first frame of the GIF until the URL image loads.

onevcat:
Currently no. But it will be an interesting feature. I will consider it in a later version. (P-R for it will be welcome as well, let me know if you want to do so.)

jdmoreira:
I'll do this if that's ok with you. P-R soon.

onevcat:
@jdmoreira Sure!

jdmoreira:
Pull Request -> #425

onevcat:
Merged.

PR #425, opened 2016-09-07
Support for different types of indicators, including gif images

jdmoreira:
This pull request is my proposed solution for issue #315.

I basically changed 'kf_indicator' from being a view to being of type 'Indicator', which is a protocol. Having a protocol allows us to add different kinds of Indicators.

I've already added a struct that conforms to 'Indicator' for Images (including GIF support) and another for NSProgressIndicator/UIActivivityIndicator.

Since the setter for 'kf_indicator' is private, the user chooses what indicator to use through the 'kf_indicatorType' enum. There are default values, plus values that receive parameters as associated values.
There's a .custom value that allows the user to pass any type that conforms to 'Indicator', allowing the user to have his own custom indicator view.

Also, since I needed to save any swift type that conforms to a  protocol in the associated objected 'kf_indicator', I had to created a wrapped/lifted version of objc_set/objc_getAssociatedObject. That's my first commit, I added this to file WrappedAssociatedObjects.swift and replaced all objc_set/get_AssociatedObject calls with calls to the wrapped version.

All suggestions and changes are welcome.


Sources/ImageView+Kingfisher.swift
```diff
@@ -184,131 +177,94 @@ extension ImageView {
     }
 }
 
+/**
+ Enum for the types of indicators that the user can choose from.
+ */
+extension ImageView {
+    public enum IndicatorType {
+        case none // no indicator
```
onevcat: To follow Swift 2 API guideline, please use upper case for enum member here.

onevcat:
Hi, @jdmoreira 

Just had a brief look at it. Great implementation!

But I have some concern before merging it. You have removed `kf_showIndicatorWhenLoading` and `IndicatorView` typealias, which were marked as `public` before. I understand that by doing so could  make code cleaner and expressive. However, as a framework, it is also important to keep back compatible as mush as possible. As semantic versioning requiring, we may need a major update for such API breaking. IMO, it is not a good idea to do so for this kind of change.

So before it could be merged, may I ask to add back the `IndicatorView` typealias as well as the `kf_showIndicatorWhenLoading` property, and make it behave as `.activity` is set when `kf_indicatorType` is set to `.none`. (and maybe mark this API as deprecated with warning).

I also commented in the code for a trivial style/convention. And it would be even greater if you can also make a p-r to swift3 branch (there you could do the breaking change by removing the `kf_showIndicatorWhenLoading`), since the main development of Kingfisher will continue there instead of Swift 2.x version.

Thanks again for it. It's impressive to me!

jdmoreira:
Absolutely. I'll do it :)

onevcat:
Fix and merged in #430

hifall:
Looks like an awesome feature! I can't wait to add this in ASAP.

Will the document be updated soon?

onevcat:
@hifall It's already in version 2.6.0. And I will continue work on it for Swift 3 later.

PR #430, opened 2016-09-12
Jdmoreira master

onevcat:


jdmoreira:
I've seen you merged and did a few changes. I'm sorry I didn't had much time the last couple of days. Thanks for taking the time.

I agree that the keeping the objc_set/getObject and using Box is a much better solution 👍

onevcat:
@jdmoreira Hi, I think your wrapper solution is also great. But the blocking is that it cannot compile in Release mode with -O optimization. It crashes LLVM in the conversion of `T` to `AnyObject`. Furthermore, in Swift 3 we cannot depend on `AnyObject` to check class (since everything is imported as `Any` now). So I guess it would be better to leave it to us developer to decide which should be in box.

I am working on Swift 3 for the indicator now. Glad and honor that you help to improve it!

PR #435, opened 2016-09-15
Add kf namespacing

onevcat:
This p-r adopts to a new namespacing system. It could avoid possible naming conflicting by introducing an extra struct layer. So we could get rid of the `kf_` prefix. Instead, we could use an easier `kf.` to limit Kingfisher related APIs in its own namespace.

### DIFF
--- a/Demo/Kingfisher-Demo/ViewController.swift
+++ b/Demo/Kingfisher-Demo/ViewController.swift
@@ -57,16 +57,16 @@ extension ViewController {
     
     override func collectionView(_ collectionView: UICollectionView, didEndDisplaying cell: UICollectionViewCell, forItemAt indexPath: IndexPath) {        
         // This will cancel all unfinished downloading task when the cell disappearing.
-        (cell as! CollectionViewCell).cellImageView.kf_cancelDownloadTask()
+        (cell as! CollectionViewCell).cellImageView.kf.cancelDownloadTask()
     }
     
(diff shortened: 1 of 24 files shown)


webfrogs:
Looks great. But I have some different ideas about the implement of namespace.It's about adding namespace extension with `Struct` type in swift.

1. Using the generic struct `Kingfisher<Base>` as the default implement for `Class` Type
2. When extension the `Struct` type like Swift 3's `Data`  type. Define a new struct named like XXXProxy to implement the `KingfisherCompatible` type which is defined before. Then writing all extension of the Proxy struct to do the  extension jobs. By doing this, We can avoid using type cast to get the right type, because I believe type casting is the bad code smell of swift.

First, making some small change of `KingfisherCompatible` protocol, There is no need to give the `set` property.
```Swift
protocol KingfisherCompatible {
    associatedtype CompatibleType
    var kf: CompatibleType { get }
}
```

Then, extension Data type can be something like this:
```Swift
struct DataProxy {
    fileprivate let value: String
    init(_ value: String) {
        self.value = value
    }
}

extension Data: KingfisherCompatible {
    typealias CompatibleType = DataProxy
    var kf: CompatibleType {
        return DataProxy(self)
    }
}

extension DataProxy {
   // Do whatever you want, and you can use value to get the exactly type: Data
}
```

webfrogs:
#441 , I create a pr ,just change the way of implemention

PR #456, opened 2016-09-27
Add option to supply default path extension

boyvanamstel:
Especially useful when using Kingfisher on macOS and wanting to add
drag-drop. Most web input fields won't accept files without a path
extension (like .jpg).

Currently only supports a single default path extension. Could later
be expanded to include a path extension per image.


onevcat:
LGTM

Issue #459, opened 2016-09-27
Crash on iPhone device (iOS 10.0.1), when options append `ResizingImageProcessor`


Limon-O-O:
When I want to resize the `https://bubble-m.b0.upaiyun.com/thumbnail/3/79a0c0f859cc46fd9463dc7377c1318c` image on my iPhone device with `ResizingImageProcessor `, 
and crash at `context.draw(self, in: CGRect(x: 0, y: 0, width: width, height: height))`, 
but not crash on simulator.

```swift
let url = URL(string: "https://bubble-m.b0.upaiyun.com/thumbnail/3/79a0c0f859cc46fd9463dc7377c1318c")!

var options: KingfisherOptionsInfo = [.transition(.fade(0.3)), .backgroundDecode, .scaleFactor(UIScreen.main.scale)]

let processor = ResizingImageProcessor(targetSize: CGSize(width: 42.0, height: 42.0))
        options.append(.processor(processor))

_ = (cell as! CollectionViewCell).cellImageView.kf.setImage(with: url,
                                                   placeholder: nil,
                                                   options: options,
                                                   progressBlock: nil,
                                                   completionHandler: nil)
```

onevcat:
Thanks for reporting! I'll take a look at it soon.

onevcat:
Should be fixed in version 3.1.1. Let me know it does not work for you!

Issue #440, opened 2016-09-16
真机播放Gif崩溃


alimysoyang:
3.0.1，在真机上播放Gif时崩溃，模拟器上不会崩溃，崩溃信息如下：
* thread #1: tid = 0x20e78a, 0x0000000184a3627c vImage`vPremultipliedAlphaBlendWithPermute_ARGB8888 + 836, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x10ada4000)
    frame #0: 0x0000000184a3627c vImage`vPremultipliedAlphaBlendWithPermute_ARGB8888 + 836
    frame #1: 0x0000000184a2eb7c vImage`vImagePremultipliedAlphaBlendWithPermute_ARGB8888 + 436
    frame #2: 0x0000000184e3b7fc CoreGraphics`CGSBlend8888toARGB8888 + 284
    frame #3: 0x000000018504ef98 CoreGraphics`ARGB32_image + 4088
    frame #4: 0x0000000184fdade8 CoreGraphics`ripl_Mark + 32
    frame #5: 0x0000000184fd79f0 CoreGraphics`RIPLayerBltImage + 972
    frame #6: 0x0000000184f368e4 CoreGraphics`ripc_DrawImage + 768
    frame #7: 0x0000000184f27610 CoreGraphics`CGContextDrawImageWithOptions + 456
    frame #8: 0x0000000102a37d0c libswiftCoreGraphics.dylib`(extension in CoreGraphics):__ObjC.CGContext.draw (__ObjC.CGImage, in : __C.CGRect, byTiling : Swift.Bool) -> () + 48
  * frame #9: 0x000000010014cf38 BoBo`CGImage.fixed.getter(self=0x000000012762d890) + 240 at Image.swift:712
    frame #10: 0x000000010014cc18 BoBo`UIImage.kf_resize(size=(width = 320, height = 230), self=0x000000012764cba0) -> UIImage + 100 at Image.swift:409
    frame #11: 0x000000010014d100 BoBo`@objc UIImage.kf_resize(to : CGSize) -> UIImage + 84 at Image.swift:0
    frame #12: 0x000000010014c830 BoBo`UIImage.kf_resize(size=(width = 320, height = 320), contentMode=scaleAspectFit, self=0x000000012764cba0) -> UIImage + 164 at Image.swift:388
[...]

PR #461, opened 2016-09-28
CGImage fix seems not necessary

onevcat:
Might be related to #459 and #440


onevcat:
This will only happen in real devices. Everything goes well in simulator.

I am not sure whether there are any side effect by removing the ARGB8888 fix (which was originally added for blur images which is not ARGB8888). It seems that all tests are still green after I removed this at least.

PR #500, opened 2016-11-04
Put `open` functions in extensions into class definition

ceyhuno:
Put `open` functions in extensions into class definition to be able to override `open` functions in subclasses. Before if you tried to do it you'd gotten the Swift error: `Declarations from extensions cannot be overridden yet`.


onevcat:
Thanks for pointing it out!

Issue #505, opened 2016-11-07
Expose `KingfisherOptionsInfoItem` extension


ManueGE:
Thanks for this project. 

I see there are some helper methods for `KingfisherOptionsInfoItem` in one extension: 

````swift
extension Collection where Iterator.Element == KingfisherOptionsInfoItem
````

It provides a lot of useful `var` to get info from an Array of `KingfisherOptionsInfoItem`, for instance get the `scaleFactor` or the `transition`.

These methods are really helpful when building custom processors, so exposing this extension would be awesome.

onevcat:
Not a bad idea. Would you mind to send a p-r for this?

ManueGE:
Sure! I'll do it soon

ManueGE:
I just created the pull request #506 . Any comments are welcome.

PR #506, opened 2016-11-08
Expose extension and vars from KingfisherOptionsInfo extension

ManueGE:
In response to #505


onevcat:
@ManueGE Great! I will add some documentation for them before we could release an update.

Thanks!

PR #517, opened 2016-11-17
Fix typo `RequrstModifier` -> `RequestModifier`

cyanzhong:
I guess `RequrstModifier.swift` should be `RequestModifier.swift`, maybe I'm wrong. 😂


onevcat:
You are definitely correct on this!
Thanks.

PR #529, opened 2016-11-30
Fix preload animated image and improve performance

xspyhack:
Prepare and preload animated frames asynchronously can improve performance apparently.

While there are more than 25 visible GIFs on screen, FPS is raised from 30 to ≈60.


onevcat:
LGTM!

PR #540, opened 2016-12-14
DefaultCacheSerializer now returns original Data for an Image if it's provided

ceyhuno:
Turns out converting a Data to UIImage and then converting it back to Data using UIImageJPEGRepresentation multiplies it's size nearly by 4. So the image ends up taking up 4 times more space when written to disk or kept in memory.


onevcat:
@ceyhuno Thanks for the p-r.

It is intended to store the data converted from image instead of the original data. The original data is the data downloaded from network, without applying any processor. Storing that data directly could work if you are not using any image processor. However, if you apply a processor on the original image, you definitely want to store the processed image instead, so there is no need to do the processing again when retrieving it from cache.

PR #560, opened 2017-01-16
Make Image representation methods public

daneden:
Attempting to access either `kf.gifRepresentation()` or `kf.animatedImageData` in my project was throwing an `inaccessible due to 'internal' protection level` error. Explicitly making those methods public fixes the issue.


onevcat:
Cool. It should be no harm to expose them. And I will add documentation for them later.

Issue #559, opened 2017-01-16
maxCachePeriodInSecond to unlimited


jovanpreet:
Hi @onevcat currently my app has two types of images. 

1. The one that comes on news feeds. For them image cache of 7 days is good.

2. Images that come in chat. Until user logout and login with different id, i need those images to be cached i.e. to clear disk cache manually only. 

How can i achieve this?

onevcat:
Instead of using the default cache, you could setup and use two different caches (`ImageCache`), one for each.

See the [cheat sheet](https://github.com/onevcat/Kingfisher/wiki/Cheat-Sheet#use-customized-downloader-and-cache-instead-of-the-default-ones) for more.

jovanpreet:
I already saw that. My issue wasn't that. I just wanna know how to implement maxCachePeriodInSecond = infinity //Unlimited.
just like maxDiskCacheSize can be set to 0 for unlimited storage size

onevcat:
Oh,

There is no such "infinity" date currently. Maybe you could set `maxCachePeriodInSecond` to a relatively large number, like "300 * 50 * 60 * 60 * 24 * 7" (maybe 300 years?)

jovanpreet:
Okay thanks i would implement that.

Just wanna ask, are you planning to implement this in future?

And one more thing. Is there a callback of images that are going to be deleted? I mean just the image keys might work

jovanpreet:
@onevcat Please help

onevcat:
Sure. It could be clear to add a `forever` option for the cache time. I will do it later.

And there is a `KingfisherDidCleanDiskCache` notification you could use for the images. You can find the API reference [here](http://cocoadocs.org/docsets/Kingfisher/3.2.4/Extensions/Notification.html#/s:ZvE10KingfisherVCSo14NSNotification4Name27KingfisherDidCleanDiskCacheS1_).

jovanpreet:
What if i added this notification on a controller and kingfisher perform cleanup when that controller is not ye initialized. then?
Do i have to implement this on every controller(bad idea).
Please provide a example with best use case.

onevcat:
Your selector will not get called if the instance (target) is not there.
You'd like to add that to a long life time object, for example your app delegate or your very first tab view controller which will exist across the whole app life time.

jovanpreet:
Ok thank you.
Will disturb you again if i got stuck in any other problem...:p

PR #566, opened 2017-01-19
Support unlimited disk cache time

onevcat:
For #559

### DIFF
--- a/Sources/ImageCache.swift
+++ b/Sources/ImageCache.swift
@@ -93,6 +93,7 @@ open class ImageCache {
     
     /// The longest time duration in second of the cache being stored in disk. 
     /// Default is 1 week (60 * 60 * 24 * 7 seconds).
+    /// Set this value to 0 or negative means the disk cache will never expire.
     open var maxCachePeriodInSecond: TimeInterval = 60 * 60 * 24 * 7 //Cache exists for 1 week
     
     /// The largest disk size can be taken for the cache. It is the total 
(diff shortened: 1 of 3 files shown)


Sources/ImageCache.swift
```diff
@@ -93,6 +93,7 @@ open class ImageCache {
     
     /// The longest time duration in second of the cache being stored in disk. 
     /// Default is 1 week (60 * 60 * 24 * 7 seconds).
+    /// Setting this to a negative value will make the disk cache never expiring.
```
yonaskolb: setting this to 0 should have the same effect as negative

Sources/ImageCache.swift
```diff
@@ -475,7 +476,7 @@ open class ImageCache {
         
         let diskCacheURL = URL(fileURLWithPath: diskCachePath)
         let resourceKeys: Set<URLResourceKey> = [.isDirectoryKey, .contentAccessDateKey, .totalFileAllocatedSizeKey]
-        let expiredDate = Date(timeIntervalSinceNow: -maxCachePeriodInSecond)
+        let expiredDate = (maxCachePeriodInSecond < 0) ? nil : Date(timeIntervalSinceNow: -maxCachePeriodInSecond)
```
yonaskolb: <= 0

Sources/ImageCache.swift
onevcat: It's intended to have only negative number for "forever". 0 means expiring immediately. And since there is a default non-zero value here, it is better to opt out 0 so we could avoid to introduce misunderstanding.

PR #582, opened 2017-02-06
Ignore progressBlock if imageUrl is not equal to current imageUrl.

Fnoz:
For the case 'setImage twice for different url, and progressBlock will cross, like A1%, A50%, B20%, A100%, B60% ...'.

### DIFF
--- a/Demo/Kingfisher-Demo/ViewController.swift
+++ b/Demo/Kingfisher-Demo/ViewController.swift
@@ -71,7 +71,7 @@ extension ViewController {
         _ = (cell as! CollectionViewCell).cellImageView.kf.setImage(with: url,
                                            placeholder: nil,
                                            options: [.transition(ImageTransition.fade(1))],
-                                           progressBlock: { receivedSize, totalSize in
+                                           progressBlock: { receivedSize, totalSize, imageURL in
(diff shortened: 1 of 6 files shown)


onevcat:
Thanks for the pull request. I noticed this issue and it's great you are trying to fix it. 👍 

But it turns that this p-r would be a breaking change of existing API, so it does not seems that it could be merged currently.

Instead of ignoring and not calling the progress callback, is it possible to just cancel the current downloading task and make sure only one downloading running at a time? I believe it could be a clearer and correct way to solve it.

It could be a bit complicated since there is now a mechanism that multiple image views could share a same downloading task. Since the cancelling would break current downloading process, so if you are not sure about it, feel free to just leave it to me. I will give it a fix soon in this week I hope.

Thanks again for considering contributing.

 :)

Fnoz:
好的，我这么改也是当做一个临时修改，我认为应该这样修改：
在setImage的时候，removePreviousBlock而不是cancel，这样之前的请求的数据不至于全部浪费了，要是能处理好断点下载，那就更好了。
我是想给imageTask加个identifier，removePreviousBlock的时候，删除fetchLoad.contents中对应identifier的内容。

PR #583, opened 2017-02-06
Ignore progressBlock if imageUrl is not equal to current imageUrl.

Fnoz:
For the case 'setImage twice for different url, and progressBlock will cross, like A1%, A50%, B20%, A100%, B60% ...'.


onevcat:
LGTM.

It's a good enough temporary workaround for now!

Issue #492, opened 2016-10-31
Display gif as non animated image by default


gkye:
Is it possible to opt out of displaying animated images by default? (its causing some memory issues). I only want to display the animated image using `AnimatedImageView` when the user chooses to do so.

onevcat:
Currently no. Maybe we could implement an option to load first several frames for a GIF in a later version.

kubajakowski:
besides memory issues it's also related to https://github.com/onevcat/Kingfisher/issues/375

onevcat:
Should be added in version 3.4.0. Use `.onlyLoadFirstFrame` for that purpose.

Issue #375, opened 2016-07-28
one-frame gif disappears on UINavigationController transition


kubajakowski:
GIFs are loaded using  UIImage.animatedImageWithImages when preloadAll flag is true. For GIFs this flag is always true and there's no way to opt-out.

```
//somewhere in kf_setImageWithResource
if shouldPreloadAllGIF() {
    options.append(.PreloadAllGIFData)
}
```

```
extension ImageView {
    func shouldPreloadAllGIF() -> Bool {
        return true
    }
}
```

Now when we have such image, it disappears when i.e UINavigationController transition is performed (easy to check on swipe-to-back gesture).
When GIF contains only one frame, Image should be created as regular image (without animation, just like when preloadAll flag is set to false) to prevent this disappearing.

onevcat:
Now you can use `onlyLoadFirstFrame` to load the first frame of a GIF.

PR #591, opened 2017-02-10
Feature gif first image

onevcat:
This pull-request implements the feature of only loading the first frame of a GIF file.

It could help to save memory when loading a list (table view) of GIF file. Originally, all the GIFs will be loaded and animated. However, it is not a good idea to do so, since it will eat quite a lot of memory. Instead, we may want to load and decode the first frame of a GIF file as a preview or thumbnail. Then user could be navigated to a detail view will full, animating GIF image.


onevcat:
This should fix #492 and #375

PR #599, opened 2017-02-21
Adding Aspect Fit/Fill to Image Resizing Processor

BrikerMan:


Sources/ImageProcessor.swift
```diff
@@ -172,20 +173,38 @@ public struct ResizingImageProcessor: ImageProcessor {
...
     public let targetSize: CGSize
     
+    
+    public enum ContentMode {
+        case none // This will be the default one
+        case aspectFit
+        case aspectFill
+    }
+    /// Target content mode of output image should be.
+    public let contentMode: UIViewContentMode
```
onevcat: This will change `ImageProcessor` to be UIKit only. We may want to keep it working on macOS too. So I suggest to not introducing `UIViewContentMode` here. We need to do change the `resize(to size: CGSize, for contentMode: UIViewContentMode)` implantation a bit too.

onevcat:
@BrikerMan Could you see my comment and do some refactoring to the existing `resize(to size: CGSize, for contentMode: UIViewContentMode)`, making it something like `resize(to size: CGSize, for contentMode: ContentMode)` so we could use it in both iOS and macOS?

Thanks!

Sources/ImageProcessor.swift
BrikerMan: ok~ will try.

onevcat:
@BrikerMan Good to see this.

Since a similar p-r #600 was merged, I am going to close this one.

Thanks for your time and considering to contributing! 👍

PR #600, opened 2017-02-21
Allow ResizingImageProcessor to resize image using size proxy aspectFit & aspectFill

Drusy:
Default behavior doesn't change, using ContentMode.none as a default value

### DIFF
--- a/Sources/ImageProcessor.swift
+++ b/Sources/ImageProcessor.swift
@@ -191,7 +191,12 @@ public struct ResizingImageProcessor: ImageProcessor {
     public init(targetSize: CGSize, contentMode: ContentMode = .none) {
         self.targetSize = targetSize
         self.targetContentMode = contentMode
-        self.identifier = "com.onevcat.Kingfisher.ResizingImageProcessor(\(targetSize), \(targetContentMode))"
+        
+        if contentMode == .none {
+            self.identifier = "com.onevcat.Kingfisher.ResizingImageProcessor(\(targetSize))"
+        } else {
+            self.identifier = "com.onevcat.Kingfisher.ResizingImageProcessor(\(targetSize), \(contentMode))"
+        }
     }
     
     public func process(item: ImageProcessItem, options: KingfisherOptionsInfo) -> Image? {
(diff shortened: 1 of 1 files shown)


Sources/ImageProcessor.swift
```diff
@@ -167,25 +167,48 @@ public struct RoundCornerImageProcessor: ImageProcessor {
...
     /// - parameter targetSize: Target size of output image should be.
+    /// - parameter contentMode: Target content mode of output image should be.
     ///
     /// - returns: An initialized `ResizingImageProcessor`.
-    public init(targetSize: CGSize) {
+    public init(targetSize: CGSize, contentMode: ContentMode = .none) {
         self.targetSize = targetSize
-        self.identifier = "com.onevcat.Kingfisher.ResizingImageProcessor(\(targetSize))"
+        self.targetContentMode = contentMode
+        self.identifier = "com.onevcat.Kingfisher.ResizingImageProcessor(\(targetSize), \(targetContentMode))"
```
onevcat: `identifier` is used to identify the post-processed images in cache, so we need to keep the `identifier` the `same` for `.none` case.

Drusy: See better solution than this one ?

```
        if contentMode == .none {
            self.identifier = "com.onevcat.Kingfisher.ResizingImageProcessor(\(targetSize))"
        } else {
            self.identifier = "com.onevcat.Kingfisher.ResizingImageProcessor(\(targetSize), \(contentMode))"
        }
```

onevcat: Not bad to me.

onevcat: We could refactor this if necessary later. Let's make the tests green first.

Drusy:
@onevcat Up to date

onevcat:
LGTM

Drusy:
@onevcat Do you plan adding some test to test the content modes ?
I tried to look at the test on `ImageProcessor` but it looks more complicated than I expected.

Also, I am missing the linked images such as `kingfisher.jpg`, `onevcat.jpg`, etc. These images are not in the project, what am I missing ?

onevcat:
@Drusy Sorry for the unclear test flow. :(

The test images has been moved to another project to keep this main repo slim recently. You could run `setup.sh` to add these test files to the project (which clones the test repo to local). I am now be back to my computer so I think I could add some tests for them very quickly.

I will add the instruction to setup test to README too.

Drusy:
@onevcat Well I let you handle this so :)
Thanks for this clarifications, will still be useful for a future pull request (tested this time) !

Continue great work on this great framework 🥇

PR #630, opened 2017-03-26
Implement crop processor

onevcat:


Sources/ImageProcessor.swift
```diff
@@ -353,19 +412,86 @@ public struct ColorControlsProcessor: ImageProcessor {
...
+    /// - Note:
+    ///   The anchor point is consisted by two values between 0.0 and 1.0.
+    ///   It indicates a related point in current image, eg: (0.0, 0.0) for top-left
+    ///   corner, (0.5, 0.5) for center and (1.0, 1.0) for bottom-right corner.
+    ///   The `size` property of `CroppingImageProcessor` will be used along with
+    ///   `anchor` to calculate a target rectange in the size of image.
+    ///    
+    ///   The target size will be automatically calculated with a reasonable behavior.
+    ///   For example, when you have an image size of `CGSize(width: 100, height: 100)`,
+    ///   and a target size of `CGSize(width: 100, height: 100)`, with a (0.0, 0.0) anchor (top-left),
```
raxityo: I think we want to say:
> a target size of `CGSize(width: 20, height: 20)`

onevcat: Yes you are definitely right! Thanks for pointing it out.

PR #633, opened 2017-03-29
fix wrong size

awuu:
when both set backgroundDecode and scaleFactor, the size is wrong and the final image is wrong too


onevcat:
It makes sense. Thanks.

PR #639, opened 2017-03-31
Added support for adding an image to the highlighted image property

MadsBogeskov:
I have added support for adding images to the `highlighted` property on a `UIImageView`

### DIFF
--- a/Sources/ImageView+Kingfisher.swift
+++ b/Sources/ImageView+Kingfisher.swift
@@ -36,17 +36,19 @@ import UIKit
  *	Set image to use from web.
  */
 extension Kingfisher where Base: ImageView {
+
     /**
      Set an image with a resource, a placeholder image, options, progress handler and completion handler.
-     
+
      - parameter resource:          Resource object contains information such as `cacheKey` and `downloadURL`.
+     - parameter on:                Closure that assigns the final image to a property on the imageview
      - parameter placeholder:       A placeholder image when retrieving the image at URL.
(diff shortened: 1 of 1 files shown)


onevcat:
@MadsBogeskov Thanks for considering to contribute. But I am afraid I cannot merge this due to these two consideration:

1. Setting a highlight image is not a regular operation, so I wonder whether it is worth to have an extension method on it. It somehow added complexity to the framework and its users, but could only benefit very limited or few developers. 

2. The `ImageView` extension contains  a web url associated object, to keep track of current loading url. In your implementation. When downloading finished, Kingfisher will try to check the url to set whether it should actually set that image or not (since there might be multiple setting invocation before an image downloading finished). In this p-r, the url will be share between normal state images and highlight images. It would cause problem if you try to set a normal image and a highlighted image at the same time.

To use Kingfisher for highlighted image property of an image view, it should be not hard to use the `KingfisherManager`'s [`retrieve` method](http://cocoadocs.org/docsets/Kingfisher/3.6.0/Classes/KingfisherManager.html#/s:FC10Kingfisher17KingfisherManager13retrieveImageFT4withPS_8Resource_7optionsGSqGSaOS_25KingfisherOptionsInfoItem__13progressBlockGSqFTVs5Int64S3__T__17completionHandlerGSqFTGSqCSo7UIImage_GSqCSo7NSError_OS_9CacheTypeGSqV10Foundation3URL__T___CS_17RetrieveImageTask) and then set the highlight image in completion block. The downloading and cache will also be handled by using the manager.

Thank you for your time and this great p-r. I'll keep on eye to this feature too to see whether we have a better choice on it.

Issue #663, opened 2017-05-08
Rename KingfisherOptionsInfoItem preloadAllGIFData


ShaharHD:
`KingfisherOptionsInfoItem.preloadAllGIFData` is used in the code to indicate for backwards compatibility the need for reading the entire GIF to memory.

I'm suggesting, in order to allow for generic support for animated image formats to rename it to `KingfisherOptionsInfoItem.preloadAllAnimationData` and it's relevant counterparts (like `Collection.preloadAllGIFData`, while having the fallback to the old `.preloadAllGIFData` when processing options using an alias.

If I have a green light for this, I'll create the relevant PR for it.

onevcat:
@ShaharHD Yes, I think that is a good direction. Thanks for it.

ShaharHD:
@onevcat When trying to run the tests on Xcode I'm getting errors on `Kingfisher/Kingfisher-TestImages/Modified/Overlay/onevcat-overlay-red-07-mac.jpg: No such file or directory` Are some files missing from the git repo?

ShaharHD:
Fixed with #664

PR #664, opened 2017-05-08
migrate to preloadAllAnimationData

ShaharHD:
Rename all instances of `preloadAllGIFData` to `preloadAllAnimationData` with proper @available notices.

References #663


ShaharHD:
@onevcat Please note that the initial `pod install` changed all the files I've put in the 1st commit (didn't touch anything)
All my changes afterwards are in the 2nd commit.

onevcat:
Thanks!
It was a travis issue and I'll restart the tests.

Sources/KingfisherOptionsInfo.swift
```diff
@@ -228,10 +232,16 @@ public extension Collection where Iterator.Element == KingfisherOptionsInfoItem
...
+    /// Whether the image data should be all loaded at once if it is an animated image.
+    public var preloadAllAnimationData: Bool {
+        return contains { $0 <== .preloadAllAnimationData }
+    }
+
     /// Whether the image data should be all loaded at once if it is a GIF.
+    @available(*, deprecated, renamed: "preloadAllAnimation")
     public var preloadAllGIFData: Bool {
-        return contains { $0 <== .preloadAllGIFData }
+        return contains { $0 <== .preloadAllAnimationData }
```
onevcat: Maybe we could just return `preloadAllAnimationData`?

ShaharHD:
Worked.
Two FYIs:
1. Version `3.7.0` was never pushed to cocoapods
2. I would appreciate if you can push `3.7.1` also with my fix after you approve it so I can base `KingfisherWebP` on the latest while I'm implementing the proper animation (right now I have it pointed to my branch for development)

Sources/Image.swift
```diff
@@ -323,15 +323,20 @@ extension Kingfisher where Base: Image {
...
             case .unknown:
                 image = Image(data: data, scale: scale)
             }
         #endif
-        
+
         return image
     }
+
+    @available(*, deprecated, message: "preloadAllGIFData is now preloadAllAnimationData")
```
onevcat: Since it is an internal API, we could just remove it instead of marking it as deprecated.

Sources/KingfisherOptionsInfo.swift
```diff
@@ -84,13 +84,17 @@ public enum KingfisherOptionsInfoItem {
...
+    /// Whether all the animated image data should be preloaded. Default it false, which means following frames will be
+    /// loaded on need. If true, all the animated image data will be loaded and decoded into memory. This option is mainly
     /// used for back compatibility internally. You should not set it directly. `AnimatedImageView` will not preload
     /// all data, while a normal image view (`UIImageView` or `NSImageView`) will load all data. Choose to use
     /// corresponding image view type instead of setting this option.
-    case preloadAllGIFData
+    case preloadAllAnimationData
+
+    /// Back compatibility for deprecated preloadAllGIFData
+    @available(*, deprecated, renamed: "preloadAllAnimation")
```
onevcat: Could you move the `deprecated` statements to an extra extension at the end of this file (like what is done [here](https://github.com/onevcat/Kingfisher/blob/master/Sources/Image.swift#L867-L974))?

onevcat:
> Version 3.7.0 was never pushed to cocoapods

Oops, didn't notice it. It should since the release script runs without issue. I guess it might be a CocoaPods issue.

> I would appreciate if you can push 3.7.1 also with my fix after you approve it so I can base KingfisherWebP on the latest while I'm implementing the proper animation (right now I have it pointed to my branch for development)

Sure. I'll release 3.7.1 once it prepared.

Sources/KingfisherOptionsInfo.swift
ShaharHD: The difference here it's a enum and not functions. But I'll fix the other ones as requested.

onevcat: It should be fine to have this static property in an extension (as long as the extension is `public`, it should also supply good back compatibility).

ShaharHD: I don't understand how you suggest having the enum value deprecated outside of the enum declaration (Swift AFAIK does not support adding to an enum through inheritance or any other extension mechanism)

onevcat: Here you are using a static property to "simulate" the syntax of an enum case:

```
@available(*, deprecated, renamed: "preloadAllAnimation")
static let preloadAllGIFData = KingfisherOptionsInfoItem.preloadAllAnimationData
```

So it is possible to move it to extension (it is not an enum member anymore). Sure you could keep it as an enum member (in which case you cannot move it to an extension), but I think it is a clever way to make it a static property here.

In there, this should also compile:

```swift
public extension KingfisherOptionsInfoItem {
    @available(*, deprecated, renamed: "preloadAllAnimation")
static let preloadAllGIFData = KingfisherOptionsInfoItem.preloadAllAnimationData
}
```

And users could get a rename warning if it is used.

ShaharHD: Tried the same, but with incorrect syntax is seems... Thanks 👍

Sources/KingfisherOptionsInfo.swift
```diff
@@ -304,3 +294,18 @@ public extension Collection where Iterator.Element == KingfisherOptionsInfoItem
...
 }
+
+// MARK: - Deprecated. Only for back compatibility.
+public extension Collection where Iterator.Element == KingfisherOptionsInfoItem {
+
+    /// Whether the image data should be all loaded at once if it is a GIF.
+    @available(*, deprecated, renamed: "preloadAllAnimation")
+    public var preloadAllGIFData: KingfisherOptionsInfoItem {
+        return .preloadAllAnimationData
+    }
```
onevcat: This should return a `Bool` with implementation like this:

```swift
@available(*, deprecated, renamed: "preloadAllAnimation")
public var preloadAllGIFData: Bool {
    return preloadAllAnimationData
}
```

onevcat:
@ShaharHD 

Great. I added another comment. Please fix that too before we could finally merge it.

Sources/KingfisherOptionsInfo.swift
ShaharHD: My bad a typo of copy & paste.
Fixed

PR #713, opened 2017-06-18
Call progressBlock when image is coming from cache.

math-nao:
ProgressBlock handler is not called when image is coming from cache. This pull request fixes this issue.

### DIFF
--- a/Sources/KingfisherManager.swift
+++ b/Sources/KingfisherManager.swift
@@ -204,6 +204,11 @@ public class KingfisherManager {
                                         options: KingfisherOptionsInfo)
     {
         
+        let triggerProgressFromCache = {
+            // image comes from cache, so let's trigger progressBlock
+            let receivedSize: Int64 = 1, totalSize: Int64 = 1
+            progressBlock?(receivedSize, totalSize)
+        };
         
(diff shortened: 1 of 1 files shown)


onevcat:
I am not sure we need this or not. In fact, there is no such "progress" when loading an image from disk. Furthermore, it does not make sense to have its size as "1". Progress calling back is mainly for displaying an indicator with percentage or remaining size value. However, it should be not needed when loading from disk.

@math-nao Thanks for contributing. Is there any good reason to also call the progress block when loading from disk?

math-nao:
I'm using it for handling a progress UIView. It works fine when image is downloaded but not when image is coming from cache; currently, there is no way to know that image is coming from cache in `retrieveImage ` method so I cannot handle this case for my needs.

I choose `receivedSize == totalSize` with a value of `1` so progress is done at 100%. I currently do not use numerical value, just percent. But of course, we could also use image size value instead to fulfill requirements.

onevcat:
There should be a `cacheType` parameter in the completion handler for the `retrieveImage` method. Could that be a help?

math-nao:
Yes, `cacheType` parameter fit my needs. Thanks.

This pull requests can be closed.

PR #741, opened 2017-07-29
Update cache result API

onevcat:


onevcat:
Oops!

onevcat:
The typo should be also fixed! Thank you.

Issue #752, opened 2017-08-24
Xcode 9, Carthage & Kingfisher compatibility


tizaks:
### Check List

Thanks for considering to open an issue. Before you submit your issue, please confirm these boxes are checked.

- [x] I have read the [wiki page](https://github.com/onevcat/Kingfisher/wiki) and [cheat sheet](https://github.com/onevcat/Kingfisher/wiki/Cheat-Sheet), but there is no information I need.
- [x] I have searched in [existing issues](https://github.com/onevcat/Kingfisher/issues?utf8=✓&q=is%3Aissue), but did find a same one.
- [x] I want to reoprt a problem instead of asking a question. It'd better to use [kingfisher tag in Stack Overflow](http://stackoverflow.com/questions/tagged/kingfisher) to ask a question.

#### What

I've been working with the Xcode 9 beta and uploading builds to TestFlight for testing, and ran into a new issue as soon as I added Kingfisher as a dependency to my project using Carthage. It seems that in Xcode 9, if the Test action of a scheme has "Gather coverage data" enabled, it will include that profiling instrumentation in the build framework that Carthage generates. Unfortunately, iTunes Connect will reject builds that include this instrumentation which means in it's current state people using Carthage won't be able to submit their apps using Kingfisher.

The Carthage team has decided they are unable to fix this issue on their side (see https://github.com/Carthage/Carthage/issues/2056) and Apple has confirmed this is expected behaviour for Xcode 9 (https://forums.developer.apple.com/thread/81893).

My suggestion would be to disable the Gather coverage data option in the scheme's test action - I understand this is probably a difficult choice to make as having that coverage data is useful when testing code, but I imagine this will become a common issue for many people as soon as Xcode 9 is properly released.

I'm happy to submit a pull request to the `swift4` branch with this small change if you agree with my proposal.

Thanks!

onevcat:
@tizaks Oh, I didn't know that! Yes, please open a pull request for it. I am glad to merge it.

Thank you!

tizaks:
Thanks! Fixed in #753

PR #753, opened 2017-08-24
Disable gather coverage data in scheme test action

tizaks:
As discussed in #752, disable "Gather coverage data" in the test action of all schemes in order to resolve issues with Kingfisher being used in Xcode 9 apps when using Carthage to compile the framework.


onevcat:
Thanks!

PR #756, opened 2017-08-29
Teameh custom placeholder

onevcat:


Sources/Kingfisher.swift
```diff
@@ -30,12 +30,14 @@ import ImageIO
...
     import AppKit
     public typealias Image = NSImage
+    public typealias View = NSView
     public typealias Color = NSColor
     public typealias ImageView = NSImageView
     typealias Button = NSButton
 #else
     import UIKit
     public typealias Image = UIImage
+    public typealias View = UIView
```
teameh: Check!

Sources/Placeholder.swift
```diff
@@ -0,0 +1,61 @@
...
+}
+
+extension Placeholder where Self: Image {
+    public func add(to imageView: ImageView) { imageView.image = self }
+    public func remove(from imageView: ImageView) { imageView.image = nil }
+}
+
+extension Image: Placeholder {}
+
+extension Placeholder where Self: View {
```
teameh: I still don't get why this is better than `extension Placeholder : View {...}`. Could you explain it if you have the time for it. Nevermind otherwise :)

teameh:
Nice 😃

Sources/Placeholder.swift
onevcat: @teameh 

By `extension View: Placeholder {...}`, you will make ALL subclasses of UIView to be a `Placeholder`. It could be too wide. While by using `extension Placeholder where Self: View`, only those ones declared as `extension MyView: Placeholder {}` could be used as a placeholder. 

I would prefer to keep it minimal, and let users to specify the conformance only when necessary.

teameh: Check. That makes sense!

PR #773, opened 2017-09-20
Implemented data download callback

Ashok28:
This adds possibility to further manipulate downloaded raw image data. Can be used i.e. for decrypting images.

### DIFF
--- a/Sources/ImageDownloader.swift
+++ b/Sources/ImageDownloader.swift
@@ -132,6 +132,18 @@ public protocol ImageDownloaderDelegate: class {
             you can implement this method to change that behavior.
     */
     func isValidStatusCode(_ code: Int, for downloader: ImageDownloader) -> Bool
+    
+    /**
+     Called when the `ImageDownloader` object successfully downloaded image data from specified URL.
+     
+     - parameter downloader: The `ImageDownloader` object finishes data downloading.
+     - parameter data:       Downloaded data.
+     - parameter url:        URL of the original request URL.
+     
+     - Note: This callback can be used to preprocess raw image data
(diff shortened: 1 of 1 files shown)


Sources/ImageDownloader.swift
```diff
@@ -141,6 +153,9 @@ extension ImageDownloaderDelegate {
     public func isValidStatusCode(_ code: Int, for downloader: ImageDownloader) -> Bool {
         return (200..<400).contains(code)
     }
+    public func imageDownloader(_ downloader: ImageDownloader, didDownload data: Data, for url: URL) -> Data? {
+        return nil
```
onevcat: Maybe we should just return the original data instead of `nil`? Any reason to return a `nil` here?

onevcat:
@Ashok28 Thanks for the commit. It could be useful and please take a look at the review comment above. Thank you!

Sources/ImageDownloader.swift
Ashok28: You're right, there's no particular reason, to return nil. I've pushed updated version.

onevcat:
And I guess it would be better to interrupt the process and call completion handler with an error if `nil` is returned in this delegate method. By this, the user could have more control on what should happen according. (Say, not showing an image and getting an error when decrypting is failing).

@Ashok28  How do you think about that?

Anyway, I will merge this for now and do some improvement based on it. Thanks!

Ashok28:
Definitely, this should have acted just as basic implementation, which I used for one specific use case, wasn't sure, if it's going to get merged. I can further improve it.

Issue #772, opened 2017-09-16
Errors in Swift 4 and Kingfisher 4.0.1


ndduong97:
### Check List

Thanks for considering to open an issue. Before you submit your issue, please confirm these boxes are checked.

- [x] I have read the [wiki page](https://github.com/onevcat/Kingfisher/wiki) and [cheat sheet](https://github.com/onevcat/Kingfisher/wiki/Cheat-Sheet), but there is no information I need.
- [x] I have searched in [existing issues](https://github.com/onevcat/Kingfisher/issues?utf8=✓&q=is%3Aissue), but did find a same one.
- [x] I want to reoprt a problem instead of asking a question. It'd better to use [kingfisher tag in Stack Overflow](http://stackoverflow.com/questions/tagged/kingfisher) to ask a question.

### Issue Description
Errors from methods changing.

#### What
After updating to version 4.0.1, I'm still getting these errors:

`'cropped(to:)' has been renamed to 'cropping(to:)'`
`'applyingFilter(_:parameters:)' has been renamed to 'applyingFilter(_:withInputParameters:)'`

#### Reproduce
The pod cannot be built with Xcode 9 GM Seed.

onevcat:
Kingfisher 4 is for Swift 4. If you are using Swift 3.2, please roll back to Kingfisher 3.

ndduong97:
Sorry, I am using Swift 4

onevcat:
Please check:

1. The Swift Version setting of Kingfisher is set to Swift 4.
2. If you have a .swift-version file under your project, make sure the content of it is "4.0".
3. Try clean your project.

If the problem is still there, I suggest to submit it to CocoaPods, since there should be no problem in Kingfisher for it, since we are already using it in our Swift 4 project for a while.

ndduong97:
Thanks I fixed it by changing the language to Swift 4 for Kingfisher.

onevcat:
Version 4.1.0 should fixed this kind of problems (mismatching versions with Kingfisher code and build setting when used with CocoaPods)

Ref: #781

PR #777, opened 2017-09-26
Fix xcode reported Swift 4 errors

jasonivers:
xcode project failed to build and auto-fixed these errors.  This is just the fixed file (that actually would build).


onevcat:
Plesae make sure you are using Swift 4.

See https://github.com/onevcat/Kingfisher/issues/772

PR #794, opened 2017-10-20
Support caching original images on a different cache

yigitcanyurtsever:
Hey hey 👋,

Here at Zeplin, we needed a way to cache the original images on a different cache (rather than the target cache) when using a custom image processor. For example, we are fetching images with a custom thumbnail image processor, and we want to store these thumbnails on a 'thumbnail' cache, and store the originals on a 'snapshots' cache. 

To accomplish this, we wrote a `originalCache` to `KingfisherOptionsInfoItem` on our fork. `originalCache` is basically used for storing and retrieving original images if paired with a non-default image processor. It defaults to `targetCache` to make it an optional item.

I've also included a typo fix and an optimization to prevent double caching if `cacheOriginalImage` is selected when used with the default image processor.

Thought this may be a good addition to the main repo. I wasn't able to write or run any tests because of the missing `Kingfisher-TestImages` folder.

### DIFF
--- a/Sources/KingfisherManager.swift
+++ b/Sources/KingfisherManager.swift
@@ -171,7 +171,7 @@ public class KingfisherManager {
                                       cacheSerializer: options.cacheSerializer,
                                       toDisk: !options.cacheMemoryOnly,
                                       completionHandler: nil)
-                    if options.cacheOriginalImage {
+                    if options.cacheOriginalImage && options.processor != DefaultImageProcessor.default {
                         let originalCache = options.originalCache
                         let defaultProcessor = DefaultImageProcessor.default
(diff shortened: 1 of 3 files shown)


onevcat:
@yigitcanyurtsever Thanks for it. It looks promising. I would take a look at it later.

BTW, for the testing images, you could take a look at https://github.com/onevcat/Kingfisher#developments-and-tests

yigitcanyurtsever:
@onevcat Thanks for looking at it! I've updated the PR with some tests. 

By the way, I didn't know that it was recommended to open a discussion first – sorry about that.

onevcat:
@yigitcanyurtsever No problem, bro!

The code looks very nice and I am so glad to merge it. I will prepare an update for it now.
Thank you!

PR #815, opened 2017-11-28
[WIP] add ImageModifier

### LINKED ISSUE #810: Allow exclusive use of .cacheOriginalImage file instead of final file
### Check List

### Issue Description
UIButtons of type .system use a property called `imageRenderingMode` of their image to determine if the image should be used in template or original mode.
I have created a custom ImageProcessor, `RenderingModeImageProcessor`, to set the rendering mode of an image.
I am trying use Kingfisher with a UIButton as follows:
```
let myProcessor = RoundCornerImageProcessor(cornerRadius: 18, targetSize: CGSize(width: 36, height: 36), roundingCorners: .all, backgroundColor: nil) >> RenderingModeImageProcessor(renderingMode: .alwaysOriginal)

button.kf.setImage(with: imageURL, for: .normal, placeholder: defaultIcon, options: [.processor(myProcessor), .scaleFactor(3.0), .cacheOriginalImage])
```

#### What

The first time the image is downloaded, everything works perfectly. My image is rounded and the rendering mode is set to .alwaysOriginal by my custom image processor.
When the image is taken from the cache, the processors are not rerun, even though the original image is cached, because the "final image" is also cached.

#### Reproduce

Create a RenderingModeImageProcessor, such as the one in the code block below (my code, MIT License):
```
 public struct RenderingModeImageProcessor: ImageProcessor {

    /// Identifier of the processor.
    /// - Note: See documentation of `ImageProcessor` protocol for more.
    public let identifier: String

    /// The rendering mode to use.
    /// Defaults to .alwaysOriginal
    public let renderingMode: UIImageRenderingMode

    /// Initialize a `RenderingModeImageProcessor`.
    ///
    /// - Parameters:
    ///   - renderingModel: The rendering mode to use.
    public init(renderingMode: UIImageRenderingMode = .alwaysOriginal) {
        self.renderingMode = renderingMode

        self.identifier = "com.identifier.here.RenderingModeImageProcessor(\(renderingMode))"
    }

    /// Process an input `ImageProcessItem` item to an image for this processor.
    ///
    /// - parameter item:    Input item which will be processed by `self`
    /// - parameter options: Options when processing the item.
    ///
    /// - returns: The processed image.
    ///
    /// - Note: See documentation of `ImageProcessor` protocol for more.
    public func process(item: ImageProcessItem, options: KingfisherOptionsInfo) -> Image? {
        switch item {
        case .image(let image):
            return image.withRenderingMode(renderingMode)
        case .data(_):
            return (DefaultImageProcessor.default >> self).process(item: item, options: options)
        }
    }
}
```

Use the ImageProcessor above with an image on a UIButton.

#### Other Comment

When using the option .forceRefresh, the processor gets a chance to run every time. This is a bad workaround though because it completely avoids caching.

#### Suggested Fix

Add an option to avoid using the final image, or to just not cache the final image
OR
Add an option to run processors even when the image is cached.
OR
A combination of both of the above.

If this problem is recognized, I can work on a PR if necessary.

ethansinjin:
Resolves #810

This adds an ImageModifier, allowing edits to be performed on a concrete type of `Image` (ex. UIImage) right before the image is used/displayed.

I've created the `KingfisherOptionsInfoItem` case and `ImageModifier` struct, but the actual modifier isn't being run yet. I'd like advice on where exactly to run the modifier. I think it should go in `downloadAndCacheImage` and `tryToRetrieveImageFromCache`, but I'm not sure; there might be a better place for it.

TODO:

- [x] add ImageModifierTests
- [ ] execute the modifier's `modify()` method somewhere in `KingfisherManager.swift`

### DIFF
--- a/Kingfisher.xcodeproj/project.pbxproj
+++ b/Kingfisher.xcodeproj/project.pbxproj
@@ -335,6 +335,10 @@
 		D9638BA61C7DC71F0046523D /* ImagePrefetcherTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = D9638BA41C7DC71F0046523D /* ImagePrefetcherTests.swift */; };
 		D9638BA71C7DCF560046523D /* ImagePrefetcherTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = D9638BA41C7DC71F0046523D /* ImagePrefetcherTests.swift */; };
 		D9638BA81C7DCF570046523D /* ImagePrefetcherTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = D9638BA41C7DC71F0046523D /* ImagePrefetcherTests.swift */; };
+		F78F5EBD1FCDDE42001A9111 /* ImageModifier.swift in Sources */ = {isa = PBXBuildFile; fileRef = F78F5EBC1FCDDE42001A9111 /* ImageModifier.swift */; };
+		F78F5EBE1FCDDE42001A9111 /* ImageModifier.swift in Sources */ = {isa = PBXBuildFile; fileRef = F78F5EBC1FCDDE42001A9111 /* ImageModifier.swift */; };
+		F78F5EBF1FCDDE42001A9111 /* ImageModifier.swift in Sources */ = {isa = PBXBuildFile; fileRef = F78F5EBC1FCDDE42001A9111 /* ImageModifier.swift */; };
+		F78F5EC01FCDDE42001A9111 /* ImageModifier.swift in Sources */ = {isa = PBXBuildFile; fileRef = F78F5EBC1FCDDE42001A9111 /* ImageModifier.swift */; };
 		FB402D0E1EDEAB7E002B62A1 /* FormatIndicatedCacheSerializer.swift in Sources */ = {isa = PBXBuildFile; fileRef = FB402D0D1EDEAB7E002B62A1 /* FormatIndicatedCacheSerializer.swift */; };
 		FB402D0F1EDEAB7E002B62A1 /* FormatIndicatedCacheSerializer.swift in Sources */ = {isa = PBXBuildFile; fileRef = FB402D0D1EDEAB7E002B62A1 /* FormatIndicatedCacheSerializer.swift */; };
 		FB402D101EDEAB7E002B62A1 /* FormatIndicatedCacheSerializer.swift in Sources */ = {isa = PBXBuildFile; fileRef = FB402D0D1EDEAB7E002B62A1 /* FormatIndicatedCacheSerializer.swift */; };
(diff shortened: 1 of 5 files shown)


onevcat:
@ethansinjin 

Thanks for the p-r. I have to say it is a much bigger one than I expected! IMO, it would be sufficient to just provide a single and light wrapper type to meet all the needs, just similar to the existing `RequestModifier`. Something like this:

```swift
import Foundation

/// Request modifier of image downloader.
public protocol ImageModifier {
    func modified(for image: Image) -> Image?
}

struct DefaultImageModifier: ImageModifier {
    static let `default` = DefaultImageModifier()
    private init() {}
    func modified(for image: Image) -> Image? {
        return image
    }
}

public struct AnyImageModifier: ImageModifier {
    
    let block: (Image) -> Image?
    
    public func modified(for image: Image) -> Image? {
        return block(image)
    }
    
    public init(modify: @escaping (Image) -> Image? ) {
        block = modify
    }
}
```

By passing a modifying block to `AnyImageModifier`, users could do everything they want to apply change to the image. Currently it seems that you are following the way of `ImageProcessor`, to have `identifier` and cascading of modifiers. However, the identifiers there are mainly used for purpose of data persisting and caching. It is not the case here for image modifier.

How do you think about it? Does the `identifier` take any responsibility here?

ethansinjin:
That's a fair point; I guess it is a bit over-architected. I'll rework it to be more like `RequestModifier`.

A few thoughts:

- I still think we should include a `RenderingModeImageModifier` and `FlipsForRightToLeftLayoutDirectionImageModifier` because I'm pretty sure those are the most common reasons this class is needed.
- I think the return parameter should be `Image` instead of `Image?`, because at the point of execution of the `ImageModifier`, a valid `Image` is already required as a parameter. So, if the change isn't possible, couldn't the original image just be returned?

ethansinjin:
Alright, please let me know what you think of the changes! @onevcat

onevcat:
@ethansinjin Great work!

> I still think we should include a RenderingModeImageModifier and FlipsForRightToLeftLayoutDirectionImageModifier because I'm pretty sure those are the most common reasons this class is needed.

Yes, I totally agree with you.

> I think the return parameter should be Image instead of Image?, because at the point of execution of the ImageModifier, a valid Image is already required as a parameter. So, if the change isn't possible, couldn't the original image just be returned?

The original thought is letting users have a chance to return `nil` if the do not want to set the image at the last minute. But I guess it could rarely happen so current implementation looks fine!

onevcat:
I think I could do the last part of applying the modifier later. Once it has been done, I will release an update version. Thank you!

ethansinjin:
Thanks for merging!

PR #818, opened 2017-12-03
new image processors: blend, compositing

yemodin:
Adds blending/compositing(macOS) processors. Allows easily blend/compositing.

For example, I was needed to blend unedited images which have the different background color so they will appear blended with the background color of my interface.


yemodin:
Also, I changed macOS version to 10.11 to successfully build and run.

onevcat:
@yemodin Looks great.

However, if we increase the minimal deploy target, we have to release another major update since it breaks compatibility. May I know which API you are using is 10.11 above?

And do not worry about the failing tests. If we could add this processor to the framework, I will check and fix these tests.

yemodin:
@onevcat I changed only demo project target, not framework target. Next lines of macOS demo require 10.11+:

yemodin:
Also, I added the test to each image processor. And made the pull request for image assets:
https://github.com/onevcat/Kingfisher-TestImages/pull/1

onevcat:
Oh, cool! I will check it now.

PR #820, opened 2017-12-05
make closure return type  -> () modify to -> Void. Void is more accurate than () in return.

chdzq:
闭包返回值Void要比()更精确，表明这是Void类型。而()一般作为空的表达式，也可以作为Void类型。感觉Void更精准


onevcat:
Thanks!

Issue #821, opened 2017-12-06
Can I change the implementation of MD5 into CommonCrypto?


chdzq:
### Issue Description
Can I change the implementation of MD5 into CommonCrypto?
like：
```
    var md5: String {
        if let data = self.data(using: String.Encoding.utf8) {
            let str = cString(using: .utf8)
            let strLength = CUnsignedInt(lengthOfBytes(using: .utf8))
            let digestLen = Int(CC_MD5_DIGEST_LENGTH)
            let result = UnsafeMutablePointer<CUnsignedChar>.allocate(capacity: digestLen)
            defer {
                result.deallocate(capacity: digestLen)
            }
            CC_MD5(str, strLength, result)
            var hash = ""
            for i in 0 ..< digestLen {
                hash = hash.appendingFormat("%02x", result[i])
            }
            return hash
        }
        return self
    }
```
it work correctly in cocoapod and framework.
thanks!

onevcat:
@chdzq  Thanks for considering of that!

The reason of current MD5 implementation is that in early age of Swift, CommonCrypto (or some other C library) cannot be used in a framework target (like Kingfisher), although you could import it in a bridge header and use it in an application target.

I am not sure about current state, should we add a Swift wrapper of CommonCrypto and link to it if we want to use `CC_MD5` or other C APIs? If so, I think I'd prefer to staying in current implementation instead of importing another layer of complexity.

If it is easy enough and we could do it without huge changes, I'd like to see a p-r on it and then we could see how it works.

Thank you!

PR #822, opened 2017-12-06
change the implementation of MD5 into CommonCrypto

chdzq:
[issues/821)](https://github.com/onevcat/Kingfisher/issues/821)


chdzq:
only testSSL timeout，other all passed。
like follow：
wx20171206-135030 2x

onevcat:
@chdzq Thanks! The changes looks great to me. I will restart the test and also check usability with CocoaPods and Carthage with these changes before we could merge it!

chdzq:
I only test Kingfisher for platform iOS and watchOS with CocoaPods

onevcat:
@chdzq I also tested it with Carthage and it works well!

I think there is one more thing before it could be merged. May I ask you to move the "CommonCrypto" from root folder to "Source"? Maybe it would be a better place where it should be. And both the swift include path in build setting and the pod settings in spec file needs update too.

Thank you!

chdzq:
稍等下，我发现一个更好的解决办法。可以是用podfile更改pod的的Kingfisher的Build Phases，然后添加一个脚本做这件事。但是我动态生成的script_build_phase是在最后，在Compile Sources之后。我在想办法放在它之前。

chdzq:
搞定了。但是这样需要在Podfile除了正常pod 'Kingfisher'外，还需要增加类似上述的命令。我待会分开提下。
我感觉这样会更好

chdzq:
I built a Demo demonstration how to use it.
 [Demo](https://github.com/chdzq/KingfisherTest.git)
then how update wiki for install, if the pr accepted.

Podfile add:
```
post_install do |installer|
    generate_kingfisher_frameworks_script installer
end


def generate_kingfisher_frameworks_script (installer)
    shell_script = <<-SP
mkdir -p "${BUILT_PRODUCTS_DIR}/CommonCrypto"
cat <<EOF > "${BUILT_PRODUCTS_DIR}/CommonCrypto/module.modulemap"
module CCommonCrypto [system] {
    header "${SDKROOT}/usr/include/CommonCrypto/CommonCrypto.h"
    export *
}
EOF
    SP

    Pod::UI.puts "generate script for kingfisher frameworks"
    project = installer.pods_project
    installer.pods_project.targets.each do |target|
        if /^Kingfisher*?/.match("#{target.name}")
            phase = target.new_shell_script_build_phase("[KF]CommonCrypto")
            phase.shell_script = shell_script
            target.build_phases.unshift(phase).uniq! unless target.build_phases.first == phase
            project.save()
        end
    end
end
```

chdzq:
Commit 9520b2a的是之前的工作方式。后面的是换一种增加Build Phase脚本，这样就不会在Pods中出现CommonCrypto文件夹了。

onevcat:
@chdzq Oh, great work!

Thanks for your effort on the build phase script. But I am afraid it would break CocoaPods integration for current framework users and it requires a major release. I am not quite willing to do so and it seems that this change could increase the degree of difficulty when adding this framework. Personally I prefer to use commit 9520b2a.

May I know is there any other concerns than introducing the CommonCrypto to Pods? IMO, it is totally worth to doing that, if we can keep user's Podfile clean without this post install script.

How do you think about it?

chdzq:
yes, you are right. do that when we can keep user's Podfile clean without this post install script.
i reset to commit 9520b2a.

chdzq:
also can test with  [Demo Project](https://github.com/chdzq/KingfisherTest.git) by clear post_install work, then clean xcode and delete the work DerivedData .

onevcat:
Fantastic!

chdzq:
我感觉这个相对好点处理方式是和OC混编，专门处理这个MD5

PR #823, opened 2017-12-06
Add option to remove image only from disk cache and not memory

SlinToWin:
Use Case:

New Images for new Data Models are downloaded in the background but the UI isn't refreshed yet.
So we want to keep the currently visible Images in memory cache so that e.g. a TableViewCell can still display them (from memory cache) but want to delete images from disk cache to keep the disk footprint low.


onevcat:
@SlinToWin LGTM

Could you also update the documentation above in the comment for this argument? Thanks!

Issue #356, opened 2016-07-08
Q: play GIF only once with AnimatedImageView


szuniverse:
Is it possible to play the gif only once with AnimatedImageView ?

halonsoluis:
..and also known when it has ended.

onevcat:
TODO: Add `repeatCount` and related delegate methods to `AnimatedImageView`.

JacobMao:
@halonsoluis I created a PR for this issue

onevcat:
Supported in #829

PR #829, opened 2017-12-20
Play gif only once with animated image view

JacobMao:
[Issue Link](https://github.com/onevcat/Kingfisher/issues/356)

**Todo List**
- [x] Adding Repeat Count

- [x] Adding delegate

### DIFF
--- a/Sources/AnimatedImageView.swift
+++ b/Sources/AnimatedImageView.swift
@@ -37,15 +37,24 @@ import ImageIO
 /// Protocol of `AnimatedImageView`.
 public protocol AnimatedImageViewDelegate: class {
     /**
+     Called after the animatedImageView has finished each animation loop.
+
+     - parameter imageView: The animatedImageView that is being animated.
+     - parameter count: The looped count.
+     */
+    func animatedImageView(_ imageView: AnimatedImageView, didPlayAnimationLoops count: UInt)
+
+    /**
      Called after the animatedImageView has reached the max repeat count.
 
      - parameter imageView: The animatedImageView that is being animated.
      */
-    func didFinishAnimate(for imageView: AnimatedImageView)
+    func animatedImageViewDidFinishAnimating(_ imageView: AnimatedImageView)
 }
 
 extension AnimatedImageViewDelegate {
-    public func didFinishAnimate(for imageView: AnimatedImageView) {}
+    public func animatedImageView(_ imageView: AnimatedImageView, didPlayAnimationLoops count: UInt) {}
+    public func animatedImageViewDidFinishAnimating(_ imageView: AnimatedImageView) {}
 }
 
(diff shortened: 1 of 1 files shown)


onevcat:
@JacobMao Fantastic!

I noticed that you also list "Adding delegate" as to-do, I think we could merge it when that also gets done.

Thanks!

JacobMao:
@onevcat It's my pleasure.

I am thinking that I should define a new protocol or just add a closure property as a callback

onevcat:
I guess I would prefer a new protocol and corresponding delegate, following the normal UIKit way, as well as avoiding the potential (and annoying) memory management things in a strong closure.

JacobMao:
I agree your opinion and I prefer to use delegate.

BTW, 快过年了，line不是很忙，哈哈

JacobMao:
@onevcat I added the delegate.

Please review it when you're free, thanks!

onevcat:
@JacobMao Thanks for the additional commits!

May I have two requests below?

1. Instead of naming it to `didFinishAnimate`, which is a more likely closure property name, I think we could just follow the Cocoa convention, to name this delegate method to something like `animatedImageViewDidFinishAnimating(_ imageView: AnimatedImageView)`. It would be more home-like for users of this framework, who might be familiar with Cocoa framework.

2. Besides of the "finish" delegate, could you add another delegate method to notify users that a loop has been done. It could have a signature like `animatedImageView(_ imageView: AnimatedImageView, didPlayAnimationLoop count: Int)`. It could be useful if the users want to do something at the time when a single loop has finished.

Thanks again for your effort. The work looks amazing!

JacobMao:
@onevcat 
Thanks for your reply!

1. I followed the name pattern of `SKSceneDelegate`-> `optional func didFinishUpdate(for scene: SKScene)`. 
That's right, I should follow Cocoa convention.
2. I was going to add some methods like you said, but I don't know whether I should do more than scope of this issue. I will add this method.

Thank you again! 🙇

onevcat:
1. Umm, it's interesting that `SKSceneDelegate` is using the shorter version like that. I guess it might be an attempt for changing of API convention. Additionally, in game and/or CG render related things, developers might be more familiar with an `onXXX` API (like quite a lot of APIs on [this page](https://docs.unity3d.com/ScriptReference/MonoBehaviour.html)). However, it is obviously a violation to Cocoa Naming Guideline on [Naming Delegate Method](https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/CodingGuidelines/Articles/NamingMethods.html#//apple_ref/doc/uid/20001282-1001803-BCIDAIJE). Anyway, if the convention changed in future, it will not be hard to adopt. But for now, let's just follow the Cocoa guideline.

2. And please just add the method containing `count` in the same p-r. That's no problem for it. :]

JacobMao:
Got it. Thanks

PR #849, opened 2018-01-25
Compatible with CocoaPods 1.4.0

onevcat:


onevcat:
It seems that static building support will force all dependencies to build as static. So I guess it is not safe yet to support it now.

PR #861, opened 2018-02-13
Fixes Gif Loop Counting and Frame Skipping

### LINKED ISSUE #860: Gif Loop Delegate Callback And Max Repeat Count Reached Too Early
### Check List

- [x] I want to reoprt a problem instead of asking a question. It'd better to use [kingfisher tag in Stack Overflow](http://stackoverflow.com/questions/tagged/kingfisher) to ask a question.

### Issue Description

`AnimatedImageViewDelegate` callbacks happen too soon when animating a gif that has more frames than the `framePreloadCount`. Also, on the first loop, a frame is skipped because of an off-by-one error.

#### What

When the buffer for an animated image (`imageView.framePreloadCount`) is smaller than the number of frames in the animated image, the delegate method `animatedImageView(_:didPlayAnimationLoops:)` is called too soon. Instead of being called at the end of the loop, it is called when the number of frames in the buffer have been displayed. That means if you have a gif with 10 frames, and a `framePreloadCount` of `5`, the delegate method will be called after 5 frames have been displayed. Similarly, if you’re using `imageView.repeatCount = .once`, the gif will stop playback after 5 frames instead of 10.

#### Reproduce

[The steps to reproduce this issue. What is the url you were trying to load, where did you put your code, etc.]

I’ll use this gif for demonstration, as it is easier to see what’s happening. It’s got 10 frames, and each frame displays a number, 0 thru 9. ([External link](https://media.giphy.com/media/sS8US86vlpHK8/giphy.gif)).

example-gif

To demonstrate this issue, we use the following code in a view controller, setting the `framePreloadCount` to `5`, when we know the gif has 10 frames.

```swift
override func viewDidLoad() {
    super.viewDidLoad()
    
    imageView.delegate = self
    imageView.autoPlayAnimatedImage = false
    imageView.framePreloadCount = 5

    let url = URL(string: "https://media.giphy.com/media/sS8US86vlpHK8/giphy.gif")!
    imageView.kf.setImage(with: url)
}
```

When we tap a play button, animation begins:

```swift
@IBAction func playButtonTapped(_ sender: UIButton) {
    playButton.isEnabled = false
    imageView.startAnimating()
}
```

And finally, when the delegate method is called after a loop, we stop animating, and re-enable the play button:

```swift
func animatedImageView(_ imageView: AnimatedImageView, didPlayAnimationLoops count: UInt) {
    imageView.stopAnimating()
    
    playButton.isEnabled = true
}
```

The following shows this sample app running. Notice the two issues:

1. On the first loop, the frame that displays the number “5” is skipped.
1. Only 5 frames are shown before the delegate callback occurs.

feb-13-2018 10-18-22

#### Other Comment

1. The sample project I created that demonstrates this issue is here: [KingfisherLooping.zip](https://github.com/onevcat/Kingfisher/files/1720449/KingfisherLooping.zip)
1. I have a fix incoming for this. I will open a pull request.

mliberatore:
Fixes #860

## What it does

Fixes two issues that occur when the image buffer holds less frames than exist in an animated image:

1. Incrementing the loop count would occur when the number of frames in the buffer were displayed, not when all frames in the animated image were displayed.
1. The first loop of the animated image would skip a frame.

## How to test

1. See the example project linked from #860. You can either open that project and change the Podfile to use `pod 'Kingfisher', :git => 'git@github.com:mliberatore/Kingfisher.git', :branch => 'fix-loop-counting'`, and run `pod install`, or you can download the already updated project here:
[KingfisherLooping.zip](https://github.com/onevcat/Kingfisher/files/1720525/KingfisherLooping.zip).
1. Run the updated project, tap the play button, and see that the gif stops playback when an entire loop has completed, and that no frames were skipped.
1. Tap the play button again and see that the entire gif plays again.
1. Adjust the `framePreloadCount` to other values to see that it still works when it is greater than or equal to the number of frames (`10`).

## Gif

This shows the updated project with the expected behavior:

feb-13-2018 10-43-20

### DIFF
--- a/Sources/AnimatedImageView.swift
+++ b/Sources/AnimatedImageView.swift
@@ -305,6 +305,7 @@ class Animator {
     fileprivate let maxTimeStep: TimeInterval = 1.0
     fileprivate var frameCount = 0
     fileprivate var currentFrameIndex = 0
+    fileprivate var currentFrameIndexInBuffer = 0
     fileprivate var currentPreloadIndex = 0
     fileprivate var timeSinceLastFrameChange: TimeInterval = 0.0
     fileprivate var needsPrescaling = true
@@ -315,7 +316,7 @@ class Animator {
     private var loopCount = 0
     
     var currentFrame: UIImage? {
-        return frame(at: currentFrameIndex)
+        return frame(at: currentFrameIndexInBuffer)
     }
 
     var isReachMaxRepeatCount: Bool {
@@ -426,21 +427,24 @@ class Animator {
      */
     func updateCurrentFrame(duration: CFTimeInterval) -> Bool {
         timeSinceLastFrameChange += min(maxTimeStep, duration)
-        guard let frameDuration = animatedFrames[safe: currentFrameIndex]?.duration, frameDuration <= timeSinceLastFrameChange else {
+        guard let frameDuration = animatedFrames[safe: currentFrameIndexInBuffer]?.duration, frameDuration <= timeSinceLastFrameChange else {
             return false
         }
         
         timeSinceLastFrameChange -= frameDuration
         
-        let lastFrameIndex = currentFrameIndex
-        currentFrameIndex += 1
-        currentFrameIndex = currentFrameIndex % animatedFrames.count
-
+        let lastFrameIndex = currentFrameIndexInBuffer
+        currentFrameIndexInBuffer += 1
+        currentFrameIndexInBuffer = currentFrameIndexInBuffer % animatedFrames.count
+        
         if animatedFrames.count < frameCount {
             preloadFrameAsynchronously(at: lastFrameIndex)
         }
-
-        if currentFrameIndex == 0 {
+        
+        currentFrameIndex += 1
+        
+        if currentFrameIndex == frameCount {
+            currentFrameIndex = 0
             currentRepeatCount += 1
 
             delegate?.animator(self, didPlayAnimationLoops: currentRepeatCount)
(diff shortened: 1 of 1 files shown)


Sources/AnimatedImageView.swift
```diff
@@ -305,6 +305,7 @@ class Animator {
     fileprivate let maxTimeStep: TimeInterval = 1.0
     fileprivate var frameCount = 0
     fileprivate var currentFrameIndex = 0
+    fileprivate var currentFrameIndexInBuffer = 0
```
mliberatore: `currentFrameIndex` now tracks the current frame of the gif being displayed. `currentFrameIndexInBuffer` serves the old purpose of `currentFrameIndex`, which is the index of the current frame of the gif as it exists in the `animatedFrames` buffer.

Sources/AnimatedImageView.swift
```diff
@@ -380,7 +381,7 @@ class Animator {
         let frameToProcess = min(frameCount, maxFrameCount)
         animatedFrames.reserveCapacity(frameToProcess)
         animatedFrames = (0..<frameToProcess).reduce([]) { $0 + pure(prepareFrame(at: $1))}
-        currentPreloadIndex = (frameToProcess + 1) % frameCount
+        currentPreloadIndex = (frameToProcess + 1) % frameCount - 1
```
mliberatore: This is the off-by-one error that caused one frame to be skipped on the first loop of a gif.

onevcat:
LGGGGGGTM

Thanks!

PR #869, opened 2018-03-01
Frankrausch master

onevcat:
@frankrausch Thanks for the work.

Instead of putting the demo project outside, I added it to the workspace, so it will be easier for development purpose without switching between projects.

I would merge it and release a new version when CI passes.


frankrausch:
Thank you—perfect!

PR #875, opened 2018-03-14
Add swift_version to spec

hcanzonetta:
* Specifies the swift version that will be used for the pod regardless of the swift version the app target uses.

Note: This feature was added on CocoaPods 1.4.0 (http://blog.cocoapods.org/CocoaPods-1.4.0/)


onevcat:
@hcanzonetta Maybe we could also remove the `s.pod_target_xcconfig` in the spec file?

hcanzonetta:
@onevcat good point. Done!

Issue #837, opened 2018-01-02
Fetching images in the background mode


krodak:
### Check List

- [x] I have read the [wiki page](https://github.com/onevcat/Kingfisher/wiki) and [cheat sheet](https://github.com/onevcat/Kingfisher/wiki/Cheat-Sheet), but there is no information I need.
- [x] I have searched in [existing issues](https://github.com/onevcat/Kingfisher/issues?utf8=✓&q=is%3Aissue), but did not find a same one.
- [x] I want to reoprt a problem instead of asking a question. It'd better to use [kingfisher tag in Stack Overflow](http://stackoverflow.com/questions/tagged/kingfisher) to ask a question.

#### What

Currently, there is no way to schedule fetch of several images in background mode and execute some action after this fetch is completed.

#### What I've tried

An obvious start would be `ImagePrefetcher`. Unfortunately, it can't be used in background mode as it uses `completionHandler` from `ImageDownloader`, thus making it impossible to use in background mode on iOS.

Next, I've tried to use `ImageDownloader` directly to create my own `BackgroundImagePrefetcher`, but there is no way to DI custom `delegateQueue` to the `URLSession` used by `ImageDownloader`.

This was described in [this issue](https://github.com/onevcat/Kingfisher/issues/634) and @onevcat created a branch with an ability to change it, but it's deleted now.

#### What would be the solution

Short term solution would be to just open `urlSession` property on `ImageDownloader`. This way developers could facilitate it to fetch images in the background and use current delegates to respond to those being finished.

Long term solution would be to refactor `ImagePrefetcher` to use `ImageDownloader` without completion handlers, but with the use of `URLSession` delegate which is the only way that Apple supports background mode at the moment.

Ankish:
@onevcat : any possible resolution to this. This stalls my main thread and it affects the tableview scrolling.

Ankish:
@krodak : can you paste the temporary-fix code here, will be really helpful for me to analyze this than doing it all over again.

krodak:
@ankish in the end I went with custom solution as there are few places that Kingfisher forces main queue (local cache access is one of them too...)

Ankish:
@krodak : 

1) If I set completion handler to nil for ImagePrefetcher - would KF still have main queue issues ?
2) If I use  KingfisherManager.shared.retrieveImage and set URL here - it still does things in main thread. Did you try that as well ?

Ankish:
@krodak : 
Few things I am trying to analyse:
[...]

PR #883, opened 2018-03-27
WIP: new Option:  .useFileDownloads 

mishagray:
Still a WIP - but would love to get feedback.

We had an issue were occasionally, our backend would like to send as Images that were HUGE and loading the whole image into RAM would basically cause our app to 'die' on older devices (without even a crash report..).

This PR adds a 'useFileDownloads" option, so that instead of using NSURLSession.DataTask it will use NSURLSession.DownloadTask.    It will by default also include 'cacheOriginalImage' behavior, by immediately moving the downloaded file from NSURLSession directly into the file cache of the targetImageCache, so it doesn't need to be 'written' again.  If there are multiple ImageTargets, it will 'copy' the file into each.  

.useFileDownloads will cause the file to 'skip' the CacheSerializer on the original download (but not on processed images).
also adds
.dataReadingOptions(NSData.ReadingOptions), will allow you to manually 'set' how data from files will be mapped to data.   Typically you may want to consider using  . dataReadingOptions(.mappedIfSafe) to assist in loading large images.

One issue I haven't thought about is if you are trying to 'clear' the file cache, but have an open 'fileMapped' image at the same time, we should be ok with having the 'delete' fail.  But I haven't seen any issues 'clearing' and image that seems to be currently 'baked' with fileMapped memory.   But also can't ever confirm when 'mappedIfSafe' actually works!

So far, this seems to work fine.    I had problems running Tests in the xcode project.  I'm probably doing it wrong.  So no additional test cases yet.   Still needs some documentation for the code modifications in a few place.

In general... this should be faster than loading the image into ram and THEN writing it to disk.

This may also allow the cache to run with a background NSURLSession?  I haven't tested that yet - I think there are still are few missing delegate methods needed to totally support background downloads.   But since background downloads ONLY support downloadTasks, this should be the first step.


Sources/ImageCache.swift
```diff
@@ -226,6 +226,50 @@ open class ImageCache {
...
+     Will MOVE the original file instead of coping the file if 'moveFile' is true.
+
+     Returns the destination URL of the copied or moved file.
+
+      */
+
+    open func addFile(_ originalFileURL: URL,
+                       moveFile: Bool,
+                       forKey key: String,
+                       processorIdentifier identifier: String = "") -> URL
```
mishagray: Should maybe make URL optional so it's only returned if the 'copy' or 'move'  was successful.

onevcat:
@mishagray Thanks for the detail explanation and it looks good at the first glance to me. I will check it later when I get a chance and let you know!

Thank you!

onevcat:
@mishagray Thanks for keeping this pull request for such a long time!

Terribly sorry for the huge delay. Recently I was busy with the new version of Kingfisher so I didn't find a chance to see this until now.

For your situation:

> our backend would like to send as Images that were HUGE and loading the whole image into RAM would basically cause our app to 'die' on older devices (

Now Kingfisher 5 supports to use a `DownsamplingImageProcessor` to handle large images better. I believe it can solve your issue in a better and cleaner way. Instead of introducing another workflow for loading images, it uses downsampling to orignal image data (before decode it to pixel format) to create thumbnails.

You can check it here in [the wiki](https://github.com/onevcat/Kingfisher/wiki/New-In-Kingfisher-5#downsampling-for-high-resolution-images).

Your changes on converting data tasks to download tasks is interesting. Yes, it enables we use a background URL session and open the possibility of resuming a download. But in my use cases, it is very rare to have an image with such a large size to benefit a lot from resuming downloading. Although we had some discussion on it in #837 , I am not still wondering whether it worths to do so.

Since it is not a trivial change, and it seems that the original issue can be solved better in Kingfisher 5, can I close it without merging?

Thanks again for your contributing and patience!

Issue #411, opened 2016-08-21
Image Prefetcher async-ness


morisk:
Hi, 
Thank you for the awesome framework I'm using!

```Swift
    public func start()
    {
        // Since we want to handle the resources cancellation in main thread only.
        dispatch_async_safely_to_main_queue { () -> () in
        ....
    }
```
*imageprefetcher.swift:146*

Why Image prefetcher doing his job on main thread? 

>Since we want to handle the resources cancellation in main thread only.

But how do I prevent hooking UI?
Probably I also don't care from what thread i will cancel the queues. 

Maybe more general question is - why everything is in main queue?

onevcat:
All image resources management will happen in main thread, since under hood Kingfisher is using something like mutable array to manage it, which is not thread safe in Cocoa. Think that you are adding a new task while some code is trying to cancel one, if we do not specify a serial thread, there will be race condition. So here we need to dispatch to main thread.

As a framework user of Kingfisher, you could always start/cancel your task in any thread. It will not block the UI since all time-consuming parts (like downloading and caching) are happened in other threads. So it is not "everything" in main queue, but just that race-condition things.

hifall:
Just curious -- can you dispatch to a serial thread other than the main thread?

onevcat:
Sure, you could. But I guess it will be a premature optimization. Or even worse. Since thread dispatching itself is not free after all.

morisk:
Hi onevcat,
Thank you for the thorough answers(just saw it). I think maybe we can make better, 
I patched Kingfisher version to behave on background thread. Since I don't need safety it was easy and I have much improved results.

 I'll make PR with proper thread handling and i will be glad if you could take a look.

hifall:
@onevcat, thanks for your kind reply.

But that doesn't answer my question. Since you call dispatch_async_safely_to_main_queue to dispatch to the main queue, it incurs costs as well, just like dispatching to a background thread (I assume).

Since ImagePrefetcher has an optional completionHandler, a time-consuming completion handler passed in by the caller can still block the main UI, right?

morisk:
@hifall you right and this is what I moved away from main thread.

onevcat:
@hifall Hi,

`dispatch_async_safely_to_main_queue` will not do any dispatch if invoked from main thread. So there is no overhead if the users start a prefetch from main thread (which will always happen for most of framework users).

[...]

PR #885, opened 2018-03-27
Add class to cache images without using main thread

joystate:
As described in this [issue](https://github.com/onevcat/Kingfisher/issues/411), prefetching images on the main thread can block UI.
This pull request implements @morisk [solution](https://github.com/onevcat/Kingfisher/issues/411#issuecomment-243200085) as separate class for caching images without going to the main thread.


onevcat:
Hi, @joystate 

Thanks for it!

I am so sorry that I recently didn't get a chance to look at the issues. The discussion on that issue looks reasonable and I wish I could interact it soon. But I wonder why you need to make these change in a separated class. Can it be an improvement to the original ImageFetcher?

JacobMao:
Hi, @joystate .
Thanks for your PR.

I also think we don't need a new class to do it. we can improve `ImageFetcher`.
If you wanna use a new object to do it, I think it's better that use a struct instead of a class.
Swift prefer to use `guard` to do the `early exit`

PR #898, opened 2018-04-13
Some minor issues fixed.

Sunnyyoung:


onevcat:
@Sunnyyoung Thank you for it!

Everything seems fine expect the change for `public init(transform: @escaping Transformer)`. It is a public API changing and will break current code. I would do some additional work leave & deprecate the old version to keep code compatible. After I have a chance to do that I will merge it.

Thank you for your supporting.

PR #902, opened 2018-04-17
`KingfisherDidCleanDiskCache` should be a constant

kukushi:
I noticed that the `KingfisherDidCleanDiskCache` changed from constant to variable in the [migration](https://github.com/kukushi/Kingfisher/commit/b35af72) to Swift 3.

IMO, this may be an unintended modification, because any changes to `KingfisherDidCleanDiskCache`'s value may lead to unexpected behaviors.

### DIFF
--- a/Sources/ImageCache.swift
+++ b/Sources/ImageCache.swift
@@ -30,7 +30,7 @@ import AppKit
 import UIKit
 #endif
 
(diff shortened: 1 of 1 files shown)


onevcat:
Yes, you are definitely correct! Thanks!

PR #914, opened 2018-05-09
Fix app to stop AnimatedImageView in background state

ParkGwangBeom:
If your app uses AnimatedImageView to load multiple gif images, you will use a lot of cpu.
The problem here is that the app wastes the cpu even if it is currently in the background, so I modified it.

### DIFF
--- a/Sources/AnimatedImageView.swift
+++ b/Sources/AnimatedImageView.swift
@@ -159,10 +159,30 @@ open class AnimatedImageView: UIImageView {
         }
     }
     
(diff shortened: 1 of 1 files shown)


onevcat:
@ParkGwangBeom Good direction as an optimization.

However, the animation playing is now could be controlled by users with invoking `startAnimating` and `stopAnimating`. Maybe we also need to consider it when start/stop? (Say, the animation should not be started automatically if the user chooses to stop it manually before switching to background.)

PR #922, opened 2018-05-20
Replace  StringProxy with Kingfisher<Base>

woshiccm:
Using Kingfisher<Base> as String proxy is a better way, like UIImageView, UIImage, etc.
```
extension String: KingfisherCompatible { }

extension Kingfisher where Base == String
```

@onevcat


onevcat:
It's magic! Thanks!

PR #957, opened 2018-07-12
Changed ImagePrefetcher to process images in a background queue instead of the main queue

gordontucker:
When we tried to use ImagePrefetcher in a collection view that loaded the next few dozen images it caused very noticeable stuttering while scrolling. By moving the logic that handles the image downloading to its own DispatchQueue, it removed the stuttering completely and runs much smoother.

I can't see any side effects from this change so I figured I'd put in a pull request in case anyone else wanted to make their app a bit smoother. If you are curious they amount of stuttering it fixed, I'm attaching videos of how it looks before and after the proposed change.

[Kingfisher-before-and-after.zip](https://github.com/onevcat/Kingfisher/files/2190402/Kingfisher-before-and-after.zip)

### DIFF
--- a/Sources/ImagePrefetcher.swift
+++ b/Sources/ImagePrefetcher.swift
@@ -53,6 +53,10 @@ public class ImagePrefetcher {
     /// The maximum concurrent downloads to use when prefetching images. Default is 5.
     public var maxConcurrentDownloads = 5
     
+    /// The dispatch queue to use for handling resource process so downloading does not occur on the main thread
+    /// This prevents stuttering when preloading images in a collection view or table view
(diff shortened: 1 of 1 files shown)


ejmartin504:
I believe our app is suffering from the same issue. We prefetch a LOT of images when we first load the app, and our tableView starts to stutter.

@onevcat I think this would be a great addition to the library 👍

onevcat:
@gordontucker @ejmartin504 

Sorry for the late reply. LGTM at the first glance. I'd take a detailed look at these changes and merge it ASAP.

Thanks!

ejmartin504:
Great, thank you @onevcat!

PR #963, opened 2018-07-25
Add `waitForCache` option for waiting caching before returning the retrieved image

### LINKED ISSUE #956: Add an option item for waiting caching before returning the retrieved image
### Check List

#### What
After downloading an image, we sometimes need to wait for caching to be completed so that we can copy the file or use the cache path for some other operations. 

It would be great if there was a way where we could pass an option item and have the completion blocks be called after the caching is completed.

#### Other Comment
We actually have a solution on our fork. I'd be happy to create a PR & discuss our solution with you.

yigitcanyurtsever:
After downloading an image, we sometimes need to wait for caching to be completed so that we can copy the file or use the cache path for some other operations. This `waitForCache` option allows us to achieve that behavior.

Fixes #956

### DIFF
--- a/Sources/KingfisherManager.swift
+++ b/Sources/KingfisherManager.swift
@@ -170,7 +170,13 @@ public class KingfisherManager {
                                       processorIdentifier:options.processor.identifier,
                                       cacheSerializer: options.cacheSerializer,
                                       toDisk: !options.cacheMemoryOnly,
-                                      completionHandler: nil)
+                                      completionHandler: {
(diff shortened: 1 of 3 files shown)


onevcat:
Thanks for contributing! Seems cool. I will look at it this weekend. :]

yigitcanyurtsever:
Hey @onevcat. 👋Did you get a chance to look at this? Let me know if I can help.

onevcat:
LGTM! Sorry for the latency!

PR #976, opened 2018-08-10
Process images on a separate queue on `KingfisherManager`

### LINKED ISSUE #975: Use a different queue for processing images on `KingfisherManager` class
### Check List

#### What

Right now, the images are processed on the current queue which is blocking the main queue. There should be a separate `processQueue` on `KingfisherManager` just like the ones in `ImageCache` and `ImageDownloader`.

#### Reproduce

- Cache an image without using any processors.
- Retrieve the image with a processor using KingfisherManager. 
or
- Download an image with a processor and use `cacheOriginalImage` option (so that it is processed on `KingfisherManager`.

yigitcanyurtsever:
I've used UUID instead of name as a suffix so that it won't break any existing projects using Kingfisher.

Fixes #975

### DIFF
--- a/Sources/KingfisherManager.swift
+++ b/Sources/KingfisherManager.swift
@@ -86,6 +86,8 @@ public class KingfisherManager {
     var currentDefaultOptions: KingfisherOptionsInfo {
         return [.downloader(downloader), .targetCache(cache)] + defaultOptions
     }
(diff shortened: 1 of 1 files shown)


onevcat:
Thanks for it!

PR #980, opened 2018-08-20
Add an activity indicator style

ParkGwangBeom:
Currently there is a default indicator style for each OS version. Since the indicator is not visible depending on the background color of the screen, I added the style item so that it can be decided by the user.


onevcat:
It's a breaking change and I guess we will not merge it recently.

To customize an indicator, you could access the indicator after setting the activity type:

```swift
cell.cellImageView.kf.indicatorType = .activity
let activity = cell.cellImageView.kf.indicator?.view as! UIActivityIndicatorView
activity.color = .red
```

PR #982, opened 2018-08-21
Fix RepeatCount equal compare with .once and .finite(1)

zhongwuzw:
Add equal compare with `.once` and `.finite(1)`.


onevcat:
Wow, good catch!

PR #983, opened 2018-08-21
Replace class-only protocol keyword with AnyObject

zhongwuzw:
`Swift3` has already changed the keyword of class-only protocol, let's change to `AnyObject` as well.
image


onevcat:
Thanks for this!

PR #987, opened 2018-08-24
Add superview check when remove placeholder view

zhongwuzw:
Only remove placeholder view from superview if superview equal to imageView, because maybe placeholder view's superview is not equal to imageView.

### DIFF
--- a/Sources/Placeholder.swift
+++ b/Sources/Placeholder.swift
@@ -77,6 +77,8 @@ extension Placeholder where Self: View {
 
     /// How the placeholder should be removed from a given image view.
     public func remove(from imageView: ImageView) {
-        self.removeFromSuperview()
+        if self.superview == imageView {
+            self.removeFromSuperview()
+        }
     }
 }


onevcat:
@zhongwuzw Thanks for this PR. I am not quite sure when the case you described would happen. In the same extension, we have added the activity as a subview to image view. And if you provide another implementation for customizing activity indicator, you may also want to provide a `remove`. So I cannot see an obvious case or necessity for it.

Could you give me an example? Thanks!

zhongwuzw:
@onevcat Emm, the reason I added the check is to make it more robust, if user use a customView `A` `extension A: Placeholder {}` as a placeholder view, then he reuse this placeholder view for other purpose, maybe added to another view, after that, if `UIImageView` reuse and load a image with another placeholder `B`, it would reset placeholder, which means remove `A` from superview. So may break something. Maybe it's the rare case? Feel free to close this `PR` if it not necessary. 😂

onevcat:
I guess it is a misuse of that view and it should be prevented from using side instead of our framework. So I cannot see an obvious reason to add this checking. Anyway, thank you for letting me know this concern! I'll watch on it to see whether we need this in future.

PR #989, opened 2018-08-25
Break retain View when retrieve image

zhongwuzw:
Currently, we would retain `Kingfisher<Self>` and `base` when we retrieve the image, because `KingfisherManager` retain `Kingfisher<Self>`, and `Kingfisher<Self>` retain `base`, it would extend the lifetime of the `base` like `UIImageView`. I think if `base` released, bypass these things like `setImage` is necessary.

### DIFF
--- a/Sources/ImageView+Kingfisher.swift
+++ b/Sources/ImageView+Kingfisher.swift
@@ -83,36 +83,43 @@ extension Kingfisher where Base: ImageView {
             options.append(.preloadAllAnimationData)
         }
         
+        let weakKingfisher = WeakKingfisher<Base>(self.base)
+        
         let task = KingfisherManager.shared.retrieveImage(
             with: resource,
             options: options,
-            progressBlock: {[weak self] receivedSize, totalSize in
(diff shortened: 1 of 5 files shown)


onevcat:
Thanks for the PR. The change looks good, but basically, it is intended to hold them longer, so the completion handler will be always called. It keeps the behavior consistent and easier to understand.

Sources/ImageView+Kingfisher.swift
```diff
@@ -83,26 +83,34 @@ extension Kingfisher where Base: ImageView {
...
                 }
-                if let progressBlock = progressBlock {
-                    progressBlock(receivedSize, totalSize)
+                let kf = Kingfisher<Base>(base)
+                guard resource.downloadURL == kf.webURL else {
+                    return
                 }
+                progressBlock?(receivedSize, totalSize)
             },
-            completionHandler: {[weak base] image, error, cacheType, imageURL in
```
zhongwuzw: `[weak base]` make no sense in here TBO.

PR #1009, opened 2018-09-15
Updated to Swift 4.2

alexanderkhitev:


onevcat:
Should be done in #1010

PR #1016, opened 2018-09-19
Swift 4.2 support for cocoapods

fassko:
Swift 4.2 support for podspec, project and .swift-version


onevcat:
Swift 4.2 should be fully supported in version 4.10.

PR #1020, opened 2018-10-03
Cleanup code

ParkGwangBeom:
1. trailing closure
2. Eliminate meaningless self-use
3. Eliminate meaningless associated values ​​parameter of enum

### DIFF
--- a/Sources/Image.swift
+++ b/Sources/Image.swift
@@ -1021,10 +1021,10 @@ extension RectCorner {
         
         var result: UIRectCorner = []
         
-        if self.contains(.topLeft) { result.insert(.topLeft) }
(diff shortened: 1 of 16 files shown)


onevcat:
@ParkGwangBeom Hi, great work! It looks much cleaner now.

However, I have a plan to make these clean-up works in 5.0. I am also already in progress of that major release. So it would be not easy to merge these changes. Could you instead check whether you could do them from the 5.0 branch and send the PR there if you'd like? 

https://github.com/onevcat/Kingfisher/tree/5.0

Thanks!

ParkGwangBeom:
@onevcat #1021 5.0 Based on PR

PR #1030, opened 2018-10-17
Add preload disk cache to memory with prefetching support

xspyhack:
This is useful when you want to warm up the image to improve retrieve time.


onevcat:
Thanks for it.

I am now rewriting the bad code of this project now, quite a few things have been changed. It would be easier to make the changes later directly in the refactored version, to avoid potential merge conflicting.

It is a useful feature and I would keep it in mind in the new version (or you could try to add it again in the new version later). Thank you!

onevcat:
Done in branch 5.0.

PR #1037, opened 2018-10-26
Add .onFailureImage(Image) option to set default image when loading request will be failed

dstranz:
**Use case**
Let's assume that we have an app with a list of users, with the following conditions:
* Each user could have an avatar, but we have also users without it.
* We don't wont to have empty gaps in UI for users without an avatar.
* We know how to create avatar URL for given user, but we are not sure if it will work properly or return 404 response.
* We would like to use the default image for users without an avatar.

**Could we use placeholder property?**
Nope. Because with placeholder user will see blinking placeholder image changing to the proper avatar (for the existing ones).

**Changelog**
☑️ Add `.onFailureImage(Image)` option to KingfisherOptionsInfo
☑️ Add support for this option in ImageView, UIButton, NSButton and WKInterfaceImage extensions
☑️ Add test cases to iOS & MacOS test suites

### DIFF
--- a/Sources/Extensions/ImageView+Kingfisher.swift
+++ b/Sources/Extensions/ImageView+Kingfisher.swift
@@ -134,6 +134,9 @@ extension KingfisherClass where Base: ImageView {
                         )
                         #endif
                     case .failure:
+                        if let image = options.onFailureImage {
+                            self.base.image = image
+                        }
                         completionHandler?(result)
                     }
                 }
--- a/Sources/Extensions/NSButton+Kingfisher.swift
+++ b/Sources/Extensions/NSButton+Kingfisher.swift
@@ -88,6 +88,9 @@ extension KingfisherClass where Base: NSButton {
                         self.base.image = value.image
                         completionHandler?(result)
                     case .failure:
+                        if let image = options.onFailureImage {
+                            self.base.image = image
+                        }
                         completionHandler?(result)
                     }
                 }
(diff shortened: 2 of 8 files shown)


onevcat:
Hi, @dstranz 

I understand your purpose, but I still wonder how this option is a must. Does this code work for your case?

```swift
imageView.kf.setImage(with: "https://domain.com/image.png") {  
    image, error, cacheType, imageURL in
    if let error = error {
        imageView.image = placeholderImage
    }
}
```

dstranz:
@onevcat Yes, it would work. But it will be more useful to set the option rather than providing a custom callback in each case.

onevcat:
Sorry for the late reply.

I got your idea. And yes, maybe it is good to have it as an option.

However, I am working on a major release in the branch `5.0`, which contains a few refactoring to this framework. I want to contain this as a feature from 5.0, could you try to rebase your work to that branch? So we could reduce the merge conflicting.

And I suggest instead of `.onFailureImage(Image)`, we'd prefer `.onFailureImage(Image?)`, which can give users a chance to set the image view empty.

I will later do some additional work based on your progress, like adding the transition animation to the failure image when `.transition` option provided.

Thank you!

dstranz:
@onevcat I've rebases my work on 5.0 branch and change PR target branch.

I've also add an option to use .failureImage(nil) and `public var onFailureImage: Optional<Image?>` to provide `.some(Image?)` when this option is available and `.none` when is not. Hope it's fine for you.

onevcat:
@dstranz I think that's the fantastic use case for double optional. Thanks!

PR #1052, opened 2018-11-26
Update Kingfisher.swift

honbolau:
Remove redundant code

### DIFF
--- a/Sources/Kingfisher.swift
+++ b/Sources/Kingfisher.swift
@@ -58,8 +58,6 @@ public final class Kingfisher<Base> {
  A type that has Kingfisher extensions.
  */
 public protocol KingfisherCompatible {
-    associatedtype CompatibleType
-    var kf: CompatibleType { get }
 }
 
 public extension KingfisherCompatible {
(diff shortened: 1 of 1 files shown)


onevcat:
@honbolau Thanks for the PR.

But I didn't see a good reason for this change. Since the `KingfisherCompatible` is a public protocol, this change at last breaks the compatibility (although it probably no one is trying to conform to it) somehow.

Maybe it brings minor performance improvement since it gets rid of the dynamic dispatch. But it should be almost ignorable. How do you think about it?

honbolau:
If ` kf ` was rewritten by other users, so what's the meaning of other users to rewrite this property? What can you do after rewriting? ` kf ` type for ` Kingfisher `, defined as follows. With a final class like this, you cannot create a SubClass, so you cannot add an stored property directly. If only add computed property and associated property and method, don't have to fully implement a ` kf ` properties.

honbolau:
Feel free to close this PR if it not necessary. 😂

onevcat:
@honbolau 😂

I'd like to merge it, but not for version 4. Could you make the same change and send a pull request to branch 5.0? I think we could contain this change in the new major update version.

Issue #1062, opened 2018-12-10
Minimum IOS version 10?


ergunkocak:
Hello,

Just wanted to ask why is the minimum IOS version requirement increased to 10?

Best

liubiaocong:
如果需要Kingfisher5可以手动导入，改下源代码，参考https://www.jianshu.com/p/cbd383504b02

onevcat:
There is now an "ios9" branch https://github.com/onevcat/Kingfisher/tree/ios9 in which iOS 9 support is provided. Changes from master would be merged to that branch eventually, but we cannot guarantee it in the future.

esbenr:
"The share for all iOS 9 and earlier is below 1.0%" is a weird reason. that depends on the consumer. Out iOS9 percentage is above 10%. The app is targeted towards children, and they often get their parents legacy devices.

Also "Just keep using version 4 if you still need to support iOS 9" makes it impossible to have Swift 5+ and iOS 9 support.

The iOS9-branch seems to keep getting out of behind. I think this needs a permanent solution or we will have to find another solution. I love Kingfisher, though.

/Esben

PR #1078, opened 2018-12-21
maintain iOS 9 compatibility

jiangpengyun:


onevcat:
Thanks for the effort. As mentioned in #1062, the main reason for starting from iOS 10 is I have a plan to adapt to `UIGraphicsImageRenderer` soon. I prefer not making the deploy target be different in the same major version, so iOS 9 was dropped from the beginning.

For this reason, I am afraid this PR cannot be merged into master. However, I did receive a few requests on it and this PR looks good to me. So I made a branch "ios9" and would you like to change the base to that branch? So as long as the `UIGraphicsImageRenderer` not being added, everyone has a chance to use that branch if they are still in iOS 9.

Thanks!

onevcat:
(And please also check the compiling error...)

onevcat:
These changes should be already in the `ios9` branch. I am closing it now.

PR #1100, opened 2019-01-29
Use fileURLWithPath instead of string for URL init

jeffreylyg:
If use string for URL init, the disk storage is invalid because FileMeta is nil and can not get image data by Data(contentsOf: fileURL)

### DIFF
--- a/Sources/Cache/ImageCache.swift
+++ b/Sources/Cache/ImageCache.swift
@@ -239,7 +239,7 @@ open class ImageCache {
         var diskConfig = DiskStorage.Config(
             name: name,
             sizeLimit: 0,
-            directory: path.flatMap { URL(string: $0) }
+            directory: path.flatMap { URL(fileURLWithPath: $0) }
         )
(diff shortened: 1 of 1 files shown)


jeffreylyg:
If no

onevcat:
Hi, @jeffreylyg 

Thanks for the PR. You are totally correct on the naming convention, the parameter `path` indeed should be a file path instead of an absolute string of a full file URL.

However, this method exists for years. And the change in this PR would be a breaking change to the current code. Instead of changing its behavior internally, I prefer to mark this method as deprecated and create a new one with correct naming and behavior inside.

How do you think about it? Would you like to send another PR for this?

onevcat:
Related work was done in #1114 

Thanks for opening this PR but I am closing it now.

jeffreylyg:
@onevcat I was planning to send another PR according your advice after the Spring Festival. I have missed a chance to contribute to your awesome project😂. Thanks for your fix anyway. :)

Issue #1101, opened 2019-01-29
Xcode 10.2 beta showing error for Image


rhals000327:
### Check List

Thanks for considering to open an issue. Before you submit your issue, please confirm these boxes are checked.

- [x] I have read the [wiki page](https://github.com/onevcat/Kingfisher/wiki) and [cheat sheet](https://github.com/onevcat/Kingfisher/wiki/Cheat-Sheet), but there is no information I need.
- [x] I have searched in [existing issues](https://github.com/onevcat/Kingfisher/issues?utf8=✓&q=is%3Aissue), but did not find a same one.
- [x] I want to report a problem instead of asking a question. It'd better to use [kingfisher tag in Stack Overflow](http://stackoverflow.com/questions/tagged/kingfisher) to ask a question.

### Issue Description
The new release of Xcode (which is 10.2 Beta) is producing errors 

#### What
https://github.com/onevcat/Kingfisher/blob/20d3f03762924aeb6ed133928f3875fa5b5c5d5e/Sources/Image/Image.swift#L233
Error showes up in above code. Saying Cannot mutate property 'kf' of immutable value 'image'.

#### Reproduce
Download Xcode 10.2 beta from apple developer site and build KingFisher. Done.

#### Other Comment
I'll make an PR for this

onevcat:
There is already a PR #1098 for Swift 5 support. We are waiting for CI support for Xcode 10.2 before we can merge it into master. If you need Swift 5 for now, please use the "swift5.0" branch instead.

PR #1102, opened 2019-01-29
Changed Image to mutable value

rhals000327:
Kingfisher isn't available to build in Xcode 10.2 Beta, and fixed the line that produces error for it.

For more information, please check #1101 this issue.


onevcat:
Thanks for it but it should be already done in #1098 

Once we can find a CI for it, we'd merge it into master. Before that, you can just use that branch in Swift 5.0 beta.

PR #1107, opened 2019-02-04
Drop deployment target to 9.3 from 10.0

zackshapiro:
Drop deployment target to 9.3


zackshapiro:
Been testing this on my phone, works well from my limited testing. Looks like it's fixed some of my crashes

zackshapiro:
I'm also testing on iOS 12 but with the deploy target of 9.3, so I'm essentially just testing that this builds and runs with sub-10 iOS deploy target

PR #1117, opened 2019-02-13
Fix for Xcode version 10.2 beta

leah:
This is just a tiny fix so Kingfisher will work with the latest beta Xcode. Xcode is unhappy that the variable is declared with a `let` instead of `var`, because it is changed later.

Hope this is an easy merge! 😄

### DIFF
--- a/Sources/Image/Image.swift
+++ b/Sources/Image/Image.swift
@@ -218,7 +218,7 @@ extension KingfisherWrapper where Base: Image {
         return image
         #else
         
-        let image: Image?
+        var image: Image?
         if options.preloadAll || options.onlyFirstFrame {
             // Use `images` image if you want to preload all animated data
             guard let animatedImage = GIFAnimatedImage(from: imageSource, for: info, options: options) else {


onevcat:
Hi,

Thanks for the PR. There is already a PR and branch for it. We will merge it when Xcode 10.2 officially released. Before that, you can use the "swift5.0" branch to use Kingfisher in Xcode beta version.

leah:
This is a very small change, unrelated to Swift 5.0. No problem though, it’s not urgent!

tprevost-phunware:
@onevcat Yeah, this also fails to build even in Swift 4.2 mode (for some reason), so I don't really think this is a Swift 5 issue (unless I'm misunderstanding something).

onevcat:
@tprevost-phunware 

No, it is not. It is due to the new compiler shipped in Xcode 10.2 has a more strict language model when handling mutating for a variable like this.

I am not sure what is your concern to just use the "swift5.0" branch (there also contains some modification for warnings in Xcode 10.2, and this PR should be just a subset of #1098 )? 

Does that branch work for Xcode 10.2 or not?

tprevost-phunware:
Ah, I see, thanks for clearing that up for me, @onevcat ! Nah I don't have a problem with using the `swift5.0` branch. I tried it just now, and it seems to compile just fine! Thanks again!

Issue #1121, opened 2019-02-25
Possible Deadlock on SessionDelegate?


zwang:
### Check List

- [x] I have read the [wiki page](https://github.com/onevcat/Kingfisher/wiki) and [cheat sheet](https://github.com/onevcat/Kingfisher/wiki/Cheat-Sheet), but there is no information I need.
- [x] I have searched in [existing issues](https://github.com/onevcat/Kingfisher/issues?utf8=✓&q=is%3Aissue), but did not find a same one.
- [x] I want to report a problem instead of asking a question. 

### Issue Description
The app stuck after fetching multiple images, debugging find there are two threads stopping at the same place fighting for a lock

#### What
Stack trace for main thread: 

```
com.apple.main-thread
0  libsystem_kernel.dylib         0x19bd7af54 __psynch_mutexwait + 8
1  libsystem_pthread.dylib        0x19bdfe0b4 _pthread_mutex_firstfit_lock_wait + 92
2  libsystem_pthread.dylib        0x19bdfe024 _pthread_mutex_firstfit_lock_slow$VARIANT$armv81 + 256
3  Kingfisher                     0x1069c10fc closure #1 in SessionDelegate.add(_:url:callback:) (SessionDelegate.swift:96)
4  Kingfisher                     0x1069a7694 partial apply for thunk for @escaping @callee_guaranteed (@guaranteed SessionDelegate, @unowned Int, @in_guaranteed SessionDataTask.TaskCallback) -> () (<compiler-generated>)
5  Kingfisher                     0x1069a7710 partial apply for specialized  (<compiler-generated>)
6  Kingfisher                     0x1069cb3d0 SessionDataTask.cancel(token:) (<compiler-generated>)
7  Kingfisher                     0x1069b9064 KingfisherWrapper<A>.cancelDownloadTask() (<compiler-generated>)
8  Plato                          0x10500bc10 GamePlayersCollectionViewCell.prepareForReuse() (<compiler-generated>)
9  Plato                          0x10500bc60 @objc GamePlayersCollectionViewCell.prepareForReuse() (<compiler-generated>)
10 UIKitCore                      0x1c8c9f83c -[UICollectionView _dequeueReusableViewOfKind:withIdentifier:forIndexPath:viewCategory:] + 588
11 UIKitCore                      0x1c8ca01b0 -[UICollectionView dequeueReusableCellWithReuseIdentifier:forIndexPath:] + 180
[...]

PR #1122, opened 2019-02-25
Makes lock behavior in session delegate more stable

onevcat:
This is a fix for #1121


zwang:
👍

onevcat:
Yes, will do it later.

PR #1123, opened 2019-02-27
Fix build with Xcode 10.2 beta

ChristopherRogers:


onevcat:
Hi, @ChristopherRogers 

Thanks for it. 

But there is already a branch "swift5.0" as well as a related PR https://github.com/onevcat/Kingfisher/pull/1098 for it. I guess it is ready to be merged and released soon (with some other fixes). Before that, you can try that branch instead to use Kingfisher in Xcode 10.2.

ChristopherRogers:
OK, I'll close this then. Thanks.

PR #1128, opened 2019-03-03
Feature/from cache than refresh

sxsasha:
Add options for get image first from cache and then refresh image (download from source). This is useful when you want to display a changeable image behind the same url at the same app session and use cache if no internet.

### DIFF
--- a/Sources/General/KingfisherManager.swift
+++ b/Sources/General/KingfisherManager.swift
@@ -173,7 +173,7 @@ public class KingfisherManager {
                 options: options,
                 completionHandler: completionHandler)
             
-            if loadedFromCache && !options.fromCacheThanRefresh {
+            if loadedFromCache && !options.fromCacheThenRefresh {
                 return nil
             }
             
--- a/Sources/General/KingfisherManager.swift
+++ b/Sources/General/KingfisherManager.swift
@@ -173,7 +173,7 @@ public class KingfisherManager {
                 options: options,
                 completionHandler: completionHandler)
             
-            if loadedFromCache {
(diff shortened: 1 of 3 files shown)


onevcat:
Thanks for the PR. But I don't understand what is the difference between this value and the `.forceRefresh` option?

sxsasha:
If  .forceRefresh set, Kingfisher will ignore the cache. But .fromCacheThenRefresh options get image from cache and then try to download

onevcat:
Oh, I see it.

I guess this addition makes the behavior not stable anymore. For these reasons:

1. The completion handler will be called twice, which is not expected actually. One invocation should correspond one and only one completion callback. The additional option logic here makes confusing on this convention and introduces a different pattern. It might be very hard to track and maintain later. There might be also some end users logic replies on the fact that the completion handler would be called once, it also breaks these kinds of size effect code.
2. As the image changed behind the same URL, end users may always want the latest image. I cannot see a good example of the use case for this option. A flickering of the image from a previous version to a newer version may also be thought as an implementation bug.
3. I guess the same behavior can be implemented easily by nesting two setting image methods, like this:
    ```swift
    imageView.kf.setImage(with: url) { result in
        imageView.kf.setImage(with: url, placeholder: result.value?.image, options: [.forceRefresh])
    }
    ```
This might be clearer to describe what is going on.

What do you think about it?

sxsasha:
Yes, you are right. Invocation twicely completion handler is will hard complicate logic and will provoke various bugs in the future. I use your example for implement cacheThenRefresh logic in my project. Thanks)

PR #1140, opened 2019-03-21
Add an option for disabling encryption for a cache file name

hasan-monument:
This option allows disabling md5 encryption for creation cache file name in DiskStorage.

### DIFF
--- a/Sources/Cache/DiskStorage.swift
+++ b/Sources/Cache/DiskStorage.swift
@@ -196,7 +196,7 @@ public enum DiskStorage {
         }
 
         func cacheFileName(forKey key: String) -> String {
-            if self.config.useEncryptionForCacheFileName {
+            if config.usesHashedFileName {
                 let hashedKey = key.kf.md5
                 if let ext = config.pathExtension {
                     return "\(hashedKey).\(ext)"
@@ -317,8 +317,8 @@ extension DiskStorage {
         /// Default is `nil`, means that the cache file does not contain a file extension.
         public var pathExtension: String? = nil
 
-        /// Default is `true`, means that md5 encryption will be used for creation the cache file name.
-        public var useEncryptionForCacheFileName: Bool = true
+        /// Default is `true`, means that the cache file name will be hashed before storing.
+        public var usesHashedFileName = true
 
         let name: String
         let fileManager: FileManager
--- a/Sources/Cache/DiskStorage.swift
+++ b/Sources/Cache/DiskStorage.swift
@@ -196,11 +196,18 @@ public enum DiskStorage {
         }
 
         func cacheFileName(forKey key: String) -> String {
-            let hashedKey = key.kf.md5
-            if let ext = config.pathExtension {
-                return "\(hashedKey).\(ext)"
+            if self.config.useEncryptionForCacheFileName {
+                let hashedKey = key.kf.md5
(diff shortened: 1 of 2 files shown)


onevcat:
@hasan-monument Thanks for it. 

The change looks good, but may I know what's the use case of it? The cache folder is visible to your users and storing the files with hashed names would protect privacy, that is especially important when you are trying save files with some personal information (like ID or names) in the URLs.

hasan-monument:
I was storing images like `id.jpg` by overriding this func `cachePath(forComputedKey key: String)` cause of some decision in our team. Now it is deprecated. If I use cache filename with md5 encryption, I need to re-download again with the new update. 

My suggestion is you should give an option to developers for that. Kingfisher also has the resource `ImageResource(downloadURL: URL, cacheKey: key)`, If developers want to use md5 encryption, they can pass the encrypted key with cacheKey or they can use different type encryption than md5.

This is just my suggestion, I respect your decision. Thanks for the review.

Sources/Cache/DiskStorage.swift
```diff
@@ -310,6 +317,9 @@ extension DiskStorage {
         /// Default is `nil`, means that the cache file does not contain a file extension.
         public var pathExtension: String? = nil
 
+        /// Default is `true`, means that md5 encryption will be used for creation the cache file name.
+        public var useEncryptionForCacheFileName: Bool = true
```
onevcat:
```suggestion
        public var usesHashFileName = true
```

Sources/Cache/DiskStorage.swift
```diff
@@ -310,6 +317,9 @@ extension DiskStorage {
         /// Default is `nil`, means that the cache file does not contain a file extension.
         public var pathExtension: String? = nil
 
+        /// Default is `true`, means that md5 encryption will be used for creation the cache file name.
```
onevcat:
```suggestion
        /// Default is `true`, means that the cache file name will be hashed before storing.
```

Sources/Cache/DiskStorage.swift
```diff
@@ -196,11 +196,18 @@ public enum DiskStorage {
         }
 
         func cacheFileName(forKey key: String) -> String {
-            let hashedKey = key.kf.md5
-            if let ext = config.pathExtension {
-                return "\(hashedKey).\(ext)"
+            if self.config.useEncryptionForCacheFileName {
```
onevcat:
```suggestion
            if self.config. usesHashFileName {
```

onevcat:
@hasan-monument Thanks for the explaination. I understood your case and I agree that we can provide an option for it.

May I ask you to change the name to `usesHashedName`, since MD5 is not an encryption but just a digest. It will be more accurate to say "hash".

And it would be great if you can also add some tests for this property too. After that, we can merge and release a new version for it. 

Thank you!

hasan-monument:
I didn't add a test case for validation md5 string, because I didn't want to import the library `CommonCrypto` to the file. If the test cases aren't enough, I can add it as well.

onevcat:
Let’s add it. May you can use kf.md5 from this https://github.com/onevcat/Kingfisher/blob/master/Sources/Utility/String%2BMD5.swift to avoid introducing the library importing.

hasan-monument:
Thanks for the hint. Added.

PR #1143, opened 2019-03-24
Fix stack overflow regression with ImagePrefetcher

AlbertTong:
Fix stack overflow when prefetching large number of images that were already previously cached to disk. A regression to version 3.3.0.

Also improve consistency for dispatching to queues using the newer CallbackQueue enum.


AlbertTong:
KingFisher Crash DiskStorage Screen Shot 2019-03-23 Regression

AlbertTong:
Screenshot of the stack overflow.
KingFisher Crash DiskStorage Screen Shot 2019-03-23 Regression

onevcat:
Thank you for it!

PR #1146, opened 2019-03-27
Use built in Result type on Swift 5

dstranz:
I think that it's good time to hide custom Result type declaration for Swift 5 clients and switch to built-in type. For Swift 4.2 and older it will be still visible.

### DIFF
--- a/Sources/Utility/Result.swift
+++ b/Sources/Utility/Result.swift
@@ -26,6 +26,7 @@
 
 import Foundation
 
+#if swift(<5)
 /// A value that represents either a success or failure, capturing associated
 /// values in both cases.
 public enum Result<Success, Failure> {
@@ -189,6 +190,7 @@ extension Result : CustomDebugStringConvertible {
         return output
     }
 }
+#endif
 
 // Deprecated
 extension Result {
--- a/Sources/Utility/Result.swift
+++ b/Sources/Utility/Result.swift
@@ -26,7 +26,9 @@
 
 import Foundation
 
-#if swift(<5)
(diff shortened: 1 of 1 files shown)


onevcat:
Thanks for it. It looks fine with a context of Xcode 10.2 and Swift 5 compiler. But it would not work well for Swift 4.2 compiler.

First, could you change the `swift(<5)` to something like `swift(>=5)`? The current PR would prevent Xcode 10.1 users to build the framework.

Furthermore, if I am correct, there are still several `Kingfisher.Result` methods (`value` getters or something like that) are not existing in `Swift.Result`. I marked them as deprecated in a version before, but by switching to `Swift.Result` now would probably break users' implementation and causes incompatible changes.

So I want to wait for a while now until the majority become to use Xcode 10.2. We can leave this PR opened until then.

dstranz:
@onevcat Take a look at this one
```
#if swift(>=4.3)
/// Result type already built-in
#else
```
it looks that "next" version number was already defined in Xcode 10.1.

MokaTheMartian:
Curious why this hasn't been pulled yet.

MoveUpwardsDev:
@onevcat I just face the issue spending 3 hours to find out why my Swift.Result is not correctly taken into account.

This PR has to be merged and new release shall be delivered to avoid the same to others.

MoveUpwardsDev:
Thanks 💯

MoveUpwardsDev:
@onevcat may i ask when you plan to release a new version to cocoapods that contains this MR ?

PR #1153, opened 2019-04-01
Update SwiftPM manifest file

Dschee:
This adds support for SwiftPM manifest based dependency managers. Specifically this adds support for installing via [Accio](https://github.com/JamitLabs/Accio) but will probably also work with SwiftPM once it's integrated into Xcode.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,32 +1,16 @@
-//
-//  Package.swift
-//  Kingfisher
-//
-//  Created by Wei Wang on 16/1/22.
-//
-//  Copyright (c) 2018 Wei Wang <onevcat@gmail.com>
-//
-//  Permission is hereby granted, free of charge, to any person obtaining a copy
-//  of this software and associated documentation files (the "Software"), to deal
-//  in the Software without restriction, including without limitation the rights
-//  to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
-//  copies of the Software, and to permit persons to whom the Software is
-//  furnished to do so, subject to the following conditions:
-//
-//  The above copyright notice and this permission notice shall be included in
-//  all copies or substantial portions of the Software.
-//
-//  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
-//  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
-//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
-//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
-//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
-//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
-//  THE SOFTWARE.
-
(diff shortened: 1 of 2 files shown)


Package.swift
```diff
@@ -1,32 +1,16 @@
...
-//  THE SOFTWARE.
-
+// swift-tools-version:4.2
 import PackageDescription
 
 let package = Package(
-  name: "Kingfisher",
-  exclude: ["Tests", "Sources/AnimatedImageView.swift", "Sources/UIButton+Kingfisher.swift", "Sources/WKInterfaceImage+Kingfisher.swift"]
+    name: "Kingfisher",
+    // platforms: [.iOS("10.0"), .macOS("10.12"), tvOS("10.0"), .watchOS("3.0")],
```
onevcat: Is it related to [this proposal](https://github.com/apple/swift-evolution/blob/master/proposals/0236-package-manager-platform-deployment-settings.md)? Is that mean if we change the `swift-tools-version` to 5.0, we can add this `platforms` to the package file?

Package.swift
```diff
@@ -1,32 +1,16 @@
...
+    products: [
+        .library(name: "Kingfisher", targets: ["Kingfisher"])
+    ],
+    targets: [
+        .target(
+            name: "Kingfisher",
+            path: "Sources"
+        )
```
onevcat: Should we also add the test target here? (Or is it possible to do so?)

Package.swift
onevcat: OK, I found explanation for it here https://github.com/JamitLabs/Accio#adding-support-for-accio.

Dschee: Yes, exactly, the reason why I tend to rather add this Swift 4.2 version of the manifest is because not everyone has updated their projects to Swift 5 yet, I think that takes around 3 months until most project are updated. But I'm personally totally fine with already updating the manifest format to Swift 5, I just haven't made different kinds of tests with it yet when using older projects, so keeping this at 4.2 is a safer bet (since Swift 5 can perfectly handle it).

Package.swift
Dschee: Adding the test target only makes sense in order to be able to run the `swift test` command which will currently not work due to the incompatibility of SwiftPM with UIKit AFAIK. Unless someone proves me wrong, it would only add a "maybe future support for SwiftPM" if we already added the test targets now.

Please also note that SwiftPM currently always checks out all dependencies listed under `dependencies` without looking at the targets they are needed for – therefore if you had dependencies that were only needed in the test targets, they would also be checked out in project where Kingfisher would be used without any purpose (since `swift test` doesn't work at the moment). Therefore I don't think the test targets should be added as of now.

Dschee:
@onevcat I answered your comments and hope I could clarify this exact design of the manifest file. If you have any more questions or disagree with one of the points, feel free to comment then I'll consider making the appropriate changes.

onevcat:
@Dschee That makes sense! Thank you for your explanation. I will try to use Accio myself later today to check everything goes fine. After that, we can merge it!

Thank you again for the changes.

Dschee:
Sounds great!

What I forgot to mention: I already tested basic building and integration support with iOS, tvOS & macOS in the branch `work/1000-frameworks` within the Demo project, as you can see [here](https://github.com/JamitLabs/Accio/blob/work/1000-frameworks/Demo/Shared/AppDependencies/Kingfisher.swift) and [there](https://github.com/JamitLabs/Accio/blob/work/1000-frameworks/Demo/Package.swift#L12). It worked as expected – feel free to add any other calls to the Kingfisher framework within the `Kingfisher.swift` file which you want Accio to officially test if there's anything else that makes sense. I'm only adding basic calls to ensure inclusion worked basically.

PR #1160, opened 2019-04-17
Fix cache not expire when setting maxCachePeriodInSecond to 0

zzzzeu:


onevcat:
As documented in the comment

> Setting this to a negative value will make the disk cache never expires.

So I guess the current logic is correct. How do you think about it?

zzzzeu:
> Setting this to a negative value will make the disk cache never expires.

That is exactly why I think it isn't quite right. 
When setting `maxCachePeriodInSecond` to 0, the expectation was the cache expire immediately, cause 0 isn't negative, don't you think?

onevcat:
Oh!

PR #1165, opened 2019-04-22
Crop image background to

DeadHipo:


DeadHipo:
I think that when specifying the background of the RoundedProcessor, you expect that the background will also be rounded to :)

onevcat:
@DeadHipo 

Hi, thanks for your contribution.

But the current state of the background color is intended. It is used for making the cropped corner part have a certain color, so when used in a container which has a solid color, it can match that background color.

PR #1183, opened 2019-05-14
Optimize MD5 String By `reduce` Function

wolfcolony:
I have a question，in "String+MD5.swift"，use “reduce” func may be better than use “map” and “joined” to create MD5 string by an [UInt8]，"map" may be cause space complexity O(n) ，it create a temp array [String]，Or are there other benefits to using "map" and "joined"?

### DIFF
--- a/Sources/Utility/String+MD5.swift
+++ b/Sources/Utility/String+MD5.swift
@@ -44,6 +44,6 @@ extension KingfisherWrapper where Base == String {
         }
         #endif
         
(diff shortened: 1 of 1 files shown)


onevcat:
@wolfcolony 

Thanks!

I am not sure which is faster, but I guess there should not be too much difference. Any chance to create some benchmark? So we can decide whether we should adopt to `reduce` instead of `map` + `join`.

PR #1184, opened 2019-05-16
Bring back macOS 10.11 into the game

vincedev:
This PR restores macOS 10.11 compatibility.

The deployment target has been set both in the macOS framework target and the demo application.
The code has also been [fixed](https://github.com/onevcat/Kingfisher/commit/86e3d9fda4dffcd17137ac5762c86ad791661786) for build to succeed.

### DIFF
--- a/Demo/Kingfisher-Demo.xcodeproj/project.pbxproj
+++ b/Demo/Kingfisher-Demo.xcodeproj/project.pbxproj
@@ -696,6 +696,7 @@
 				GCC_NO_COMMON_BLOCKS = YES;
 				INFOPLIST_FILE = "Demo/Kingfisher-macOS-Demo/Info.plist";
 				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/../Frameworks";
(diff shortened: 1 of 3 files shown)


onevcat:
Hi, thanks for it.

There is now already a branch "ios9" in which I guess the same thing was contained. I also downgrade the supported version for macOS there, can you check that branch?

For some reason described [here](https://github.com/onevcat/Kingfisher/pull/1078) and some other similar issues, we do not have a plan to introduce it back to master or a release.

vincedev:
Hi @onevcat 

Well alright I'll check the `ios9` branch.
Reading the issue linked, I can't see really why these changes would be an issue here, at least for macOS :
The timer scheduling code was the sole piece incompatible with macOS 10.11 APIs.. everything else builds fine. Also this change would not block any other development on iOS
Despite what Apple advises, macOS 10.11 remains an operating system in use because people tend to less update their systems than on mobile devices.

So.. well I think there's a little chance you consider supporting macOS 10.11 officially again, that's why I'm arguing here ;)

PR #1189, opened 2019-05-18
Improve animated image resizing (again?) and supports background decoding.

xspyhack:
This pull request contains **Image Resizing Improvement** (by using Image/IO instead of Core Graphics) and **Background Decode Supporting**, both for reducing the main thread CPU usage.

### Benchmark

Core Graphics (current)
Screen Shot 2019-05-18 at 2 37 11 PM

Image/IO
Screen Shot 2019-05-18 at 11 08 15 AM

Image/IO + BackgroundDecode
Screen Shot 2019-05-18 at 11 10 57 AM


Sources/Views/AnimatedImageView.swift
```diff
@@ -488,18 +496,21 @@ extension AnimatedImageView {
...
         private func loadFrame(at index: Int) -> UIImage? {
-            guard let image = CGImageSourceCreateImageAtIndex(imageSource, index, nil) else {
+            let options: [CFString: Any] = [
+                kCGImageSourceCreateThumbnailFromImageIfAbsent: true,
+                kCGImageSourceCreateThumbnailWithTransform: true,
+                kCGImageSourceShouldCacheImmediately: true,
+                kCGImageSourceThumbnailMaxPixelSize: max(size.width, size.height)
+            ]
+
+            guard let cgImage = CGImageSourceCreateImageAtIndex(imageSource,
```
onevcat: I guess this would trigger a trap if the size is `.zero`. Maybe we should follow the existing logic to ignoring the `needsPrescaling` when `.zero` is given.

Issue #1152, opened 2019-03-30
Error fetching files from disk.


hasankose:
### Check List
- [x] I have read the [wiki page](https://github.com/onevcat/Kingfisher/wiki) and [cheat sheet](https://github.com/onevcat/Kingfisher/wiki/Cheat-Sheet), but there is no information I need.
- [x] I have searched in [existing issues](https://github.com/onevcat/Kingfisher/issues?utf8=✓&q=is%3Aissue), I've found one similiar issue, but I couldn't be sure if they are same.

### Issue Description

I can not load files which have already been saved to disk with Kingfisher V < 5 when I update Kingfisher to V > 5 (I tested with the version 5.3.1).

#### What
The function above always returns true, because the `estimatedExpirationDate` is always less than the current date, so, it is always in the past after I've done the migration to Kingfisher 5. 
DiskStorage.swift, line 143
```swift
if meta.expired(referenceDate: referenceDate) {
    return nil
}
```

#### Reproduce

Save files to disk with Kingfisher V < 5, then update Kingfisher to V > 5. Open the app without the Internet. The files won't be shown up.

#### Other Comment
Probably I need to extend expiration dates in modificationDate for currently saved files if there is a migration to Kingfisher 5 from an earlier version. However, I don't think it is a good idea to change the attribute of a file. 

ie: When I set the expiration config to .never and download a file, the modification date of the file is set to 2062(something like that). It means the file never expires until the date, If I change my config to X days with the new app update, the attribute is not gonna be changed. So the file never expires anymore.

I was gonna create a PR, but the change affects few classes, I wanted to discuss first, what is the reason for writing the date to attributes instead of use expiration attribute in the framework dynamically? or Am I doing something wrong? What might be a reason for the issue?

Also, this issue might be related to the issue (1104).

onevcat:
@hasan-monument 

Thanks for your detail description and pointing out the behavior.

> Save files to disk with Kingfisher V < 5, then update Kingfisher to V > 5. Open the app without the Internet. The files won't be shown up.

As you may already know, both the disk cache and memory cache was fully rewritten in Kingfisher 5. The new version is now using a different strategy to determine cache expiration (by using the modify date attribute of the file), and there are also several image processor changes, which requires an invalidating of old caches.

[...]

PR #1196, opened 2019-05-29
Add option to disable memory cache TTL extending

dstranz:
# Use case:
Let's assume that we are having an image that is valid for a limited time. For example, USD/EUR rating chart that should be reloaded every 3 minutes. So we are using `.memoryCacheExpiration(.seconds(3 * 60))` option.

It's not worth to cache it on disk, because we will replace it every 3 minutes. So we are using `.cacheMemoryOnly` option.

## What will be the expected result?
After reloading the chart view we will get from Kingfisher cached image if not older than 3 minutes or a fresh one.

## What is actual behavior?
After reloading the chart view we will get from Kingfisher cached image if the time between last image load isn't longer than 3 minutes or a fresh one.

# Proposed change
In many cases actual behavior is correct. Why not extending TTL time for frequently used user avatar or media that will be the same every time we will load it?

But in other cases when the image should be not older than given TTL time we should add an option to disable extending functionality.

That's the reason why I've added a new option to Kingfisher called `.memoryCacheExpirationNotExtendable`.


dstranz:
I haven't got any problems with tests locally. They also pass on most of travis builds but not on all, so they are not reliable.

@onevcat Do you know what will a be better way to test cache TTL extending - probably without delay?

onevcat:
@dstranz 

Thanks for the proposal and implementation!

The CI sometimes does not work well due to some limitation on infrastructure. I will check it later, for both the use case and the tests.

dstranz:
Thank you for the suggestion @RomanPodymov.

It looks that it's not the only cause of failure. 

IMO the idea to test cache that way, with delays doesn't work well on CI because (probably) there are some slow downs and cache time elapsed before the last delayed block.

https://travis-ci.org/onevcat/Kingfisher/jobs/540614120 
✅  testCacheImageExtendingExpirationTask pass
🔴  testCacheImageExpirationTask failed

https://travis-ci.org/onevcat/Kingfisher/jobs/540614130
🔴  testCacheImageExtendingExpirationTask failed
✅  testCacheImageExpirationTask failed

I need to figure out better way to test cache with/without `memoryCacheExpirationNotExtendable` option.

Sources/General/KingfisherOptionsInfo.swift
```diff
@@ -203,6 +203,10 @@ public enum KingfisherOptionsInfoItem {
     /// value to overwrite the config setting for this caching item.
     case memoryCacheExpiration(StorageExpiration)
     
+    /// By default, every read operation for an item is extending it's expiration time. Use this setting to
+    /// disable this behavior and expire item regardless it's next requests.
+    case memoryCacheExpirationNotExtendable
```
onevcat: It's good to have this option. However, maybe we could make it more generic. Currently, we are extending the expiration duration according to the original `.memoryCacheExpiration` setting. Let's make it something like `.memoryCacheAccessExtendingExpiration(StorageExpiration)` so we can control it better. For example, in your case, you send a `. memoryCacheAccessExtendingExpiration(.seconds(0))` to prevent the extending, while others can choose `. memoryCacheAccessExtendingExpiration(.days(3))` to extend it by 3 days or `.memoryCacheAccessExtendingExpiration(.expired)` to make it expired immediately.

And maybe we could do the same thing to the disk cache, so we can solve #1152 as well.

Tests/KingfisherTests/ImageViewExtensionTests.swift
```diff
@@ -652,6 +652,84 @@ class ImageViewExtensionTests: XCTestCase {
...
+                group.leave()
+            }
+        }
+        
+        group.notify(queue: .main, execute: exp.fulfill)
+        
+        waitForExpectations(timeout: 3, handler: nil)
+    }
+    
+    func testCacheImageExpirationTask() {
```
onevcat:
```suggestion
    func testCacheImageNotExtendingExpirationTask() {
```

Tests/KingfisherTests/ImageViewExtensionTests.swift
```diff
@@ -652,6 +652,84 @@ class ImageViewExtensionTests: XCTestCase {
         group.notify(queue: .main) { exp.fulfill() }
         waitForExpectations(timeout: 3, handler: nil)
     }
+    
+    func testCacheImageExtendingExpirationTask() {
```
onevcat:
```suggestion
    func testImageCacheExtendingExpirationTask() {
```

onevcat:
CI machines are quite slow and it is not safe to judge a duration like 0.05s. I suggest make it 0.1, 0.3, 0.5s, which can improve the CI stability. Or instead of checking whether it is retrieved from memory cache or not, you can check the expiration duration after the access to confirm whether the time is extended, which might be also stable than a timer.

dstranz:
> Or instead of checking whether it is retrieved from memory cache or not, you can check the expiration duration after the access to confirm whether the time is extended, which might be also stable than a timer.

I've just rewritten both failing tests and then see your comment. Looks like we had similar idea. I'm only not sure if it's the best way to obtain expiration date:
```swift
let cacheKey = result.value!.source.cacheKey as NSString
let expirationTime1 = ImageCache.default.memoryStorage.storage.object(forKey: cacheKey)?.estimatedExpiration
```

>It's good to have this option. However, maybe we could make it more generic.
I will look at that when tests will pass for current solution 😉

Sources/General/KingfisherOptionsInfo.swift
dstranz: @onevcat Please take a look at the implementation of `.memoryCacheAccessExtendingExpiration(StorageExpiration)`.

In case of not extending I use `.memoryCacheAccessExtendingExpiration(.never)`.

onevcat: @dstranz Great!

> In case of not extending I use .memoryCacheAccessExtendingExpiration(.never).

I can understand your concerns here. However, I am not feeling good to have different meanings on `.never`: in the `.memoryCacheExpiration` it means "not expiring at all" but in `.memoryCacheAccessExtendingExpiration` now it is "not extending". This would definitely brings confusion and issues in future, even you have written great documentation on it.

Maybe we would need another enum besides of `StorageExpiration` (something like `ExpirationExtending`), or maybe we need to reconsider the name of `.memoryCacheAccessExtendingExpiration` if we want to keep using `StorageExpiration` as the associated value type. I personally prefer the former. Do you have any better idea?

dstranz: OK, we could use `.memoryCacheAccessExtendingExpiration(.seconds(0))` as "not extend" case. I understand your reason to not use `.never` but in such case without special treatment it will change TTL to 0 seconds and expire cache instead of not extending. Of course, we could add there `guard expiration.timeInterval != 0` to change default behavior.

I had something similar to `ExpirationExtending` yesterday, but later removed that. What do you think about such cases:
```swift
enum ExpirationExtending {
case .none
case .cacheTime
case .expirationTime(_ expiration: StorageExpiration)
}```

Default value will be `.cacheTime` as before change.

Comment (onevcat) [reply Sources/General/KingfisherOptionsInfo.swift]: Yes. I think your `ExpirationExtending` definition is so nice!

Comment (dstranz) [reply Sources/General/KingfisherOptionsInfo.swift]: @onevcat I've just pushed the code with `ExpirationExtending`.

File: Sources/Cache/MemoryStorage.swift [hunk 4]
```diff
@@ -121,26 +121,29 @@ public enum MemoryStorage {
...
+        
+        /// Use this when you actually access the memory cached item.
+        /// By default, this will extend the expired data for the accessed item.
+        ///
+        /// - Parameters:
+        ///   - key: Cache Key
+        ///   - extendingExpiration: expiration value to extend item expiration time:
+        ///     * .never will not extend expiration
+        ///     * nil (default value) will use object cache expiration settings
+        /// - Returns: cached object or nil
```
onevcat: Maybe also update the comment for `ExpirationExtending` to match current implementation?

Sources/Cache/Storage.swift
```diff
@@ -81,6 +81,20 @@ public enum StorageExpiration {
     }
 }
 
+/// Represents the expiration extending strategy used in storage to after access.
+///
+/// - none: The item expires after the original time, without extending after access.
+/// - cacheTime: The item expiration extends by the original cache time after each access.
+/// - expirationTime: The item expiration extends by the privided time after each access.
```
onevcat:
```suggestion
/// - expirationTime: The item expiration extends by the provided time after each access.
```

Sources/Cache/Storage.swift
```diff
@@ -81,6 +81,20 @@ public enum StorageExpiration {
...
+public enum ExpirationExtending {
+    /// The item expires after the original time, without extending after access.
+    case none
+    /// The item expiration extends by the original cache time after each access.
+    case cacheTime
+    /// The item expiration extends by the privided time after each access.
```
onevcat:
```suggestion
    /// The item expiration extends by the provided time after each access.
```

onevcat:
All green! Thanks a lot for that, @dstranz  It is awesome.

PR #1200, opened 2019-06-02
Add feature to cache failed image url

yliu342:
### Use Case
Add cache to image urls which have been retrieved before but does not return a valid response to prevent making calls to retrieve the url again.

### Expected Result
If the url has already been retrieved and user does not specify a `forceRefresh` then when downloading that URL again return a `.failedURL` error.

### DIFF
--- a/Sources/General/KingfisherError.swift
+++ b/Sources/General/KingfisherError.swift
@@ -63,6 +63,7 @@ public enum KingfisherError: Error {
     /// - URLSessionError: An error happens in the system URL session. Code 2003.
     /// - dataModifyingFailed: Data modifying fails on returning a valid data. Code 2004.
     /// - noURLResponse: The task is done but no URL response found. Code 2005.
+    /// - failedURL: The url has been requested before and does not have a valid URL response
     public enum ResponseErrorReason {
         
         /// The response is not a valid URL response. Code 2001.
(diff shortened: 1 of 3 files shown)


yliu342:
Hi, I'm fairly new to this library but I like how this library keeps async image loading simple and light weight. On recent use of the library I found out that whenever you tries to set an image to the image view but the URL does not exist then it will always tries to load the data. This seems to be a waste of user's data and wonder if we should keep a cache of the failed url event just in memory. That's the reason why this draft PR exist. I'm not sure if this is something you would want to consider?

onevcat:
Hi, thanks for this!

However, I am not sure whether it is a good idea to do that or not. Or at least, I guess it might be not a strategy which should be by default. It can lead to unexpected failing when your server is just down temporarily (for example a 500 error but be fixed later). The side effect of this behavior might impact the normal usage and I am personally against it.

Maybe in some certain cases it would boost and useful, but for most users, I guess this makes things shadowed and annoying.

yliu342:
Hi sorry for the late response. 
I totally agree with you it may lead to unexpected failing when user is not aware of it and it should not be on by default. But do you think it is necessary to add an option for it so people can choose to opt in into this feature?

yliu342:
This seems to be a rare case

PR #1208, opened 2019-06-13
修改scale

QDeveloper:
经常有gif在scale为3的设备上是模糊的，经过检查是这里的问题，希望可以改正


onevcat:
Thanks for the changes!

However, not like the image assets on device (which suffixed by `@2x` or `@3x`), the network images usually do not contain a scale factor. So Kingfisher will leave the decision to framework users. If you need to scale the image, you can use the `.scaleFactor` option when setting an image. 

Additional, changing the `base` property is not always a good idea and would break current implementation, so I am closing this PR.

PR #1209, opened 2019-06-14
Make Kingfisher Xcode 11 Swift Package Manager compatible

kylinroc:
Current version has some trouble with Xcode 11 built-in SwiftPM support, so I'm trying fix it.


onevcat:
LGTM. I changed the base to the branch for Xcode 11, and will eventually merge it later when other things prepared. Thank you!

PR #1201, opened 2019-06-04
Experimental: Support Project Catalyst

imWildCat:
According to the official documentation (https://developer.apple.com/documentation/uikit/creating_a_mac_version_of_your_ipad_app),

```swift
#if !targetEnvironment(UIKitForMac)
// Code to exclude from Mac.
#endif
```

These AND conditions are added to support the compilation for Project Catalyst

PR #1211, opened 2019-06-14
UIKit for Mac Support

samcat116:
This PR adds support for running UIKit apps on macOS. It also fixes an error in the podspec


onevcat:
Thanks for this!

The `#if` changes look good but the settings for swift versions are intended. You may need to upgrade your cocoapods version to 1.7 or above for them (check [this blog](http://blog.cocoapods.org/CocoaPods-1.7.0-beta/)) and we really want to keep the language compatible version as it is now.

onevcat:
An earlier pr for this topic #1201 was merged so I am closing this one. Thank you again @samcat116 .

Issue #1221, opened 2019-06-25
Platform specific files not properly handled in SPM


dstranz:
### Check List

Thanks for considering to open an issue. Before you submit your issue, please confirm these boxes are checked.

- [x] I have read the [wiki page](https://github.com/onevcat/Kingfisher/wiki) and [cheat sheet](https://github.com/onevcat/Kingfisher/wiki/Cheat-Sheet), but there is no information I need.
- [x] I have searched in [existing issues](https://github.com/onevcat/Kingfisher/issues?utf8=✓&q=is%3Aissue), but did not find a same one.
- [x] I want to report a problem instead of asking a question. It'd better to use [kingfisher tag in Stack Overflow](http://stackoverflow.com/questions/tagged/kingfisher) to ask a question.

#### What

iOS Project with Kingfisher as dependency installed through SPM in Xcode 11 failed to build due to AppKit imports

#### Reproduce

* Create/Open iOS Project in Xcode 11
* Add Kingfisher as dependency via built-in SPM
* Build

#### Other Comment

Build will failed due to this and similar errors:
```NSButton+Kingfisher.swift:28:8: "AppKit" is not available when building for iOS. Consider using `#if !os(iOS)` to conditionally import this framework.```

I don't know if it's possible to import different files per platform like in Pods, but maybe it will be enough to wrap MacOS files in `#if !os(iOS)` conditions?

PR #1222, opened 2019-06-25
Fix SPM errors in Xcode 11

dstranz:
As described in issue #1221 iOS app is unable to build because AppKit, UIKit and WatchKit wasn't conditionally imported in source files. I added `canImport` when it was necessary.

I need to add `platforms` to `Package.swift` to tell Xcode that it can use iOS 10 as development target for Kingfisher and fix a lot of "not available" errors.


onevcat:
Maybe it is duplicated to #1209 ?

The related changes are now under xcode11 branch.

dstranz:
@onevcat I missed that branches.

Personally I think it's better to use enum in platforms rather than strings:
`platforms: [.iOS("10.0"), .macOS("10.12"), .tvOS("10.0"), .watchOS("3.0")],`
vs
`platforms: [.iOS(.v10), .macOS(.v10_12), .tvOS(.v10), .watchOS(.v3)],`

and  `canImport` rather than `#if os(macOS)` but booth are will works fine.

onevcat:
@dstranz Oh, I agree with you! Can you instead send a PR to that branch? Big thanks!

dstranz:
@onevcat There are some other changes from master included in my branch - is it fine or should I create another branch from xcode11?

onevcat:
Yes, I suggest a new branch based on xcode11!

dstranz:
Reopened as #1223

PR #1249, opened 2019-08-06
subclassing fix for ImageDownloader

seanroehnelt:
The `open` variant of `downloadImage(...)` is never called by Kingfisher while the `internal` one is. This removes the legacy variant and allows for subclassing of the modern one.

Open to ideas to retain the `deprecated` `downloadImage(...)` method.

### DIFF
--- a/Sources/Networking/ImageDownloader.swift
+++ b/Sources/Networking/ImageDownloader.swift
@@ -193,8 +193,17 @@ open class ImageDownloader {
         }
     }
 
+    // MARK: Dowloading Task
+    /// Downloads an image with a URL and option. Invoked internally by Kingfisher. Subclasses must invoke super.
+    ///
+    /// - Parameters:
+    ///   - url: Target URL.
+    ///   - options: The options could control download behavior. See `KingfisherOptionsInfo`.
+    ///   - completionHandler: Called when the download progress finishes. This block will be called in the queue
+    ///                        defined in `.callbackQueue` in `options` parameter.
+    /// - Returns: A downloading task. You could call `cancel` on it to stop the download task.
     @discardableResult
-    func downloadImage(
+    open func downloadImage(
         with url: URL,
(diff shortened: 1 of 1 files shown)


onevcat:
Thanks for pointing it out. But I don't think we should remove the existing method [here](https://github.com/onevcat/Kingfisher/pull/1249/files#diff-8a19ae40cc437108f4ba47ff5e9ade27L326). This breaks the source compatibility heavily.

It's fine to mark the new one as "open", as well as remove the "deprecated" one, but please just keep the existing open one at least.

seanroehnelt:
Sure, no problem I can put it back. I removed it because it needs to be called manually and is never called by Kingfisher. I'm not sure how to document the methods to let consumers know which method they can subclass. I mean, which methods will/won't be called by Kingfisher internally.

onevcat:
@seanroehnelt I understand your concern and situation well. The part around was a result of a recent refactoring, and yes, maintaining this kind of Object-Orientation relationship is not that easy.

In fact, it should be very rare to subclass the downloader, so it is not carefully designed from the very beginning. I would check it again later to see what we can do to improve it and make it more stable. Before that, may I ask what's your main purpose of subclassing it? Maybe there is a better way for you and I can give some suggestion on it.

spesic:
@onevcat In our project we subclass it to build a configurable automatic retry-er, so that when the download fails in certain scenarios - we can make `n` additional attempts to download asset.

seanroehnelt:
@onevcat This is specifically for when our image server has not yet generated the exact image -sized to our request. It returns a 408 telling us to retry later instead of keeping the http connection open.

seanroehnelt:
@onevcat this CI failure doesn't look related to my change and not able to repro it locally.

onevcat:
Got it. That makes sense.

Thanks for the fix.

PR #1257, opened 2019-08-10
fix thread safety issues of AnimateImageView frames

luochaojing:


onevcat:
@luochaojing LGTM. Thanks!

PR #1270, opened 2019-08-30
Avoid memory shortage when memory is not released in time when a larg…

yutian105:
When a large number of network images are loaded in a short time, there is not enough memory. When the images are serialized, the memory is not released in time.


onevcat:
LGTM!

Issue #1286, opened 2019-09-19
Add option to disable Disk cache TTL extending


makesource:
#### What

Maybe related with https://github.com/onevcat/Kingfisher/pull/1196.
But I wonder why not support this option in DiskCache.
I use both MemoryCache and DiskCache to optimize image load and works well in most of cases.
But when the image corresponds to static URL changed, Kingfisher never refresh cached image until extended expired date (which  updated every cache hit).
If the image frequently accessed, this expire date extends without limitation.
So that's why I need to control cache extend option in DiskCache.

If there's not similiar discussion on this topic, could I write a new PR on this?

onevcat:
Yes, I agree that we can add the same feature for the disk cache. Can you please create a PR for that? Thanks!

makesource:
resolve in PR https://github.com/onevcat/Kingfisher/pull/1287.
Thanks!

PR #1287, opened 2019-09-22
Add options regarding disk cache TTL extending

makesource:
As described in https://github.com/onevcat/Kingfisher/issues/1286, 
This PR is for diskCache TTL extending strategy.

## Use case
If you use diskCache option, you may want to control disk Cache expiration date.
But It is hard to control the expiration because of the TTL extending.
(In worst case, the image will never expired if frequently accessed)
So in this PR, I want to get some more accurate control of cache by offering diskCache expirationExtending options. (like memorycache extending option in #1196)

Feel free to give any suggestions and thoughts :) 
Thanks.

### DIFF
--- a/Sources/Cache/DiskStorage.swift
+++ b/Sources/Cache/DiskStorage.swift
@@ -119,7 +119,7 @@ public enum DiskStorage {
             config.fileManager.createFile(atPath: fileURL.path, contents: data, attributes: attributes)
         }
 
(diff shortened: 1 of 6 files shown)


makesource:
Hard to set up DiskCache Tests because `fileAttributeDate` rounds up miliseconds 😭

onevcat:
No worry! I will check the test things.

onevcat:
Related #1152

PR #1292, opened 2019-09-26
Remove black border from KFImage

yonaskolb:
This removes the border around KFImage. Here's an example of the issues the border causes
Screen Shot 2019-09-26 at 4 35 39 pm


yonaskolb:
@onevcat was this trying to fix something about the frame? The frame line above is interesting. Does that serve a purpose?

onevcat:
It is a mistake and I was using it as debugging purpose. It should be already fixed in this branch and I will tag it soon. https://github.com/onevcat/Kingfisher/tree/fix/swift-ui-border

onevcat:
See #1293

PR #1310, opened 2019-10-11
Remove taskIdentifier callback lock and call from main

onevcat:
This should be a potential fix for #1309


HowardCsie:
When will this be released?

onevcat:
Released in 5.9.0.

PR #1364, opened 2019-12-08
Throw an error for creating file failure

### LINKED ISSUE #1361: Storing image to disk cache fails but return success when file name is not hashed
When storing image to diskStorage we end up calling :
```
func store(
            value: T,
            forKey key: String,
            expiration: StorageExpiration? = nil) throws
        {
```
which fails when trying to store with a key containing a '/' character, but still returns success.
the issue is due to the fact we are using 
`createFile(atPath path: String, contents data: Data?, attributes attr: [FileAttributeKey : Any]? = nil) -> Bool`
And when we take a look at foundation we can see this function is a compatibility function that returns true or false if the file created with success
```
    /* These methods are provided here for compatibility. The corresponding methods on NSData which return NSErrors should be regarded as the primary method of creating a file from an NSData or retrieving the contents of a file as an NSData.
     */
```
Unfortunately I cannot find which method creating a file returns a NSError as mentioned in the above commentary.
It would be best to call this function but in the case we cant checking the boolean resulting from the `createFile(atPath path: String, contents data: Data?, attributes attr: [FileAttributeKey : Any]? = nil) -> Bool` function before returning success would be a good step forward.

onevcat:
This fixes #1361 .

### DIFF
--- a/Sources/Cache/DiskStorage.swift
+++ b/Sources/Cache/DiskStorage.swift
@@ -116,7 +116,14 @@ public enum DiskStorage {
                 // The estimated expiration date.
                 .modificationDate: expiration.estimatedExpirationSinceNow.fileAttributeDate
             ]
-            config.fileManager.createFile(atPath: fileURL.path, contents: data, attributes: attributes)
+            let fileWrote = config.fileManager.createFile(
+                atPath: fileURL.path, contents: data, attributes: attributes
+            )
+            if !fileWrote {
+                throw KingfisherError.cacheError(
+                    reason: .cannotCreateCacheFileOnDisk(path: fileURL.path, key: key, data: data)
+                )
+            }
         }
 
         func value(forKey key: String, extendingExpiration: ExpirationExtending = .cacheTime) throws -> T? {
--- a/Sources/General/KingfisherError.swift
+++ b/Sources/General/KingfisherError.swift
@@ -100,6 +100,7 @@ public enum KingfisherError: Error {
     /// - imageNotExisting: The requested image does not exist in cache. Code 3006.
     /// - cannotConvertToData: Cannot convert an object to data for storing. Code 3007.
     /// - cannotSerializeImage: Cannot serialize an image to data for storing. Code 3008.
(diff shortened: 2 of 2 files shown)


Sources/Cache/DiskStorage.swift
```diff
@@ -116,7 +116,14 @@ public enum DiskStorage {
                 // The estimated expiration date.
                 .modificationDate: expiration.estimatedExpirationSinceNow.fileAttributeDate
             ]
-            config.fileManager.createFile(atPath: fileURL.path, contents: data, attributes: attributes)
+            let fileWrote = config.fileManager.createFile(
+                atPath: fileURL.path, contents: data, attributes: attributes
+            )
+            if !fileWrote {
```
denisPoifol: Guard would seem more fitting to me.
Otherwise it looks great to me

Sources/General/KingfisherError.swift
```diff
@@ -139,6 +140,13 @@ public enum KingfisherError: Error {
         /// - original: The original image data, if exists.
         /// - serializer: The `CacheSerializer` used for the image serializing.
         case cannotSerializeImage(image: KFCrossPlatformImage?, original: Data?, serializer: CacheSerializer)
+
+        /// - cannotCreateCacheFile: Cannot create target cache file at the given path and key. Code 3009.
```
denisPoifol: Depending on if you intend to reuse this error case elsewhere or not, I would have given an explanation that this is due to the file manager not returning a clear error message.

Sources/Cache/DiskStorage.swift
onevcat: As the last statement of this function, it is in fact not guarding a condition but check the result. So I prefer using `if` here.

Sources/General/KingfisherError.swift
onevcat: I agree with you!

onevcat:
:( Ummm, I found a way to make it as two steps (`Data.write` followed by an attribute setting), so we can contain a Foundation error when writing to the disk cache.

I would close this and have another PR later.

onevcat:
I believe #1365 should solve it in a better way.

PR #1368, opened 2019-12-10
Move podspec config to SwiftUI subspec

keith:
This seems to only be used by the SwiftUI sources so it doesn't need to
be at the top level.


onevcat:
LGTM!

keith:
Thanks!

Issue #1362, opened 2019-12-04
JPEG compression quality fixed at 1.0


marco-palumbo:
https://github.com/onevcat/Kingfisher/blob/15fc36194a3b31ffbfe3e02322175d2af23e9da5/Sources/Image/Image.swift#L243

It would be nice if the JPEG compression quality could be parametric.

I'd be happy to fix this issue, but I'm not sure where's the best place to add the parameter.

onevcat:
Thanks for it. I don't think it is a bad idea to add it.

But before that, may I know some context? Are you trying to use this API directly or use it from somewhere else?

marco-palumbo:
Sorry, you're right: I've been too concise.
In the meanwhile, I've taken a deeper look into the code and realized that the issue is broader than what I was initially thinking.

### Context
I'm using Kingfisher in a collection view to display unprocessed thumbnail images. For the full-size images, I don't use Kingfisher. I do cache on disk all the images: Kingfisher's cache for thumbnail and my cache for full images. All images are JPEG.

### Issue
Looking at the file size of the cached images, I've noticed that the size of thumbnails was often bigger than that of the full images.
Digging into the code, I found out that unprocessed images are never saved using their original _data_.

https://github.com/onevcat/Kingfisher/blob/15fc36194a3b31ffbfe3e02322175d2af23e9da5/Sources/General/KingfisherManager.swift#L335-L336

They are always re-compressed with a compression quality of 1.0. That is what made me think that the compression quality should be parametrized. But, even if my initial proposal is still a valid enhancement, the real issue here is that there should be a way to cache the image original data even if it hasn't been processed.

onevcat:
@marco-palumbo 

Umm. Yes, you are right, that could be a size problem. The current implementation follows a unified workflow, regardless of whether the image is processed or not. Maybe we can improve this by the combination of:

1. By default, if no processor is applied, we should prefer to cache the original data. This helps to prevent multiple compression of the image.
2. Providing an option to specify image quality when saving images as JPEG.

And before these can be applied, if it is urgent to you, I suggest you can create your own `CacheSerializer` and use it as an option to override current behavior. Could be something like this:

```swift
public struct OriginalDataCacheSerializer: CacheSerializer {
    public func data(with image: KFCrossPlatformImage, original: Data?) -> Data? {
        return original
    }
    
    public func image(with data: Data, options: KingfisherParsedOptionsInfo) -> KFCrossPlatformImage? {
[...]

PR #1373, opened 2019-12-17
Feature customizable default cache serializer

onevcat:
For #1362

### DIFF
--- a/Sources/Cache/CacheSerializer.swift
+++ b/Sources/Cache/CacheSerializer.swift
@@ -111,9 +111,16 @@ public struct DefaultCacheSerializer: CacheSerializer {
     /// If `original` is `nil`, the input `image` will be encoded as PNG data.
     public func data(with image: KFCrossPlatformImage, original: Data?) -> Data? {
         if preferCacheOriginalData {
-            return original ?? image.kf.data(format: original?.kf.imageFormat ?? .unknown)
+            return original ??
+                image.kf.data(
+                    format: original?.kf.imageFormat ?? .unknown,
+                    compressionQuality: compressionQuality
+                )
         } else {
-            return image.kf.data(format: original?.kf.imageFormat ?? .unknown)
+            return image.kf.data(
+                format: original?.kf.imageFormat ?? .unknown,
+                compressionQuality: compressionQuality
+            )
         }
     }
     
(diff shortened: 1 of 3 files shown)


Sources/Cache/CacheSerializer.swift
```diff
@@ -80,8 +81,22 @@ public struct DefaultCacheSerializer: CacheSerializer {
     
     /// The default general cache serializer used across Kingfisher's cache.
     public static let `default` = DefaultCacheSerializer()
-    private init() {}
-    
+
+    /// The compression quality when converting image to a lossy format data. Default is 1.0.
+    public var compressionQuality: CGFloat = 1.0
```
Adobels: Does it have to be of `CGFloat`? Can't it be a `Double` as it was?

Adobels: compressionQuality might be defined as an enumeration with a default value set to high.
`public var compressionQuality: JPEGCompressionLevel = .high`
```
enum JPEGCompressionLevel {
  case custom(Double)
  case max, high, med, low

  var rawValue: Double {
    switch self {
    case .max:
      return 1.0
    case .high:
      return 0.9
    case .med:
      return 0.5
    case .low:
      return 0.2
    case .custom(let customValue):
      return customValue
    }
  }
}
```

onevcat: `UIImage` receives a `CGFloat` when converting to JPEG data: https://developer.apple.com/documentation/uikit/uiimage/1624115-jpegdata

So I guess keeping them aligned would reduce the mental pressure of users.

What do you mean by "a `Double` as it was"? I don't think this value once being a `Double` anywhere... 😕

Adobels: I have made the mistaken assumption about the original type of the compressionQuality property because CoreGraphics library was imported.
Below, updated code proposition.
```
enum JPEGCompressionLevel {
  case custom(CGFloat)
  case max, high, med, low

  var rawValue: CGFloat {
    switch self {
    case .max:
      return 1.0
    case .high:
      return 0.9
    case .med:
      return 0.5
    case .low:
      return 0.2
    case .custom(let customValue):
      return customValue
    }
  }
}
```

Example of use:
var cacheSerializer = DefaultCacheSerializer()
cacheSerializer.compressionQuality = .max
cacheSerializer.compressionQuality = .custom(0.8)

PR #1386, opened 2020-01-16
Converts local resource to file provider

onevcat:


woolie:
This seems to be working well. Can we have a new release for this please?

onevcat:
Yup. Should be soon if I can confirm there is no other emergency task. Before we can create a tag, you can use the master branch or the commit hash temporarily.

PR #1390, opened 2020-02-06
Feature / CommonCrypto implementation of MD5

Adobels:
Update a proprietary `String` `md5` implementation, to a `CommonCrypto` `CC_MD5` implementation.

CC_MD5 implementation leads to a faster calculation times.

**Example:**

Md5 calculated ten thousand times for a string "lded dezd zedz ed ez" leads to follwing results:

Proprietary implementation: 1.952 sec
CC_MD5 implementation: 0.784 sec

```
    func testMD5Proprietary() {

        measure {
            (0...10_000).map {_ in
            _ = "lded dezd zedz ed ez".kf.md5
            }
        }

    }
    
    func testCC_MD5() {
        
        measure {
            (0...10_000).map {_ in
                _ = "lded dezd zedz ed ez".md5
            }
        }
    }
```


onevcat:
Thanks for the PR! It totally makes sense. But if you check the history of the String+MD5.swift file, you will find that [it was](https://github.com/onevcat/Kingfisher/commit/ce869e762bffba8a3d36cbcfc83d362eeed89bc3#diff-83528e464139a7490373fe498f77c544) using `CC_MD5` when calculating MD5 hash. However, Xcode gives us a warning on that:

> This function is cryptographically broken and should not be used in security contexts. Clients should migrate to SHA256 (or stronger).

It is the fact that MD5 is not for crypto purposes anymore, but I believe it is still serving well in Kingfisher for a file hashing purpose. However, there is no way (as I know) to ignore this warning.

The result, in turn, makes CocoaPods unhappy since a warning would prevent pushing to pod spec, unless we ignore all warnings. I don't like that.

So a customized implementation of MD5 is used here. Maybe we can migrate to another digest way like SHA256, but it breaks the current cache totally so we can only do it in the next major release.

By the way, it is not that fair to do performance test like that. The compiled CommonCrypto is highly optimized, but the customized MD5 is compiled in Debug mode without optimization (I guess you are just running these tests with `-Onone`?) It would slower than the pure C implementation, but since the algorithm is the same, I think the gap between them would be smaller. 

Any improvement which not triggering a warning is welcome, but I am sorry replacing it with the deprecated `CC_MD5` is not an option unless we can find a way to suppress the warning.

Adobels:
I haven't checked the history of the file before the PR. Ok, I see it now, you already had the CommonCrypto implementation before.

Concerning a performance results, you are right. When optimization is enabled the result is inverse. The proprietary implementation is faster than the CC_MD5.

Thanks for the feedback. I am closing the PR

PR #1394, opened 2020-02-23
Fixes race condition for State of CacheCallbackCoordinator

bguidolim:
I've enabled the thread sanitizer to check if everything was ok in my project, then I found an issue in this class.

To reproduce the issue, you need to make several calls to KingfisherManager, not sure if the options affect the reproducibility. Here is the code I am using:

```swift
let processor: RoundCornerImageProcessor = .init(cornerRadius: SpotlightManager.avatarTargetSize.width / 2.0,
                                                 targetSize: SpotlightManager.avatarTargetSize,
                                                 backgroundColor: .clear)

let options: [KingfisherOptionsInfoItem] = [.backgroundDecode,
                                            .downloadPriority(URLSessionTask.lowPriority),
                                            .callbackQueue(.dispatch(avatarQueue)),
                                            .processor(processor),
                                            .scaleFactor(UIScreen.main.scale),
                                            .cacheSerializer(FormatIndicatedCacheSerializer.png),
                                            .waitForCache,
                                            .diskCacheExpiration(.expired)]

KingfisherManager.shared.retrieveImage(with: imageURL, options: options) { result in
    switch result {
    case .success(let value):
        let pngData: Data? = value.image.pngData()
        searchableUser.attributeSet.thumbnailData = pngData
    case .failure:
        break
    }
}
```

### DIFF
--- a/Sources/General/KingfisherManager.swift
+++ b/Sources/General/KingfisherManager.swift
@@ -628,11 +628,22 @@ class CacheCallbackCoordinator {
     private let shouldWaitForCache: Bool
     private let shouldCacheOriginal: Bool
 
-    private (set) var state: State = .idle
+    private let stateQueue: DispatchQueue
+    private (set) var state: State {
+        set {
+            stateQueue.sync { threadSafeState = newValue }
+        }
+        get {
+            stateQueue.sync { threadSafeState }
+        }
+    }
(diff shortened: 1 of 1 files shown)


onevcat:
LGTM!

PR #1399, opened 2020-03-05
Change contributing documentation

ionutivan:
Trivial change of link for seeing existing issues and add new issue. Old one was redirecting to a 404 page.


onevcat:
Good catch! God knows how it can be that...

Thanks.

PR #1405, opened 2020-03-10
BUILD_LIBRARY_FOR_DISTRIBUTION = YES;

RomanPodymov:
Hello.
Thank you for Kingfisher.
In this pull request I added `BUILD_LIBRARY_FOR_DISTRIBUTION = YES;` to **project.pbxproj**. It will be useful for developers who are using different Swift versions.


onevcat:
If I am correct, this flag is basically for building xcframework. The prebuilt ones are already provided in the release page https://github.com/onevcat/Kingfisher/releases

RomanPodymov:
> If I am correct, this flag is basically for building xcframework. The prebuilt ones are already provided in the release page https://github.com/onevcat/Kingfisher/releases

Hello @onevcat 
For example, if you install Kingfisher via Carthage, you should run `carthage bootstrap` every time you switch Swift/Xcode version. I think that with `BUILD_LIBRARY_FOR_DISTRIBUTION = YES;` it is not necessary.

onevcat:
Thank you for the explanation.

However, according to [Apple's suggestion](https://swift.org/blog/library-evolution/):

> **Library evolution support should only be used when a framework is going to be built and updated separately from its clients.** In this scenario, a client built against an old version of the framework can be run with a new version of the framework without being recompiled.

As an open-source project, Kingfisher obviously doe not fall into this. You can (and should) always compile this framework as your product binary. I guess it is more likely an abusing of the `BUILD_LIBRARY_FOR_DISTRIBUTION` flag to bypass the carthage build when not using the same compiler.

I am closing it since the current behavior is intended.

PR #1421, opened 2020-03-29
Fix DownsamplingImageProcessor example

mattlogan:
`UIView` doesn't have a `size` property.

You can get a view's size with `UIView.bounds.size` -- I assume that's what was intended here.


onevcat:
Thanks for it!

PR #1430, opened 2020-04-13
Binary stability

Wassmd:
We need to make SDK ABI compatible. Please check https://swift.org/blog/library-evolution/

Just one change needed i.e BUILD_LIBRARY_FOR_DISTRIBUTION = YES in build settings. This will make KF SDK to run on higher version of Swift and Xcode without being compilation required in the higher version.


onevcat:
Duplicated #1405 

This is not a flag that should be on by default. If you just want to embed Kingfisher as binary, you can choose to download the prebuilt xcframework versions in the release page. Otherwise, the framework is open-sourced and should be built every time you need to ship your final binary.

onevcat:
But if you really want to build Kingfisher in the `BUILD_LIBRARY_FOR_DISTRIBUTION` way, I suggest you can choose CocoaPods and pass the flag there in the build config. However, it does not make much sense to enable it as a default value for an open-source project.

PR #1435, opened 2020-04-18
fix issue with image modifier not working on provider provided images

lkzhao:
Issue: 
When using a ImageProvider, imageModifier is not applied to the provider provided image. ImageModifier does apply to subsequent access from the cache.

Not sure if this is the best way fixing this issue. Let me know.


onevcat:
Nice catch!

PR #1451, opened 2020-05-15
Enable Define Module in Build Settings

daniele-pizziconi:
Rationale:
Enabling "Define Module" will trigger an automatic generation of the module.modulemap (https://clang.llvm.org/docs/Modules.html#id16), essential file to be able to use Kingfisher in libraries (that expose Kingfisher interfaces), which in turn will be used by Obj-c libraries/apps. 
This is affecting both SPM and Carthage usage, not CocoaPods since the latter will enable 'Define Module' in case of hybrid (objc-swift) consumer apps.


onevcat:
It seems that this PR fails the build now. I am not sure about the impact on current projects if we apply this. But I'd like to keep an eye to see the possibility.

daniele-pizziconi:
@onevcat, I've fixed the broken build. I had to enable CLANG_ENABLE_MODULES, a setting complementary with the DEFINE_MODULES.

PR #1480, opened 2020-06-30
Reduce synchronous I/O when checking `ImageCache.isCached`

raja-baz:
We use `ImageCache.isCached` on the main thread in our application and we noticed that it would sporadically cause the main thread to hang on I/O. Some investigation showed that this was due to 2 things.

1. A straight-up bug when building the file URL in `DiskStorage` which uses a variant of `URLByAppendingPathComponent:` that does disk I/O to figure out if the path it's building is for a directory or a file. In our case we know it's a file and this is completely unnecessary and we can simply use `URLByAppendingPathComponent:isDirectory:` and avoid this. 
2. While KingFisher's ultimate source-of-truth on whether an image is cached or not is whether the file exists on disk, it is possible to cheaply put an in-memory hash table check before doing so that filters out most unnecessary trips to the file system without changing the underlying semantics. Note that this should be carefully reviewed before merging as I am unsure if that structure needs to be synchronized(it currently isn't).

### DIFF
--- a/Sources/Cache/DiskStorage.swift
+++ b/Sources/Cache/DiskStorage.swift
@@ -46,6 +46,8 @@ public enum DiskStorage {
         public let directoryURL: URL
 
         let metaChangingQueue: DispatchQueue
+        
+        var maybeCached : Set<String>?
 
         /// Creates a disk storage with the given `DiskStorage.Config`.
         ///
@@ -72,6 +74,17 @@ public enum DiskStorage {
             metaChangingQueue = DispatchQueue(label: cacheName)
 
             try prepareDirectory()
+            
+            do {
+                maybeCached = Set()
+                try config.fileManager.contentsOfDirectory(atPath: directoryURL.path).forEach { (fileName) in
+                    maybeCached?.insert(fileName)
+                }
+            } catch {
+                // Just disable the functionality if we fail to initialize it properly. This will just revert to
+                // the previous behaviour which is to check file existence on disk directly
+                maybeCached = nil
+            }
         }
 
         // Creates the storage folder.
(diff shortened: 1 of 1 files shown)


onevcat:
@raja-baz Thanks for it.

> A straight-up bug when building the file URL in DiskStorage which uses a variant of URLByAppendingPathComponent: that does disk I/O

Oh, I never thought that it would be an expensive operation. I will check it.

> While KingFisher's ultimate source-of-truth on whether an image is cached or not is whether the file exists on disk, it is possible to cheaply put an in-memory hash table check before doing so that filters out most unnecessary trips to the file system without changing the underlying semantics.

It is a good idea to create a cache table in memory to prevent checking the file system. But the implementation in your PR would bring quite a few false positive, which is unacceptable. Think the case that disk cache is purged by the system (it is in the cache folder so that is totally possible) after the cache table being created. The cache system would think that the cache is on disk, but finally, nothing is found.

Anyway, I will consider whether there is a better way to improve the performance here. Thank you for your valuable suggestions!

raja-baz:
My design goal with the cache was to do a best-effort approach that changes as little and as low-risk as possible while being strictly better in terms of performance.

A false positive means we check the file system unnecessarily before finding out that the image isn't cached. However, right now that's the case for all operations. So it is strictly better than before with no visible downside. Unless I'm missing something?

Basically: I'm okay with false positive (avoidable I/O that we didn't avoid) but absolutely must never have a false negative (bug which causes network fetch when we could've loaded from disk). And I wanted to achieve that with as little change and intrusiveness in the code

raja-baz:
PS: I even had "maybe" in the name to indicate that this could have false positives

onevcat:
Ah! I got your idea. It seems that I misunderstood your implementation. It makes sense to add these as a "precondition" check for the actual disk file checking.

I will take a detailed look at these again.

PR #1485, opened 2020-07-11
allow custom hashing of file names

brindy:
Hello!  Firstly, thanks for creating and keeping this library maintained.  We've been using it over at DuckDuckGo since we built version 7 of our Privacy Browser.  https://github.com/duckduckgo/iOS

Long story short, this PR adds the ability to set a custom hashing function, but keeps all current defaults in place. I decided to add it as an extra property to the config rather than replace `config.usesHashedFileName` so that it stays backwards compatible.

The longer version is that we have been improving our favicon support and adding a simple (slightly native) implementation.  We decided to key against domain (as we try 3 different locations for images) and using sha256 with a salt for the file name, but I struggled to see how to do that with the latest version.

Please ket me know what you think! 

Thanks!

### DIFF
--- a/Sources/Cache/DiskStorage.swift
+++ b/Sources/Cache/DiskStorage.swift
@@ -259,8 +259,8 @@ public enum DiskStorage {
         }
 
         func cacheFileName(forKey key: String) -> String {
-            if config.usesHashedFileName {
-                let hashedKey = key.kf.md5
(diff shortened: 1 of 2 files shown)


brindy:
I realise now that using ImageResource(url, key) is probably the right way to do this.  Sorry for the noise!

PR #1240, opened 2019-07-22
Add extension to handle NSTextAttachment as KingfisherCompatible

PalringoBen:
This will allow using Kingfisher on NSTextAttachment in an attributed label. I've followed the same format for other extensions that add compliance for KingfisherCompatible.

We can now add a image to a label using Kingfishers cache like this
```
// Create a label
let label = UILabel()

// Create a attributed string
let attributedText = NSMutableAttributedString(string: "some string")

// Create the text attachment
let textAttachment = NSTextAttachment()

// Set the size of the attachment
let capHeight = label?.font.capHeight ?? 0
textAttachment.bounds = CGRect(x: 0, y: -1, width: capHeight + 2, height: capHeight + 2)

// Request the image from Kingfisher
textAttachment.kf.setImage(with: url)

// Add the text attachment to the beging of the text attributedText
attributedText.replaceCharacters(in: NSRange(), with: NSAttributedString(attachment: textAttachment))

// set the attributedText on the label
label.attributedText = attributedText
```

PR #1495, opened 2020-08-07
Add `NSTextAttachment` support

onevcat:
A better version of #1240


fukemy:
hello, i updated kingfisher library but can not find kf extension to NsTextAttachment

PR #1501, opened 2020-08-24
Allow asynchronous RequestModifiers

hoaCed:
Hi 👋 
This allows the RequestModifier to run asynchronously. This might be useful for refreshing auth headers before sending a request for example.
However, as a consequence, the DownloadTask chain has to be modified to use completions.


Sources/General/KingfisherManager.swift
```diff
@@ -441,11 +460,11 @@ public class KingfisherManager {
         }
     }
 
-    @discardableResult
     func loadAndCacheImage(
         source: Source,
         context: RetrievingContext,
-        completionHandler: ((Result<RetrieveImageResult, KingfisherError>) -> Void)?) -> DownloadTask.WrappedTask?
+        taskHandler: ((DownloadTask.WrappedTask?) -> Void)? = nil,
```
denisPoifol, line 466: Same thing here

Sources/Networking/ImageDownloader.swift
```diff
@@ -202,35 +202,64 @@ open class ImageDownloader {
     ///   - completionHandler: Called when the download progress finishes. This block will be called in the queue
     ///                        defined in `.callbackQueue` in `options` parameter.
     /// - Returns: A downloading task. You could call `cancel` on it to stop the download task.
-    @discardableResult
     open func downloadImage(
         with url: URL,
         options: KingfisherParsedOptionsInfo,
-        completionHandler: ((Result<ImageLoadingResult, KingfisherError>) -> Void)? = nil) -> DownloadTask?
+        taskHandler: ((DownloadTask?) -> Void)? = nil,
```
denisPoifol, line 208: Same thing here

Sources/Networking/ImageDownloader.swift
```diff
@@ -202,35 +202,64 @@ open class ImageDownloader {
...
+                )
             }
-            return nil
         }
+    }
+
+    private func downloadImage(with url: URL,
+                               request: URLRequest,
+                               options: KingfisherParsedOptionsInfo,
+                               taskHandler: ((DownloadTask?) -> Void)? = nil,
```
denisPoifol, line 261: Same thing here

Sources/Networking/ImageDownloader.swift
```diff
@@ -330,20 +359,21 @@ open class ImageDownloader {
     ///   - completionHandler: Called when the download progress finishes. This block will be called in the queue
     ///                        defined in `.callbackQueue` in `options` parameter.
     /// - Returns: A downloading task. You could call `cancel` on it to stop the download task.
-    @discardableResult
     open func downloadImage(
         with url: URL,
         options: KingfisherOptionsInfo? = nil,
         progressBlock: DownloadProgressBlock? = nil,
-        completionHandler: ((Result<ImageLoadingResult, KingfisherError>) -> Void)? = nil) -> DownloadTask?
+        taskHandler: ((DownloadTask?) -> Void)? = nil,
```
denisPoifol, line 366: Same thing here

denisPoifol (changes requested):
This seems ok on a functional level but I would tweak one ore two things :
- taskHandler closures should never be nil
- this change is leading to a new major version so it might be the opportunity to remove deprecated methods.

onevcat:
Hey, @hoaCed 

Thanks a lot for this PR. It looks nice itself and I am sure it would do its work perfectly!

However, it changed the public interface in a non-trivial way and obviously requires a major update version. I am not quite sure if this worths enough to apply these changes now since there is an easy enough way to separate your async operation and get what you want before sending them to Kingfisher's request modifier. It would keep the implementation of Kingfisher to keep simple and make the logic (especially how to handle the async error) clearly.

So, in my opinion, it is a low priority. I will leave this PR open for a while before we can find some other things requiring a major update. I think it might be a better chance to apply these too.

Thank you for your hard work!

onevcat:
Thanks again for this PR and all the contribution! I've implemented an async request modifier with a bit of trade-off. Now it does not change the current released APIs, but with a callback block to get back the `DownloadTask`. Good news is the download task is not being used for most of times so I think it worths to do this tradeoff.

I will merge #1589 instead and release a new version containing this.

Issue #1508, opened 2020-09-14
app crashs on iOS 14 when use Kingfisher in List


hassfers:
Hello

yes I know iOS 14 is still in beta but maybe it helps you ;-) 

I use kingfisher in a SwiftUI List to load pictures from urls. 
The initial state is shown as expected, but when I start scrolling the app crashs with a `Bad Access`- Error. 

If I delete the KFImage everything working well. 

If I use the same url for every cell it also works fine. 

Maybe there is some problem when SwiftUI tries to reuse the Cell and can't load the image somehow. 

Does anybody has an idea what to do ?

onevcat:
Hi,

Thanks for letting me know this.

But I tried the [SwiftUI List demo](https://github.com/onevcat/Kingfisher/blob/master/Demo/Demo/Kingfisher-SwiftUI-Demo/Views/SwiftUIList.swift) in the project and it seems working well and no crash happens.

Is it only on real devices or also on simulators? Do you have a sample or some code snippet that we can use to reproduce it?

Thank you!

hassfers:
Hello and thanks for the fast reply. 

After some more investigations I found out, that it only occurs on device, and only if the `KFImage` is in a `NavigationLink`. So it could also a bug with SwiftUI itself. Or Im doing something wrong. 

I updated the `SwiftUI List demo` and got the same `Bad Access` Error. 

```swift
import class Kingfisher.ImageCache
import KingfisherSwiftUI
import SwiftUI

struct SwiftUIList : View {

    let index = 1 ..< 700

    var body: some View {
        NavigationView{
            List(index) { i in
                ListCell(index: i)
            }.navigationBarTitle(Text("SwiftUI List"), displayMode: .inline)
        }
    }


    struct ListCell: View {
        let index: Int
        var body: some View {
            NavigationLink(destination: Text("\(index)")) {
                ListCellImage(index: index)
                    .scaledToFit()
                    .frame(maxWidth: 100, maxHeight: 100)
            }
        }
    }


    struct ListCellImage: View {

        @State var done = false

        var alreadyCached: Bool {
            ImageCache.default.isCached(forKey: url.absoluteString)
        }

        let index: Int
        var url: URL {
            URL(string: "https://github.com/onevcat/Flower-Data-Set/raw/master/rose/rose-\(index).jpg")!
        }

        var body: some View {
            HStack(alignment: .center) {
                Spacer()
                KFImage(url, isLoaded: $done)
                    .resizable()
[...]

PR #1510, opened 2020-09-16
Move appear life cycle to placeholder

onevcat:
And also ensure image setting on main thread


onevcat:
This is a workaround for #1508. But a real fix is still being waited from Apple.

hyouuu:
Please revert this change as it causes flickering every time swiftUI re-renders - see details at https://github.com/onevcat/Kingfisher/issues/1549

hyouuu:
@onevcat ^

onevcat:
@hyouuu This fix was intended as a workaround of Apple's bug on iOS 14. However, the flickering seems to be a side effect. I will see how we can solve it.

hyouuu:
@onevcat yes the intention is well understood and appreciated - it's just the side effect is too much :) Hope you can find a more appropriate fix soon and thanks!

PR #1513, opened 2020-09-16
Remove unnecessary capture of binder in KFImage

mpdifran:
This was causing a compile issue of `KingfisherSwiftUI` on version 5.15.1.


mpdifran:
Looks like this compiles in Xcode 12. Should we make a 6.0 release for Xcode 12 changes so those of us still using Xcode 11 can use the latest?

Issue #1517, opened 2020-09-17
SessionDelegate naming conflict between Kingfisher and Alamofire in Xcode 12 GM


sachajln:
### Check List

- [x] I have read the [wiki page](https://github.com/onevcat/Kingfisher/wiki) and [cheat sheet](https://github.com/onevcat/Kingfisher/wiki/Cheat-Sheet), but there is no information I need.
- [x] I have searched in [existing issues](https://github.com/onevcat/Kingfisher/issues?utf8=✓&q=is%3Aissue), but did not find a same one.
- [x] I want to report a problem instead of asking a question. It'd better to use [kingfisher tag in Stack Overflow](http://stackoverflow.com/questions/tagged/kingfisher) to ask a question.

#### What

There seems to be a naming conflict inside the objc header file with the `SessionDelegate` class when using both Kingfisher and Alamofire in Xcode 12 GM. (You can see the issue on Alamofire side [here](https://github.com/Alamofire/Alamofire/issues/3303), more discussion [here](https://stackoverflow.com/questions/63614048/xcode-12-sessiondelegate-has-different-definitions-in-different-modules))

```
/Users/sachajln/ios-workspace/MyProject/Carthage/Build/iOS/Kingfisher.framework/Headers/Kingfisher-Swift.h:836:1: error: 'SessionDelegate' has different definitions in different modules; first difference is definition in module 'Kingfisher.Swift' found method with designater initializer
- (nonnull instancetype)init OBJC_DESIGNATED_INITIALIZER;
^
/Users/sachajln/ios-workspace/MyProject/Carthage/Build/iOS/Alamofire.framework/Headers/Alamofire-Swift.h:772:1: note: but in 'Alamofire.Swift' found method with no designater initializer
- (nonnull instancetype)init SWIFT_UNAVAILABLE;
^
```

#### Reproduce

Use Alamofire and Kingfisher frameworks inside a project and do a `xcodebuild -project YOUR_PROJECT.xcodeproj -sdk iphoneos` with Xcode 12 GM command line tools.

#### Comments

I'm using Carthage to build and copy the frameworks in my project

timbms:
I am having exactly the same problem with latest Xcode official release

Senocico:
The same...

alperkayabasi:
Issue still exists in Xcode 12 official release.

The only solution for now is to rename the "SessionDelegate" to "KFSessionDelegate" or something like that in Kingfisher module.

[...]

PR #1519, opened 2020-09-18
session delegate name change in order to fix conflicts with alamofire on Xcode 12 GM

bogdanmatasaru:
fix for this issue: https://github.com/onevcat/Kingfisher/issues/1517


alperkayabasi:
This is important, a lot of apps are using Kingfisher and Alamofire together.
After Xcode 12 official release, builds fail.

onevcat:
@bogdanmatasaru Thank you for this PR. Your fix works perfectly!

But I guess I can fix it in #1523, with a long-ago intended change to make the framework better.

Issue #1511, opened 2020-09-16
Crashes on SessionDelegate.remove() when canceling Download Task


jgarcia-mirego:
### Check List

Thanks for considering to open an issue. Before you submit your issue, please confirm these boxes are checked.

- [x] I have read the [wiki page](https://github.com/onevcat/Kingfisher/wiki) and [cheat sheet](https://github.com/onevcat/Kingfisher/wiki/Cheat-Sheet), but there is no information I need.
- [x] I have searched in [existing issues](https://github.com/onevcat/Kingfisher/issues?utf8=✓&q=is%3Aissue), but did not find a same one.
- [x] I want to report a problem instead of asking a question. It'd better to use [kingfisher tag in Stack Overflow](http://stackoverflow.com/questions/tagged/kingfisher) to ask a question.

#### What

We are using Kingfisher to download images for our iOS app and I have encountered many crashes when trying to cancel a download task

**EXC_BAD_ACCESS KERN_INVALID_ADDRESS 0x0000000000000020**

```
Crashed: com.apple.main-thread
0  CFNetwork                      0x18436b058 HTTPParser::HTTPParser(__CFAllocator const*, HTTPParserClient*, HTTPParser*) + 44
1  CFNetwork                      0x18424b27c HTTPMessage::commonInitialization(unsigned char, HTTPMessage const*) + 212
2  CFNetwork                      0x18424b27c HTTPMessage::commonInitialization(unsigned char, HTTPMessage const*) + 212
3  CFNetwork                      0x18424b41c HTTPMessage::HTTPMessage(HTTPMessage const*) + 112
4  CFNetwork                      0x18436c1dc HTTPRequestMessage::HTTPRequestMessage(HTTPRequestMessage const*) + 36
5  CFNetwork                      0x18422a74c HTTPRequest::HTTPRequest(HTTPRequest const*) + 24
6  CFNetwork                      0x18422ce8c URLRequest::initialize(URLRequest const*, unsigned char) + 656
7  CFNetwork                      0x1841e2bd8 _createRequestCopy(__CFAllocator const*, _CFURLRequest const*, unsigned char) + 152
8  CFNetwork                      0x184355de0 -[NSURLRequest mutableCopyWithZone:] + 44
9  libswiftFoundation.dylib       0x105a95a0c URLRequest.init(_bridged:) + 77948
[...]

PR #1537, opened 2020-09-28
Cancel moved to avoid crash

Statsminizzle:
I maintain a couple of apps that uses your library to great success.
However there are an annoying crash that happens in both apps - even though the actual implementation of cancelling download tasks is very different in the apps.

A simple way it can be replicated would be to do this:
`func collectionView(_ collectionView: UICollectionView, didEndDisplaying cell: UICollectionViewCell, forItemAt indexPath: IndexPath) {
        if let cell = cell as? TestCell {
            cell.imageView.kf.cancelDownloadTask()
        }
    }`

In that case - if you have a long list - so you can get your scrolling velocity high and maintain that velocity it will eventually crash.

From my debugging the cause seems to be that since the underlying _URLSessionDataTask_ is cancelled prior to the callback then it can actually have become nil when we are later using the task in the callback.

I have tested this solution in both my apps and the crash has disappeared and I have seen no side-effects from this change.

It seems that since I actually began the investigation of this, an issue (https://github.com/onevcat/Kingfisher/issues/1511) has been opened from a different user regarding the same problem I have tried to solve here. At least, when inspecting their crash logs it looks very similar to what I experience in my apps where it crashes in _SessionDelegate.remove()_ by _URLRequest._unconditionallyBridgeFromObjectiveC(_:)_

Let me know what you think.

### DIFF
--- a/Sources/Networking/SessionDataTask.swift
+++ b/Sources/Networking/SessionDataTask.swift
@@ -101,9 +101,6 @@ public class SessionDataTask {
         guard let callback = removeCallback(token) else {
             return
         }
-        if callbacksStore.count == 0 {
-            task.cancel()
-        }
         onCallbackCancelled.call((token, callback))
     }
 
--- a/Sources/Networking/SessionDelegate.swift
+++ b/Sources/Networking/SessionDelegate.swift
@@ -72,6 +72,7 @@ class SessionDelegate: NSObject {
             // No other callbacks waiting, we can clear the task now.
             if !task.containsCallbacks {
                 let dataTask = task.task
+                dataTask.cancel()
                 self.remove(dataTask)
             }
         }


onevcat:
@Statsminizzle Thanks a lot for this!

It looks promising at the first glance. I will soon check and verify it carefully and deeper to see if there is any side-effect (since all the tests are passing so I guess it is fine!).

onevcat:
Checked it.

In the ideal Swift world, that crash should not happen. But when things come to Objective-C bridging or some low layer memory management (as the #1511 indicates a EXC_BAD_ACCESS crash), I am afraid it is an implementation detail that we cannot know easily.

Let's merge it and see if it can fix that issue!

Thank you for your contribution and I will tag it soon.

Issue #1518, opened 2020-09-18
Backward matching of the unlabeled trailing closure is deprecated


lexrus:
### Check List

Thanks for considering to open an issue. Before you submit your issue, please confirm these boxes are checked.

- [x] I have read the [wiki page](https://github.com/onevcat/Kingfisher/wiki) and [cheat sheet](https://github.com/onevcat/Kingfisher/wiki/Cheat-Sheet), but there is no information I need.
- [x] I have searched in [existing issues](https://github.com/onevcat/Kingfisher/issues?utf8=✓&q=is%3Aissue), but did not find a same one.
- [x] I want to report a problem instead of asking a question. It'd better to use [kingfisher tag in Stack Overflow](http://stackoverflow.com/questions/tagged/kingfisher) to ask a question.

### Issue Description

I was just compiling our code base with Xcode 12. I got a dozens warnings sort of:

<img width="849" alt="Screen Shot 2020-09-18 at 3 06 31 PM" src="https://user-images.githubusercontent.com/219689/93567047-94943780-f9c0-11ea-8393-aa0833e8065a.png">

Seems like this new proposal trying to catch my eyes.
https://forums.swift.org/t/se-0286-forward-scan-for-trailing-closures/38529

Okay, I'm good to add the `completionHandler:` label. But after some "auto fix"s and "auto completion"s. Some API callings could be ambiguous to my colleagues, like:

``` language=swift
imageView.kf.setImage(with: LocalFileImageDataProvider(fileURL: URL(string: "")!)) { (_, _) in
    // How do I know this is a progress block from this API perspective?
} completionHandler: { _ in

}
```

This might not be an issue with Kingfisher. But IMO some API improvements could be made.
At least we should prevent Xcode to automatically hide the `progressBlock:`.

I'd prefer something like what Alamofire did:
``` language=swift
AF.download("")
    .downloadProgress { _ in

    }
    .responseJSON { _ in

    }
```

Regards.

onevcat:
It is a quite annoying change in Swift 5.3, basically only for writing better SwiftUI code. For any other use cases, it is a bit of regression to apply the new trailing closure rule.

For you case, if you have both the progress block and completion block, try to explicitly name the first block to avoid the ambiguous meaning:

```swift
imageView.kf.setImage(
    with: LocalFileImageDataProvider(fileURL: URL(string: "")!),
    progressBlock: { (_, _) in

    }
) { _ in

}
```

Or even write out them all:

```swift
[...]

PR #1546, opened 2020-10-04
Feature KF builder

onevcat:
This implements #1518

### DIFF
--- a/Demo/Demo/Kingfisher-Demo/Base.lproj/Main.storyboard
+++ b/Demo/Demo/Kingfisher-Demo/Base.lproj/Main.storyboard
@@ -1,9 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<document type="com.apple.InterfaceBuilder3.CocoaTouch.Storyboard.XIB" version="3.0" toolsVersion="17147" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES" useSafeAreas="YES" colorMatched="YES" initialViewController="peg-r0-mlo">
+<document type="com.apple.InterfaceBuilder3.CocoaTouch.Storyboard.XIB" version="3.0" toolsVersion="17156" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES" useSafeAreas="YES" colorMatched="YES" initialViewController="peg-r0-mlo">
     <device id="retina5_5" orientation="portrait" appearance="dark"/>
     <dependencies>
         <deployment identifier="iOS"/>
-        <plugIn identifier="com.apple.InterfaceBuilder.IBCocoaTouchPlugin" version="17120"/>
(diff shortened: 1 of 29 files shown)


Kaspik:
Thanks for this PR! I have a question about migration tho - specifically about the options (not sure if I should open an issue or if this is the right thing).

Previously, we had an extension where we dynamically added options like this:
```swift
var options: KingfisherOptionsInfo = [.targetCache(ImageCache.default), .downloader(ImageDownloader.default), .backgroundDecode]
if let targetCache = resource?.targetCache.cache {
   options.append(.targetCache(targetCache))
}

if let retryStrategy = retryStrategy {
   let delayStrategy = DelayRetryStrategy(maxRetryCount: retryStrategy.maxRetryCount, retryInterval: .seconds(retryStrategy.retryInterval))
   options.append(.retryStrategy(delayStrategy))
}

if let imageProcessor = imageProcessor {
   options.append(.processor(imageProcessor.processor))
}

return Promise { seal in
   kf.setImage(with: resource?.resource,
        placeholder: placeholderImage,
        options: options,
        completionHandler: { result in
            switch result {
                case let .success(retrieveResult):
                    seal.fulfill(retrieveResult.image)
                case let .failure(error):
                    seal.reject(error)
            }
      })
}
```

With the new KF chaining, the options is not a thing anymore, so I wonder what would be the best way how to migrate to it? Or should we just stick with the existing approach and not migrate as you are not gonna deprecated it _ever_? :) Thanks!

onevcat:
@Kaspik 

As you said, the options are chained in this PR and you can just use it to make your code more compact and readable:

```swift
// It is a sample and maybe it won't compile
KF.url(resource?.resource)
            .placeholder(placeholderImage)
            .targetCache(resource?.targetCache.cache)
            .retry(retry)
            .backgroundDecode()
            .setProcessor(imageProcessor.processor)
            .onSuccess{ seal.fulfill($0.image) }
            .onFailure { seal.reject($0) }
            .set(to: imageView)
```

> I found the `.retry` is not receiving a `nil` value now, so it is a bit hard to use. I will make it receive an optional value in an upcoming release. Before that, you can make a `NoRetry` yourself to make it easier for nil-`retryStrategy` case:
> ```swift
> struct NoRetry: RetryStrategy {
>   func retry(context: RetryContext, retryHandler: @escaping (RetryDecision) -> Void) {
>     retryHandler(.stop)
>   }
> }

I believe compared to the original style, the new one reads better.

It is totally fine to just stick with the existing approach! Adding the new KF chain-able calls only provides a new style for using Kingfisher in a good context, especially it is fully compatible with Kingfisher's SwiftUI support. That means, you can almost just copy the `KF` code from UIKit world to SwiftUI world, then by just changing `KF` to `KFImage`, the same magic will happen in the SwiftUI app.

There is no plan to deprecate the existing `ImageView` extensions methods. So just keep using it without worrying.

Kaspik:
Thanks! Happy to hear there is no plan to deprecate it. We will migrate some of the codebase but will leave these if-else condition-based parts as they are. :)

onevcat:
No problem! I myself is still using a lot of the extension method yet. 😂 (Too lazy...)

PR #1547, opened 2020-10-04
Version 6.0

onevcat:


Sources/General/KF.swift
```diff
@@ -0,0 +1,982 @@
...
+            for: state,
+            placeholder: placeholderImage,
+            parsedOptions: options,
+            progressBlock: progressBlock,
+            completionHandler: resultHandler
+        )
+    }
+    #endif // end of canImport(UIKit)
+
+    #if canImport(AppKit)
```
Brett-Best, line 197: I had to change this to `#if canImport(AppKit) && !targetEnvironment(macCatalyst)`

PR #1558, opened 2020-10-27
Avoid racing for accessing originalRequest

### LINKED ISSUE #1511: Crashes on SessionDelegate.remove() when canceling Download Task

onevcat:
This patch adds/adjusts some lock behaviors. This should help to avoid racing for accessing originalRequest properties and fix #1511

### DIFF
--- a/Sources/Networking/SessionDelegate.swift
+++ b/Sources/Networking/SessionDelegate.swift
@@ -75,7 +75,7 @@ class SessionDelegate: NSObject {
                 let dataTask = task.task
 
                 self.cancelTask(dataTask)
-                self.remove(dataTask)
+                self.remove(task)
             }
         }
         let token = task.addCallback(callback)
@@ -98,11 +98,11 @@ class SessionDelegate: NSObject {
         return DownloadTask(sessionTask: task, cancelToken: token)
     }
 
-    private func remove(_ task: URLSessionTask) {
+    private func remove(_ task: SessionDataTask) {
         lock.lock()
         defer { lock.unlock() }
 
-        guard let url = task.originalRequest?.url else {
+        guard let url = task.originalURL else {
             return
         }
         tasks[url] = nil
@@ -189,7 +189,7 @@ extension SessionDelegate: URLSessionDataDelegate {
     func urlSession(_ session: URLSession, task: URLSessionTask, didCompleteWithError error: Error?) {
         guard let sessionTask = self.task(for: task) else { return }
 
-        if let url = task.originalRequest?.url {
+        if let url = sessionTask.originalURL {
             let result: Result<URLResponse, KingfisherError>
             if let error = error {
                 result = .failure(KingfisherError.responseError(reason: .URLSessionError(error: error)))
@@ -256,7 +256,7 @@ extension SessionDelegate: URLSessionDataDelegate {
         guard let sessionTask = self.task(for: task) else {
             return
         }
-        remove(task)
+        remove(sessionTask)
         sessionTask.onTaskDone.call((result, sessionTask.callbacks))
     }
 }
--- a/Sources/Networking/SessionDelegate.swift
+++ b/Sources/Networking/SessionDelegate.swift
@@ -73,13 +73,20 @@ class SessionDelegate: NSObject {
             // No other callbacks waiting, we can clear the task now.
             if !task.containsCallbacks {
                 let dataTask = task.task
-                dataTask.cancel()
+
+                self.cancelTask(dataTask)
                 self.remove(dataTask)
             }
         }
         let token = task.addCallback(callback)
         tasks[url] = task
         return DownloadTask(sessionTask: task, cancelToken: token)
+    }
+
+    private func cancelTask(_ dataTask: URLSessionDataTask) {
+        lock.lock()
+        defer { lock.unlock() }
+        dataTask.cancel()
     }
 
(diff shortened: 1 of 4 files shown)


Sources/Networking/SessionDelegate.swift
```diff
@@ -92,22 +99,22 @@ class SessionDelegate: NSObject {
     }
 
     private func remove(_ task: URLSessionTask) {
+        lock.lock()
+        defer { lock.unlock() }
+
         guard let url = task.originalRequest?.url else {
```
danielbyon-hulu, line 105: Wouldn't you want to use `task.originalURL` here as well? Also in `task(for:)`, so we only ever call `task.originalRequest?.url` once, and that's in the `SessionDataTask` init. Then I don't think you need to adjust the locks (or at least I don't think it will affect the behavior)

danielbyon-hulu (approved):

PR #1576, opened 2020-12-19
Fixes to enable catching in google cloud.

carles-estevadeordal:
I've made a fix for google storage links that they where not effectively catching, I don't know if you will want this because it adds some host specific logic on the cacheKey. But with this fix it effectively catches google storage api resources...


onevcat:
Hi,

Thanks for this PR. I understood that Google API is appending some query part to the URL, so the default caching mechanism might be invalid due to this.

But I am afraid this PR may bring too much assumption and side effects, and it is really a very specified workaround. So I prefer to not merging it.

Instead, if anyone needs this, it should be easy enough to create an `ImageResource` to implement the same logic. Not like the normal URL resource, `ImageResource` allows you to specify your own cache key, so you can ignore the query there.

Anyway, thank you for the work to point it out and give feedback!

PR #1620, opened 2021-02-06
有没有好的方案，把URLSessionDataDelegate更多的协议实现暴露出来

Jack-Ving:
https://github.com/onevcat/Kingfisher/issues/1619

尝试一种方案。把URLSessionDataDelegate的其他协议方法，转发给业务方定制化处理

### DIFF
--- a/Sources/Networking/ImageDownloader.swift
+++ b/Sources/Networking/ImageDownloader.swift
@@ -165,9 +165,9 @@ open class ImageDownloader {
         setupSessionHandler()
     }
     
-    /// You could set the extral handler before a downloading task starts.
+    /// You could set the extra handler before a downloading task starts.
     public func addExtraSessionDelegateHandler(_ handler: URLSessionDataDelegate) {
-        sessionDelegate.extralHandler = handler
+        sessionDelegate.extraHandler = handler
     }
     
     deinit { session.invalidateAndCancel() }
--- a/Sources/Networking/ImageDownloader.swift
+++ b/Sources/Networking/ImageDownloader.swift
@@ -166,9 +166,9 @@ open class ImageDownloader {
     }
     
     /// You could set the extra handler before a downloading task starts.
-    public weak var extraSessionDelegateHandler:URLSessionDataDelegate? {
-        set {sessionDelegate.setExtraHandler(newValue)}
-        get {return sessionDelegate.getExtraHandler()}
+    public var extraSessionDelegateHandler:URLSessionDataDelegate? {
+        set {sessionDelegate.extraHandler = newValue}
+        get {return sessionDelegate.extraHandler}
     }
     
     deinit { session.invalidateAndCancel() }
--- a/Sources/Networking/ImageDownloader.swift
+++ b/Sources/Networking/ImageDownloader.swift
@@ -164,7 +164,12 @@ open class ImageDownloader {
         authenticationChallengeResponder = self
         setupSessionHandler()
     }
-
+    
+    /// You could set the extral handler before a downloading task starts.
+    public func addExtraSessionDelegateHandler(_ handler: URLSessionDataDelegate) {
+        sessionDelegate.extralHandler = handler
+    }
+    
     deinit { session.invalidateAndCancel() }
 
     private func setupSessionHandler() {
--- a/Sources/Networking/ImageDownloader.swift
+++ b/Sources/Networking/ImageDownloader.swift
@@ -166,8 +166,9 @@ open class ImageDownloader {
     }
     
     /// You could set the extra handler before a downloading task starts.
-    public func addExtraSessionDelegateHandler(_ handler: URLSessionDataDelegate) {
-        sessionDelegate.extraHandler = handler
+    public weak var extraSessionDelegateHandler:URLSessionDataDelegate? {
+        set {sessionDelegate.setExtraHandler(newValue)}
+        get {return sessionDelegate.getExtraHandler()}
     }
     
     deinit { session.invalidateAndCancel() }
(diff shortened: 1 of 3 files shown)


onevcat:
我可能更倾向于把 `SessionDelegate` 写成 `open` 然后通过 subclass 的方式来实现自定义逻辑。
当然，这个 PR 里的实现也是一种办法。我考虑一下哪种会更好。

Sources/Networking/SessionDelegate.swift
```diff
@@ -45,6 +45,7 @@ class SessionDelegate: NSObject {
 
     private var tasks: [URL: SessionDataTask] = [:]
     private let lock = NSLock()
+    public var extralHandler: URLSessionDataDelegate?
```
halonsoluis, line 48: minor typo -> **extralHandler** should be **extraHandler**

Sources/Networking/SessionDelegate.swift
```diff
@@ -45,6 +45,7 @@ class SessionDelegate: NSObject {
 
     private var tasks: [URL: SessionDataTask] = [:]
     private let lock = NSLock()
+    public var extraHandler: URLSessionDataDelegate?
```
onevcat, line 48: At least, maybe we want it a `weak` property?

onevcat: `SessionDelegate` is not a public class yet. So I don't see a reason to mark this as `public`.

Sources/Networking/ImageDownloader.swift
```diff
@@ -164,7 +164,12 @@ open class ImageDownloader {
         authenticationChallengeResponder = self
         setupSessionHandler()
     }
-
+    
+    /// You could set the extra handler before a downloading task starts.
+    public func addExtraSessionDelegateHandler(_ handler: URLSessionDataDelegate) {
```
onevcat, line 169: Can we expose a setter/getter to `sessionDelegate.extraHandler`, which would be more flexible.

Sources/Networking/SessionDelegate.swift
Jack-Ving: 非常感谢，我的开发陋习，没有仔细考虑，我突然感觉学到了，我改下weak和private。

Sources/Networking/ImageDownloader.swift
Jack-Ving: 非常感谢，我为了方便，直接public了，没有仔细考虑，确实不符合`SessionDelegate is not a public class yet`。 学到了，多谢。

Sources/Networking/ImageDownloader.swift
```diff
@@ -164,7 +164,13 @@ open class ImageDownloader {
         authenticationChallengeResponder = self
         setupSessionHandler()
     }
-
+    
+    /// You could set the extra handler before a downloading task starts.
+    public weak var extraSessionDelegateHandler:URLSessionDataDelegate? {
```
onevcat, line 169:
```suggestion
    public var extraSessionDelegateHandler:URLSessionDataDelegate? {
```

For a getter/setter, there should be no need to mark it as `weak`.

Sources/Networking/SessionDelegate.swift
```diff
@@ -145,6 +146,15 @@ class SessionDelegate: NSObject {
...
     }
+    
+    /// You could set the extra handler before a downloading task starts.
+    func setExtraHandler(_ handler: URLSessionDataDelegate?) {
+        extraHandler = handler
+    }
+    func getExtraHandler() -> URLSessionDataDelegate? {
+        return extraHandler
+    }
+
```
onevcat, line 157: I prefer to remove these methods and make `extraHandler` to `internal`, so we can just use the property without this boilerplate method calling.

Sources/Networking/ImageDownloader.swift
Jack-Ving: 好的，多谢！

onevcat:
Nice!

onevcat:
It seems that these changes cause an error while loading.

```
2021-03-09 11:51:23.423663+0900 Kingfisher-Demo[44705:9436274] 44705: CFNetwork internal error (0xc01a:/Library/Caches/com.apple.xbs/Sources/CFNetwork_Sim/CFNetwork-1220.1/Loading/Connection/Loader Side/URLConnectionLoader.cpp:266)

```

I am not sure what is happening now, but reverting these changes can solve the issue. I am going to revert it and tag a patch version. I will also keep digging to see what is the problem/

Sources/Networking/SessionDelegate.swift
```diff
@@ -259,4 +261,43 @@ extension SessionDelegate: URLSessionDataDelegate {
...
+    }
+    @available(iOS 9.0, OSX 10.11, tvOS 9.0, watchOS 2.0, *)
+    func urlSession(_ session: URLSession, dataTask: URLSessionDataTask, didBecome streamTask: URLSessionStreamTask) {
+        extraHandler?.urlSession?(session, dataTask: dataTask, didBecome: streamTask)
+    }
+    func urlSession(_ session: URLSession, dataTask: URLSessionDataTask, didBecome downloadTask: URLSessionDownloadTask) {
+        extraHandler?.urlSession?(session, dataTask: dataTask, didBecome: downloadTask)
+    }
+    func urlSession(_ session: URLSession, task: URLSessionTask, needNewBodyStream completionHandler: @escaping (InputStream?) -> Void) {
+        extraHandler?.urlSession?(session, task: task, needNewBodyStream: completionHandler)
```
dreampiggy, line 290: If you implements the URLSessionDelegate method with completionBlock, you `MUST` callback the handler, even if you don't have anything logic.

PR #1627, opened 2021-02-12
Fix a bug on iOS 13 swiftUI

HugoSay:
Using the mainCurrentOrAsync queue causes the current warning on iOS 13
"Modifying state during view update, this will cause undefined behavior"
This seems to hapen if the image is already loaded and the callback is called immediately during the view render.

Using mainAsync lets the view view finish its rendering before modifying its state

### DIFF
--- a/Sources/SwiftUI/ImageBinder.swift
+++ b/Sources/SwiftUI/ImageBinder.swift
@@ -109,7 +109,7 @@ extension KFImage {
                             let r = RetrieveImageResult(
                                 image: image, cacheType: value.cacheType, source: value.source, originalSource: value.originalSource
                             )
-                            CallbackQueue.mainCurrentOrAsync.execute {
+                            CallbackQueue.mainAsync.execute {
                                 done(.success(r))
                             }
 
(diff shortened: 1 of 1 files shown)


HugoSay:
It's an issue I have on my app, should I create a sample project demonstrating the issue ? 
I don't know if I could reproduce this with a unit test sadly 😟

onevcat:
In the current version, the `ImageBinder.start` should only be called in `onAppear` of the `KFImage` rendering body, which should not be during the view update.

The `mainCurrentOrAsync` invocation is used for avoiding unexpected flickering (when reloading the image from cache), so it is necessary for now. 

As I tested, I myself didn't see a warning of this when running the sample app on iOS 13. Is there any condition to reproduce this warning? May I know the Kingfisher version and Xcode version you are using?

Thanks!

akmarinov:
I'm seeing the same issue on iOS 13, though this PR doesn't address it.
I have a `List` with cells that have a `KFImage` and some text.

Kingfisher 6.1.1
Xcode 12.4

HugoSay:
Hi, sorry for the late response,
I'll try to commit some code that triggers the issue. 

Thanks for your feedback @akmarinov, I'll try to come with another fix for this issue then ! 👍

onevcat:
A sample or a view can reproduce it would be welcomed! I am glad to check in detail on what's going on once I can reproduce it.

HugoSay:
@akmarinov if you'd care to test https://github.com/onevcat/Kingfisher/pull/1641 I'd be curious to know if this fixes your issue

PR #1641, opened 2021-03-01
Fix issue on iOS 13

HugoSay:
Update from #1627

Some race conditions could cause the load of the image to trigger "Modifying state during view update, this will cause undefined behavior" on the onApear call.

Moved the logic to the Binder in and makes it an observable object in order to reduce the work done in the body and prevent some potential issues

### DIFF
--- a/Sources/SwiftUI/ImageBinder.swift
+++ b/Sources/SwiftUI/ImageBinder.swift
@@ -33,8 +33,9 @@ extension KFImage {
 
     /// Represents a binder for `KFImage`. It takes responsibility as an `ObjectBinding` and performs
     /// image downloading and progress reporting based on `KingfisherManager`.
-    class ImageBinder {
+    class ImageBinder: ObservableObject {
 
+        @Published var loadedImage: Image?
         let source: Source?
         var options = KingfisherParsedOptionsInfo(KingfisherManager.shared.defaultOptions)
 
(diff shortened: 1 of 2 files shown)


onevcat:
Thanks for the PR.

It is basically what Kingfisher did to load an image for SwiftUI before version 6.

Using an `ObservableObject` helps to make the state management better, but since every time the view reused (in a `List` or `LazyVStack`) would trigger a recreation of the binder, the image result will be lost and there is no chance for Kingfisher to load it again, due to the `onAppear` will not be called twice for the view.

Currently Kingfisher is using a local `@State` to bypass this problem. I know it is not ideal, but it is the best way what I can think out to handle it now. The real solution is using `@StateObject` from iOS 14. But it is not easy to provide it as an API for both iOS 13 and iOS 14.

onevcat:
BTW, I personally didn't see this warning myself when running on an iOS 13 device. Is there a sample view to reproduce it? Are you still using iOS 13 SDK?

HugoSay:
As I said in my last PR It's an issue that I reproduce in my main project but I couldn't pinpoint how to reproduce it in a sample project yet. (Did not take the time to do so yet) 

I understand your point about the observable object, although, when trying the updated code with the sample app I didn't run into any issue on the swiftUI List view.

onevcat:
Ummm,

The behavior is now quite different from when I introduced the `@State` based image view. Maybe you are right on this. I will check what's going on!

Thanks for the innovation.

onevcat:
@HugoSay 

I applied some similar changes in the branch `fix/unexpected-ref` (which also fixed another state setting issue). Maybe you can have a try to see if the warning is fixed or not.

This PR contains some unintended changes so I guess it might be quite hard to merge without breaking existing things. So let's continue this in #1642

Issue #1621, opened 2021-02-09
KF image lags when scrolling inside GeometryReader


ForceGT:
The code snippet is this

```
ScrollView(.vertical){
            
            VStack(alignment: .leading){
                //HEADER IMAGE
                GeometryReader{ geometry in
                    
                    if geometry.frame(in: .global).minY > -360{
                        
                        KFImage(URL(string: wallpaper.image ?? Config.DUMMY_URL))
                            .blur(radius: CGFloat(blurRadius), opaque: true)
                            .aspectRatio(contentMode: showImage ? .fit :.fill)
                            .offset(y: (-geometry.frame(in: .global).minY))
                            .frame(width: UIScreen.main.bounds.width,
                                   height: geometry.frame(in: .global).minY > 0 ? geometry.frame(in: .global).minY + 360 : 360, alignment: .center)
                            .onTapGesture {
                                withAnimation(.easeInOut(duration: 0.75)) {
                                    showImage.toggle()
                                }
                            }
                    }
                }// :HEADER IMAGE
                .frame(width:UIScreen.main.bounds.width,height: 360,alignment: .center)
```

ForceGT:
Posting a screen record just in case 

https://user-images.githubusercontent.com/31688051/107320658-28c57f00-6ac7-11eb-80c1-6ba97c313fdf.mov

onevcat:
@ForceGT Not sure how this can happen. 

Recently we had a few changes on how to identify the loaded image, which is possible to this issue. Maybe it worths trying the `master` branch to see if there is any difference?

Thanks!

justintime06:
Using the master branch worked for me, thank you so much!!!!

onevcat:
The same fixes should be already released as a part of 6.1.1. I am closing it now. If the issue is still there, feel free to reopen it.

Issue #1618, opened 2021-02-05
KFImage image disappear when scroll


kaich:
Use `KFImage`  in cell of  `LazyVGrid`.  When I scroll the grid, the loaded image disappeared and never show again.

dopefaceee:
yes smae here

onevcat:
Any code snippet? I tried the code below and everything works fine:

```swift
struct ContentView : View {

    let index = 1 ..< 700
    let columns = Array<GridItem>(repeating: GridItem(.fixed(100)), count: 3)
    var body: some View {
        ScrollView {
            LazyVGrid(columns: columns, content: {
                ForEach(index) { i in
                    KFImage
                        .url(URL(string: "https://github.com/onevcat/Flower-Data-Set/raw/master/rose/rose-\(i).jpg")!)
                        .resizable()
                        .scaledToFit()
                }
            })
        }
    }
}
```

dopefaceee:
Using KingFisher v6.1.0

```swift
struct PokedexView: View {
    private let gridItems = [GridItem(.flexible()), GridItem(.flexible())]
    @ObservedObject var viewModel = PokeomonViewModel()
    var body: some View {
        NavigationView {
            ScrollView {
                LazyVGrid(columns: gridItems, spacing: 16) {
                    ForEach(viewModel.pokemon) { pokemon in
                        PokemonCell(pokemon: pokemon, viewModel: viewModel)
                    }
                }
            }
            .navigationTitle("Pokedex")
        }
    }
}
```
```swift
struct PokemonCell: View {
    let pokemon: Pokemon
    let viewModel: PokeomonViewModel
    let backgroundColor: Color
    
    init(pokemon: Pokemon, viewModel: PokeomonViewModel) {
        self.pokemon = pokemon
        self.viewModel = viewModel
        self.backgroundColor = Color(viewModel.backgroundColor(forType: pokemon.type))
    }
    
    var body: some View {
        ZStack {
            VStack(alignment: .leading) {
                Text(pokemon.name.capitalized)
                    .font(.headline)
                    .foregroundColor(.white)
                    .padding(.top, 8)
                    .padding(.leading)
                HStack {
                    Text(pokemon.type)
                        .font(.subheadline).bold()
                        .foregroundColor(.white)
                        .padding(.horizontal, 16)
                        .padding(.vertical, 8)
                        .overlay(
                            RoundedRectangle(cornerRadius: 20)
                                .fill(Color.white.opacity(0.25))
[...]

Issue #1603, opened 2021-01-21
Crash in taskIdentifier.get


hijamoya:
### Check List

Thanks for considering to open an issue. Before you submit your issue, please confirm these boxes are checked.

- [x] I have read the [wiki page](https://github.com/onevcat/Kingfisher/wiki) and [cheat sheet](https://github.com/onevcat/Kingfisher/wiki/Cheat-Sheet), but there is no information I need.
- [x] I have searched in [existing issues](https://github.com/onevcat/Kingfisher/issues?utf8=✓&q=is%3Aissue), but did not find a same one.
- [x] I want to report a problem instead of asking a question. It'd better to use [kingfisher tag in Stack Overflow](http://stackoverflow.com/questions/tagged/kingfisher) to ask a question.

### Issue Description

We receive a crash after upgrade to 6.0.1, this never happened before 6.0.0 version.

Here is the trace:

```
Hardware Model:     iPhone9,2
Process:            shopping
Identifier:         com.lativ.iShopping.CN
Version:            2.2.4
Role:               Foreground
OS Version:         iOS 11.2.6
Exception Type:     EXC_BAD_ACCESS 


EXC_BAD_ACCESS: Attempted to dereference garbage pointer 0xa6d09bec8.

0  libobjc.A.dylib         _objc_retain
1  Kingfisher              merged generic specialization <Kingfisher.Box<Kingfisher.Indicator>> of Kingfisher.getAssociatedObject<A>(Any, Swift.UnsafeRawPointer) -> A?
2  shopping                taskIdentifier.get (ImageView+Kingfisher.swift:436:54)
3  shopping                onDataReceived (KingfisherOptionsInfo.swift:382:20)
4  shopping                Kingfisher.SessionDelegate.urlSession(_: __C.NSURLSession, dataTask: __C.NSURLSessionDataTask, didReceive: Foundation.Data) -> () (SessionDelegate.swift:184:28)
5  <compiler-generated>    urlSession
6  CFNetwork               ___67-[NSURLSession delegate_dataTask:didReceiveData:completionHandler:]_block_invoke.264
7  Foundation              ___NSBLOCKOPERATION_IS_CALLING_OUT_TO_A_BLOCK__
8  Foundation              -[NSBlockOperation main]
9  Foundation              -[__NSOperationInternal _start:]
10 libdispatch.dylib       __dispatch_client_callout
11 libdispatch.dylib       __dispatch_block_invoke_direct$VARIANT$mp
[...]

PR #1642, opened 2021-03-02
Drop `@State` based image result setting

onevcat:
Storing and setting the image with a `@State` is tricky and turns out to be not stable.

This PR reverted to use the `@ObservedObject` with a new `loadImmediately` method to trigger the source loading manually, to give framework users a chance to control the loading process to prevent flickering or some other bugs with `ObservedObject`.

Ref: #1603 #1618 #1621 #1641


HugoSay (commented):
Thanks for that quick pr and your answers ! 😃

Sources/SwiftUI/ImageBinder.swift
```diff
@@ -102,21 +103,17 @@ extension KFImage {
...
-                            )
-                            CallbackQueue.mainCurrentOrAsync.execute {
-                                done(.success(r))
-                            }
+
+                            self.isLoaded.wrappedValue = true
+
+                            let animation = self.fadeTransitionDuration(cacheType: value.cacheType)
+                                .map { duration in Animation.linear(duration: duration) }
+                            withAnimation(animation) { self.loaded = true }
```
HugoSay, line 112: shouldn't this block be called from the `mainCurrentOrAsync` callback queue ?

onevcat: Yes. You are right!

PR #1647, opened 2021-03-04
Add imageFrameCount

danielebogo:
Refs. #1645

## What it does

This PR adds the property `imageFrameCount` to the a `KFCrossPlatformImage`. This property is an optional _Int_ and it's set by default when an animated image is created. Since the `KFCrossPlatformImage` already stores an `imageSource` we can retrieve easily the frame count and store it.

This property is the equivalent of `animatedImageFrameCount` used in `SDAnimatedImage`

I didn't set the property as read only to give the possibility to custom processors to also set the value.

## How to test

- Unit test should be 🟢

### DIFF
--- a/Sources/Image/Image.swift
+++ b/Sources/Image/Image.swift
@@ -43,12 +43,18 @@ import CoreGraphics
 import ImageIO
 
 private var animatedImageDataKey: Void?
+private var imageFrameCountKey: Void?
 
 // MARK: - Image Properties
 extension KingfisherWrapper where Base: KFCrossPlatformImage {
     private(set) var animatedImageData: Data? {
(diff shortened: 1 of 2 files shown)


danielebogo:
@onevcat not sure why CI is failing on
```
KingfisherManagerTests.testRetrievingAlternativeSourceCanCancelUpdatedTask()
KingfisherManagerTests.testRetrievingAlternativeSourceCanCancelUpdatedTask()
```
All tests are passing locally for me

onevcat:
Hi, thanks for it. It looks nice.

The failing tests are due to a CI performance issue on feature tests. I will see what I can do.

danielebogo:
Thanks @onevcat ! CI is all 🟢 now!

PR #1671, opened 2021-03-29
为文件缓存hash之后的文件名自动添加后缀

tctony:
#1670


onevcat:
It solves a partial of #1670, but once the key changed to something else (such appended by a non-default processor, the key would become something like `image.jpg@size123`), this would be not valid anymore.

The feature is useful itself, and I will see if there is a better way to implement it.

PR #1676, opened 2021-04-05
Modify ImageDownloaderDelegate to accept URLReponse

FaizanDurrani:
Changed `imageDownloader(_:didDownload:for:)` to `imageDownloader(_:didDownload:with:)` accepting the returned URLResponse from the request.

This allows modifying the image data (like decrypting) based on the response the server sends (for example a specific header)

### DIFF
--- a/Sources/Networking/ImageDownloader.swift
+++ b/Sources/Networking/ImageDownloader.swift
@@ -194,10 +194,10 @@ open class ImageDownloader {
             }
         }
         sessionDelegate.onDidDownloadData.delegate(on: self) { (self, task) in
-            guard let url = task.originalURL else {
+            guard let response = task.task.response else {
                 return task.mutableData
             }
-            return (self.delegate ?? self).imageDownloader(self, didDownload: task.mutableData, for: url)
+            return (self.delegate ?? self).imageDownloader(self, didDownload: task.mutableData, with: response)
         }
     }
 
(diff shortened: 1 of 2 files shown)


onevcat:
@FaizanDurrani 

Thanks for this PR.

I believe it is a good direction and I'd like to have this in the new version. However, the implementation here breaks the public API and it requires a major upgrade. I am not sure if it worths a major version for just this commit. Can you try to keep the original protocol method (and mark it as deprecated), while adding a new delegate method? It would be nice that all the current behavior not changed.

Sure. If you do not have time for this, please feel free to let me know and I can also do it later when I can find more spare time.

Best regards.

FaizanDurrani:
Updated the PR to not break the existing implementations. that being said, there is a slight catch to this, intead of passing the `originalUrl` the url is taken from the `URLResponse` which is the final url after all the redirects, this may or may not be an issue.

A potential fix for it would be to also pass in the original URLRequest or the whole URLSessionDataTask instead of just the response so the method signature becomes:

`imageDownloader(_: ImageDownloader, didDownload data: Data, for task: URLSessionDataTask) -> Data?`

onevcat:
Yes. It would be nicer and should be no harm to pass all the task here. I will try to improve it, add documentation and tests after merging it.

Thanks!

Issue #1679, opened 2021-04-12
Use AnimatedImageView gif stop play in UITableView cell


liangdahong:
### Check List

Thanks for considering to open an issue. Before you submit your issue, please confirm these boxes are checked.

- [x] I have read the [wiki page](https://github.com/onevcat/Kingfisher/wiki) and [cheat sheet](https://github.com/onevcat/Kingfisher/wiki/Cheat-Sheet), but there is no information I need.
- [x] I have searched in [existing issues](https://github.com/onevcat/Kingfisher/issues?utf8=✓&q=is%3Aissue), but did not find a same one.
- [x] I want to report a problem instead of asking a question. It'd better to use [kingfisher tag in Stack Overflow](http://stackoverflow.com/questions/tagged/kingfisher) to ask a question.

#### What

Use AnimatedImageView in the cell of UITableView and set runLoopMode to .default at the same time. When the cell is clicked, the Gif will stop playing. sorry for the poor English, Thanks.

#### Other Comment

[KFDemo1.zip](https://github.com/onevcat/Kingfisher/files/6297788/KFDemo1.zip)

<img  width="22%" src="https://user-images.githubusercontent.com/12118567/114414500-70f54080-9be1-11eb-854b-06cd6d6b9ce1.gif"/>

liangdahong:
My colleague found the reason, Because the highlighted image does not support animated images. SDWebImage is processed as follows.
- https://github.com/SDWebImage/SDWebImage/blob/master/SDWebImage/Core/SDAnimatedImageView.m (406 lines)
![image](https://user-images.githubusercontent.com/12118567/114890783-e78e7a00-9e3d-11eb-9b68-19c52e83ac6e.png)

onevcat:
Oh, that is really a hidden thing 😱 I did not know it until now.

Do you want to create a PR for this or check if there is any chance to make it work?

liangdahong:
I am trying to create PR 😄.

onevcat:
I guess we can close it as #1685 merged. I will tag a new release for it soon. Thank you for the PR!

PR #1685, opened 2021-04-15
fix：AnimatedImageView Highlight gif stop play

liangdahong:
- https://github.com/onevcat/Kingfisher/issues/1679


onevcat:
Do not worry about the tests. They are some GitHub runner issues.

PR #1686, opened 2021-04-16
[FIX] Crash for GIF with a lot of frames

### LINKED ISSUE #1381: Crash in decoded(scale:)
### Issue Description

I've encountered this problem when the app was running in the background, and this issue was reported from users in crashlytics.

#### Reproduce

```
Kingfisher                     0x101de3760 specialized Kingfisher<A>.decoded(scale:) + 751 (Image.swift:751)
Kingfisher                     0x101de81e8 closure #1 in closure #3 in ImageCache.retrieveImage(forKey:options:completionHandler:) + 724 (Image.swift:724)
Kingfisher                     0x101de56a4 thunk for @escaping @callee_guaranteed () -> () + 4392113828 (<compiler-generated>:4392113828)
```

This exception occurs inside `processImage (604) ImageDownloader.swift` and crashed in Kingfisher<A>.decoded(scale:) + 751 with attempt unwrap nil value (makeImage from CGContext returns nil)
Unfortunately, I can't provide you with the URL. because it stores on the local device.

wow-such-amazing:
There are several things I did :) 

**GIFHeavyViewController**
Contains an example that I used for testing. Previously this controller was causing crashes on my iPhone X.

**GraphicsContext**
An enum with static functions to handle logic for begin/context/end. Basically it is a copy of the code from the ImageDrawing. But this way the code could be reused both inside Animator and ImageDrawing. Right now I use it only for the code that I implemented. But if you feel like keeping this component I can update ImageDrawing. I hope GraphicsContext doesn't break Mac OS implementation. Tell me if I messed up something 😄 

**Animator**
To fix the crash I came with an idea, that the crash occurs due to the fact that CGContext gets freed when we still try to use it. And I assumed that this could happen due to the frequent calls of begin/end on the context. So instead of calling begin/end each time we want a decoded frame, I decided to begin the context in the Animator initialiser and end the context in the Animator de-initialiser. And after these changes GIFHeavyViewController doesn't crash the app anymore. Plus looks like CPU usage decreased after these changes.

This is a raw implementation. I opened this PR to gather some initial feedback so that I don't code something wrong. Let me know what do you think 🙂 

Fixes #1381

### DIFF
--- a/Sources/Image/ImageDrawing.swift
+++ b/Sources/Image/ImageDrawing.swift
@@ -290,21 +290,22 @@ extension KingfisherWrapper where Base: KFCrossPlatformImage {
             
             return vImage_Buffer(data: data, height: height, width: width, rowBytes: rowBytes)
         }
-        
-        guard let context = beginContext(size: size, scale: scale, inverting: true) else {
+        GraphicsContext.begin(size: size, scale: scale)
+        guard let context = GraphicsContext.current(size: size, scale: scale, inverting: true, cgImage: cgImage) else {
             assertionFailure("[Kingfisher] Failed to create CG context for blurring image.")
             return base
         }
         context.draw(cgImage, in: CGRect(x: 0, y: 0, width: w, height: h))
-        endContext()
+        GraphicsContext.end()
         
         var inBuffer = createEffectBuffer(context)
         
-        guard let outContext = beginContext(size: size, scale: scale, inverting: true) else {
+        GraphicsContext.begin(size: size, scale: scale)
+        guard let outContext = GraphicsContext.current(size: size, scale: scale, inverting: true, cgImage: cgImage) else {
             assertionFailure("[Kingfisher] Failed to create CG context for blurring image.")
             return base
         }
-        defer { endContext() }
+        defer { GraphicsContext.end() }
         var outBuffer = createEffectBuffer(outContext)
         
         for _ in 0 ..< iterations {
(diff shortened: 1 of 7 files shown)


wow-such-amazing:
There is one thing I noticed. I tested this implementation in our app with several other heavy GIFs and I found that the main thread becomes loaded almost 100% in some cases. 

Do you think we can find a way to off-load this work to another queue? Will it make sense to you?

Though it doesn't crash which is definitely good 😄 

Screenshot 2021-04-16 at 18 35 18

onevcat:
Hi, @wow-such-amazing 

I can verify the crash and fix in the PR! That just works and for now I didn't inspect a regression on other parts. Great work!

Before we can merge it, I have to ask for some modification.

**GIFHeavyViewController**

This is nice and we definitely want it in the demo app. But can you remove the GifHeavy.gif from the app bundle and load it from the URL below? This helps keep the repo size small for downloading.

https://raw.githubusercontent.com/onevcat/Kingfisher-TestImages/master/DemoAppImage/GIF/GifHeavy.gif

And please remember a commit squash to remove this file from commit history.

** GraphicsContext**

Yes, it is a good start point to copy it as new methods. But I still believe we can reuse a lot of it. I prefer a refactoring for extracting common parts at least, to avoid any duplicated code.

The macOS implementation looks good to me. I will try to do some more regression tests when it prepared.

**Animator**

It is nice!

> Do you think we can find a way to off-load this work to another queue? Will it make sense to you?

It truly worths. But I think the image drawing of `Animator` is already happening in another thread. So before we can start, we need to know what are the operations in main thread causing a high CPU usage. Do you have any clue or instrument result?

wow-such-amazing:
Thanks for the feedback, I will update and push changes :) 

Regarding the Main Thread usage, looks like it comes from the CADisplayLink. I guess when GIF has a lot of frames with frequent changes it affects the Main Thread. So I'm not sure it we can do something about that 🤔

Screenshot 2021-04-19 at 11 01 05

wow-such-amazing:
I'll have a look at the failed test a bit later :)

onevcat:
It is some GitHub CI machine issues please ignore it for now!

wow-such-amazing:
Alright, then it's good for review I think 😄

Sources/Image/ImageDrawing.swift
```diff
@@ -457,55 +458,42 @@ extension KingfisherWrapper where Base: KFCrossPlatformImage {
...
-    }
-    
-    func endContext() {
-        #if os(macOS)
-        NSGraphicsContext.restoreGraphicsState()
+    /// Returns decoded image of the `base` image at a given scale. It will draw the image in a plain context and
+    /// return the data from it. This could improve the drawing performance when an image is just created from
+    /// data but not yet displayed for the first time.
+    ///
+    /// - Parameter context: The context for frawing.
```
onevcat, line 466:
```suggestion
    /// - Parameter context: The context for drawing.
```

onevcat:
I guess it's all done and I am going to merge it. Cheers! 🎉 

@wow-such-amazing Thanks for this fantastic work!

wow-such-amazing:
Yay! Awesome 🕺

Issue #1657, opened 2021-03-11
Support animated images (like GIFs) in SwiftUI


wilg:
Currently, the SwiftUI integration does not support animated images at all. It seems like the easiest thing to do would be just to use the existing AnimatedImageView wrapped in UIViewRepresentable if the image is animated.

This would be really great to have since there aren't any other good solutions for SwiftUI GIFs at the moment.

Binlogo:
It's a nice idea and easy way to wrap AnimatedImageView using UIViewRepresentable, for supporting SwiftUI GIFs animated image display. 

```swift
@available(iOS 13.0, *)
struct KFAnimatedImage: UIViewRepresentable {
    
    var resource: Resource?

    func makeUIView(context: Context) -> AnimatedImageView {
        return AnimatedImageView()
    }
    
    func updateUIView(_ uiView: AnimatedImageView, context: Context) {
        uiView.kf.setImage(with: resource)
    }
    
}
```

Usage: 

```swift
@available(iOS 13.0, *)
struct AnimatedImageDemo: View {
    
    var body: some View {
        KFAnimatedImage(resource: ImageLoader.gifImageURLs.first!)
    }

}
```

It works. 

![image](https://user-images.githubusercontent.com/7845507/116273925-c2422a00-a7b4-11eb-9432-baa0800aaa31.png)


Though, we could try a more SwiftUI way to implement a KFAnimatdImage. 👀

onevcat:
@Binlogo Thanks for pointing it out and these code snippets.

Actually it is not hard to wrap it as a `UIViewRepresentable`. Providing all the supported options is also possible and I will try to add it to Kingfisher later. (If you already have a version, I am glad to receive a PR!)

> Though, we could try a more SwiftUI way to implement a `KFAnimatdImage`. 👀

We can try to make the API of the wrapper type better in SwiftUI, but implementing it all by SwiftUI is not an option. It is not that fitting in the today's SwiftUI concept yet, at least for now, I am afraid.

Binlogo:
@onevcat I just made a quick demo, if wrapping the `AnimatedImageView` as `UIViewRepresentable` is ok, I'd like to have a try to provide a rough PR for discussion.

sslash:
How would you set/call the `resizable` property on a KFANimatedImage? The modifier doesn't seem to exist, but I'm not sure if it's accessible through another property on KFANimatedImage?

E.g normally I'd do this to render an image:
```
KFImage(url).resizable()
[...]

PR #1701, opened 2021-04-29
Support animated images in SwiftUI

Binlogo:
Solves: #1657

### DIFF
--- a/Demo/Demo/Kingfisher-Demo/SwiftUIViews/AnimatedImageDemo.swift
+++ b/Demo/Demo/Kingfisher-Demo/SwiftUIViews/AnimatedImageDemo.swift
@@ -1,0 +1,83 @@
+//
+//  AnimatedImageDemo.swift
+//  Kingfisher
+//
+//  Created by wangxingbin on 2021/4/27.
+//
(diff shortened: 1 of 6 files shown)


onevcat:
Oh! You are really fast. Thank you so much!

I am not sure if all these things are supported in the animated image view, since it is so different from a normal image view and I expected a very simple wrapper just around the `AnimatedImageView`. But I will take a look at your implementation and see if all work.

Good job!

Binlogo:
Got your point. If just a single wrapper, it can hardly interact with other view or do other operations. 

I see a better implementation is on the go in your mind. 👀 Expect to take some improve advice.

⚠️ By the way, an known issue: after setting a processor, the processor works as expected, but animated image would be static.

onevcat:
It seems to be a few code duplication. I've done some refactoring and opened #1705 for the future work. @Binlogo  Thanks for providing this starting point and adding this nice feature!

buluoray:
Does 6.3.x support KFAnimatedView?

onevcat:
No. It is a feature added in 7.0, which is currently in beta.

PR #1705, opened 2021-05-09
SwiftUI animated image view

onevcat:


onevcat:
@wilg It is not possible in current Kingfisher's implementation now. You have to determine which type it is before you can create a correct type of view.

I remembered there is a library which can determine the file type from a URL, by only loading the first few bytes. But I cannot find it now. If you use that, it should be quite quick to know the format of a remote file and then build the view with correct type.

If you cannot find it either, maybe it could be a good idea to build one, or even embed this feature in Kingfisher. If you'd like, you can open an issue so I can track it as my backlog.

Thanks!

PR #1724, opened 2021-06-10
Modified for convention

wogus3602:
Usually return nothing is written as () -> Void.

However, in some places, it is written as ()->() expression.

How about editing for conventions?


onevcat:
Nice! `() -> Void` is always preferred. Thanks for catching it!

PR #1763, opened 2021-07-29
Use processing queue to fetch data in KingfisherManager

ConfusedVorlon:
I'm using Kingfisher to provide thumbnails for a collection view browsing interface for images on my local drive.
The images are large .heic images (typically 40mb)

Setting the async image using the following:

```
imageView?.kf.setImage(with: self.photoItem?.thumbUrl,
                               options:ImageBrowserViewItem.imageProcessorOptions)
```

results in significant blocking on the main thread while `provider.data {}` is called. 
The user interface becomes unusable for significant periods of time. Obviously this is exactly what `kf.setImage` is trying to avoid

all I have done is push the provider.data onto the processingQueue which seems to do the job nicely.

I haven't run tests as none appear in the XCode interface when I open the package. If there is a simple way to run them manually, then it would be great to have some details in contributing.md

thank you

### DIFF
--- a/Sources/General/ImageSource/ImageDataProvider.swift
+++ b/Sources/General/ImageSource/ImageDataProvider.swift
@@ -89,8 +89,10 @@ public struct LocalFileImageDataProvider: ImageDataProvider {
     /// The key used in cache.
     public var cacheKey: String
 
-    public func data(handler: (Result<Data, Error>) -> Void) {
-        handler(Result(catching: { try Data(contentsOf: fileURL) }))
+    public func data(handler:@escaping (Result<Data, Error>) -> Void) {
+        DispatchQueue.global(qos: .userInitiated).async {
+            handler(Result(catching: { try Data(contentsOf: fileURL) }))
+        }
     }
 
     /// The URL of the local file on the disk.
--- a/Sources/General/KingfisherManager.swift
+++ b/Sources/General/KingfisherManager.swift
@@ -344,33 +344,35 @@ public class KingfisherManager {
         completionHandler: ((Result<ImageLoadingResult, KingfisherError>) -> Void)?)
     {
         guard let  completionHandler = completionHandler else { return }
-        provider.data { result in
-            switch result {
-            case .success(let data):
-                (options.processingQueue ?? self.processingQueue).execute {
-                    let processor = options.processor
-                    let processingItem = ImageProcessItem.data(data)
-                    guard let image = processor.process(item: processingItem, options: options) else {
-                        options.callbackQueue.execute {
-                            let error = KingfisherError.processorError(
-                                reason: .processingFailed(processor: processor, item: processingItem))
-                            completionHandler(.failure(error))
+        (options.processingQueue ?? self.processingQueue).execute {
+            provider.data { result in
+                switch result {
+                case .success(let data):
+                    (options.processingQueue ?? self.processingQueue).execute {
+                        let processor = options.processor
+                        let processingItem = ImageProcessItem.data(data)
+                        guard let image = processor.process(item: processingItem, options: options) else {
+                            options.callbackQueue.execute {
+                                let error = KingfisherError.processorError(
+                                    reason: .processingFailed(processor: processor, item: processingItem))
+                                completionHandler(.failure(error))
+                            }
+                            return
                         }
-                        return
+                        
+                        options.callbackQueue.execute {
+                            let result = ImageLoadingResult(image: image, url: nil, originalData: data)
+                            completionHandler(.success(result))
+                        }
                     }
-
+                case .failure(let error):
                     options.callbackQueue.execute {
-                        let result = ImageLoadingResult(image: image, url: nil, originalData: data)
-                        completionHandler(.success(result))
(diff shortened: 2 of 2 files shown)


onevcat:
Thanks for it.

Are you using a custom processors from another package or so? May I know more about the `imageProcessorOptions` you are using?

> If there is a simple way to run them manually, then it would be great to have some details in contributing.md

The tests should be just available as soon as you open the project as either an Xcode project or a Swift Package. This sounds strange but I will check it! Also submit it to GitHub action to run.

ConfusedVorlon:
These are the processing options I'm using. All 'vanilla' stuff I think.

```
    static let imageProcessorOptions:KingfisherOptionsInfo = {
        let processor = KFImageAspectProcessor() |>
            DownsamplingImageProcessor(size: CGSize.init(side: 452)) |>
            RoundCornerImageProcessor(cornerRadius: 10)
        
        //use .png serializer to keep alpha in rounded corners
        //https://onevcat.github.io/Kingfisher/Structs/FormatIndicatedCacheSerializer.html#/s:10Kingfisher30FormatIndicatedCacheSerializerV3pngACvpZ
        let options:KingfisherOptionsInfo = [
            .cacheSerializer(FormatIndicatedCacheSerializer.png),
            .processor(processor),
        ]
        
        return options
    }()
```

ConfusedVorlon:
I don't think processing options are the issue here though. The problem is that when you call kf.setImage, that results in an immediate call to:

```
//LocalFileImageDataProvider
    public func data(handler: (Result<Data, Error>) -> Void) {
        handler(Result(catching: { try Data(contentsOf: fileURL) }))
    }
```

_on the main thread_. 

Simply calling `Data(contentsOf: fileURL)` for a 40MB image on my mac takes up to 0.1 seconds.
This is happening for (say) 20 images at once - so it completely destroys UI performance

onevcat:
Then I guess it would be a better place to fix it in `LocalFileImageDataProvider`. The process queue is originally for image processing instead of loading. What we need is an async disk data loading here. I will see what I can do.

ConfusedVorlon:
update - thinking again on the processor, I have updated it to the following:

```
    static let imageProcessorOptions:KingfisherOptionsInfo = {
        let maxSide = 452
        let processor = DownsamplingImageProcessor(size: CGSize.init(side: maxSide * 3)) |>
            KFImageAspectProcessor() |>
            DownsamplingImageProcessor(size: CGSize.init(side: maxSide)) |>
            RoundCornerImageProcessor(cornerRadius: 10)
        
        //use .png serializer to keep alpha in rounded corners
        //https://onevcat.github.io/Kingfisher/Structs/FormatIndicatedCacheSerializer.html#/s:10Kingfisher30FormatIndicatedCacheSerializerV3pngACvpZ
        let options:KingfisherOptionsInfo = [
            .cacheSerializer(FormatIndicatedCacheSerializer.png),
            .processor(processor),
        ]
        
        return options
    }()
```

which has given a massive boost on the processing time (approx 10x for the large slow images)
It should be fine as long as I don't get any excessively long/thin images.

it would be nice if there was a way to turn on logging in the processor so that it was easier to benchmark this kind of change without having to fork/update the code. Perhaps a processor option???

I'm just wrapping the processor call in provideImage

```
                        let start = DispatchTime.now()
                        guard let image = processor.process(item: processingItem, options: options) else {
                            options.callbackQueue.execute {
                                let error = KingfisherError.processorError(
                                    reason: .processingFailed(processor: processor, item: processingItem))
                                completionHandler(.failure(error))
                            }
                            return
                        }
                        let end = DispatchTime.now()
                        
                        let nanoTime = end.uptimeNanoseconds - start.uptimeNanoseconds // <<<<< Difference in nano seconds (UInt64)
                        let timeInterval = Double(nanoTime) / 1_000_000_000 // Technically could overflow for long running tests

                        print("Time to process:\(timeInterval)")
```

ConfusedVorlon:
> Then I guess it would be a better place to fix it in `LocalFileImageDataProvider`. The process queue is originally for image processing instead of loading. What we need is an async disk data loading here. I will see what I can do.

that makes sense - although I guess the same thing might apply in principle to other ImageDataProviders.

e.g. 

```
    public func data(handler: (Result<Data, Error>) -> Void) {
        let data = Data(base64Encoded: base64String)!
        handler(.success(data))
    }
```

could also be slow for a particularly large string.

anyway - many thanks for your swift responses. It's much appreciated.

ConfusedVorlon:
This is a much more minimal solution which seems to do the job.

onevcat:
> it would be nice if there was a way to turn on logging in the processor so that it was easier to benchmark this kind of change without having to fork/update the code. Perhaps a processor option???

The benchmark support is on my backlog but I need to find some time for it.

> although I guess the same thing might apply in principle to other ImageDataProviders.

Yes for some of them, but the in-memory converting from `String` to `Data` should be pretty fast. But again, it requires more profiling. Thanks for pointing it out!

onevcat:
I've fixed the tests in #1764 and let's track this there instead!

@ConfusedVorlon Thank you again for this!

PR #1775, opened 2021-08-13
minimal version target iOS 12.0

gre4ixin:


onevcat:
They are using Xcode 12.3 or earlier. It seems that iOS 12.0 was hidden from Xcode 12.4 according to the discussion here.

https://developer.apple.com/forums/thread/672668

After some reconsideration, I guess indeed we should make these values in project to be 12.0 as well. I will modify it in the project setting and send a new PR to change them in correct place. Keep an eye on the new v7.0 beta releases.

Best regards.

gre4ixin:
> They are using Xcode 12.3 or earlier. It seems that iOS 12.0 was hidden from Xcode 12.4 according to the discussion here.
> 
> 
> 
> https://developer.apple.com/forums/thread/672668
> 
> 
> 
> After some reconsideration, I guess indeed we should make these values in project to be 12.0 as well. I will modify it in the project setting and send a new PR to change them in correct place. Keep an eye on the new v7.0 beta releases.
> 
> 
> 
> Best regards.

Thank you 😊

PR #1787, opened 2021-08-27
Expose information about animation duration on AnimatedImageView

AttilaTheFun:
Hi @onevcat! I've been using Kingfisher for years and I love it.

I'm working on a feature for my app that adds a progress bar below gifs and videos.
AVPlayerItem has the information I need for videos, but for gifs I need to know the duration of the animation and the current frame + duration of each frame.

This information was already available internally, so this PR just exposes it in a readonly format.
There are no functional changes. Let me know what you think!


AttilaTheFun:
After some experimentation I'm going to try a different approach - closing this for now.

PR #1789, opened 2021-08-28
Expose AnimatedImageView fields to allow consumers to observe GIF progress

AttilaTheFun:
I'm working on a feature for my app that shows a progress bar below gifs and videos.
In order to animate the bar I need to know a couple more things about the animated image view and gif that aren't currently exposed.

The only functional change here is calling didPlayAnimationLoops *after* the loop completes (unless it plays to the end and stops). I believe this is what consumers would expect (at least, it's what I expected). This is necessary because I need to restart the progress bar animation when the gif loops back to the beginning, and currently this delegate method is being called while the gif is still showing the last frame.

### DIFF
--- a/Sources/Image/GIFAnimatedImage.swift
+++ b/Sources/Image/GIFAnimatedImage.swift
@@ -70,7 +70,7 @@ public struct ImageCreatingOptions {
 
 // Represents the decoding for a GIF image. This class extracts frames from an `imageSource`, then
 // hold the images for later use.
-class GIFAnimatedImage {
+public class GIFAnimatedImage {
     let images: [KFCrossPlatformImage]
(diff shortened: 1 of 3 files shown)


onevcat:
LGTM!

onevcat:
Yet we also need expose/rewrite some documentation in comments for them. But I guess I can handle it later.

@AttilaTheFun Thanks for it!

AttilaTheFun:
Awesome, thanks @onevcat!
I can point my app at the Kingfisher master branch for now, but how does your release process work?
Will these changes eventually make it into the 7.x release branch?

onevcat:
Yes. I will prepare a new beta release of 7.0 soon.

PR #1793, opened 2021-08-31
Added Writing options for diskStorage store operation

ignotusverum:
Straw-man proposal for providing more writing options for store operation as per discussion here https://github.com/onevcat/Kingfisher/discussions/1792.

**My question:**

I am wondering what would be the best way to pass any additional attributes to the `DiskStorage`? One of the requirements that we currently have for one of the projects - is to have `.complete`(FileProtectionType) protection as we need to ensure that any data we're storing on disk is encrypted at rest - at least when the device is locked.

**Proposed API:**

```swift
...
kf.setImage(with: url,
            placeholder: placeholder,
            options: [.diskStoreWriteOptions([.atomic, .completeFileProtection])])
...
```

Where new image resources for the `url` path will have `.atomic` & `.completeFileProtection` [writing options](https://developer.apple.com/documentation/foundation/nsdata/writingoptions).

### DIFF
--- a/Sources/Cache/DiskStorage.swift
+++ b/Sources/Cache/DiskStorage.swift
@@ -122,11 +122,13 @@ public enum DiskStorage {
         ///   - key: The key to which the `value` will be stored. If there is already a value under the key,
         ///          the old value will be overwritten by `value`.
         ///   - expiration: The expiration policy used by this store action.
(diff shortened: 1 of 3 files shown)


onevcat:
LGTM!

PR #1820, opened 2021-09-26
Add image download support to `CPListItem`

waynehartman:
This adds image download support for `CPListItem`.  `CPGridButton` does not get this same support because its image cannot be updated asynchronously without using private API.


onevcat:
This looks nice and promising. It extends the boundary of Kingfisher. Thank you @waynehartman !

I am not a CarPlay developer and have never tried it before. It seems that when building with iOS, the CarPlay related code is also compiled (with an error on CI).

I will try to read the CarPlay programming documentation and make sure I can know what is going on. After some brief look, it seems that I have to start from preparing a CarPlay simulator before I can try it. 

It is also welcome if you can give some suggestion or fix it, so we can continue this.

waynehartman:
> This looks nice and promising. It extends the boundary of Kingfisher. Thank you @waynehartman !
> 
> I am not a CarPlay developer and have never tried it before. It seems that when building with iOS, the CarPlay related code is also compiled (with an error on CI).
> 
> I will try to read the CarPlay programming documentation and make sure I can know what is going on. After some brief look, it seems that I have to start from preparing a CarPlay simulator before I can try it.
> 
> It is also welcome if you can give some suggestion or fix it, so we can continue this.

Taking a quick look at the build run, it looks like it has having a problem I do not experience with the latest iOS SDK. I’ll check it out later this afternoon on Xcode 12; my changes were done in Xcode 13.

onevcat:
Oh！

That is a good hint! I will also try it tomorrow (too late in my time zone now).

waynehartman:
> Oh！
> 
> That is a good hint! I will also try it tomorrow (too late in my time zone now).

I tried it on Xcode 12.5, but it built just fine and all tests passed. :|

waynehartman:
> Oh！
> 
> That is a good hint! I will also try it tomorrow (too late in my time zone now).

I fired up Xcode 12.4 (what you’re using for the build process) and it indeed breaks. Will investigate what the differences are between the two SDKs.

waynehartman:
> Oh！
> 
> That is a good hint! I will also try it tomorrow (too late in my time zone now).

Because the SDK made a change that is not backward compatible with SDK <14.5, I put in a `compiler(>=5.4)` check that was also introduced in Xcode 12.5 so that older users of the SDK can still use this capability, albeit in a slightly crippled way.

onevcat:
@waynehartman Thanks for it. It seems that I need a CarPlay Entitlements to build a demo app and test it even on the CarPlay simulator. It requires special apply process and seems to beyond what I can do now.

Since all is addition to current library so it should not break anything. I will merge and tag it without further validation as long as the CI passes.

Thank you again for it, this is an area I have never thought about Kingfisher can help.

Issue #1827, opened 2021-10-01
onFailedImage issue with SwiftUI on v 7


skellock:
Hello! Thank you kindly for this awesome library! In our app (SwiftUI and iOS 14+), we have different categories of images being cached and it's a breeze to do this stuff with Kingfisher.  😍 

We were rocking 5.x and this morning I upgraded us to 7.x. We were using `KingfisherSwiftUI` and I followed the migration guide(s) and in a few minutes, we were up & running. Great docs!

But I did run into a problem...

## TL;DR

Is `.onFailureImage` gone in version 7 on SwiftUI?

## Longer Explanation

It seems like `.onFailureImage` is no longer a thing for SwiftUI? Digging a little deeper, I see that the internals (and uikit) still support it, so I'm wondering if this was an oversight?

I flailed around a bit trying to write my own extension to remedy this issue, but it doesn't seem to have any effect.

```swift

public extension KFOptionSetter {
    func onFailureImage(_ image: KFCrossPlatformImage? = nil) -> Self {
        options.onFailureImage = .some(image)
        return self
    }
}

public extension KFImageProtocol {
    func onFailureImage(_ image: KFCrossPlatformImage? = nil) -> Self {
        options.onFailureImage = .some(image)
        return self
    }
}

```

A work around for now might be to handle my own fall-back error image using `.onSuccess` and `.onFailure`... and I'm glad to do that... but I just wanted to check to see if indeed `.onFailureImage` is gone in SwiftUI. 

I'm probably overlooking something here, eh? That's usually the case when I file issues 😆 

Thanks again!

onevcat:
I didn't aware of `. onFailureImage` was working in Kingfisher 5...It is a regression and I will see if I can add it back soon.

onevcat:
This modifier is added back per #1829. @skellock Can you try to see if it works for you?

I will find a time to merge and tag it if it does not bring any other side effect.

skellock:
Hey, thanks for taking a look!

I've updated my SPM to point to 01a875223111e25c2fcf9d6ed30fc5bb86b85a8e but unfortunately some recent commits around CarPlay is [preventing Xcode 13 from compiling](https://github.com/onevcat/Kingfisher/blob/1aba65263ebf20ef5849d25bcf92f48b68e17574/Sources/General/KF.swift#L230) rewarding me with this error:

```
[...]

PR #1834, opened 2021-10-04
Fixes incorrect `@available` for a flavour of `setImage`

skellock:
I ran into some issues compiling `master` related to some recent CarPlay merges due to this error:

```
Kingfisher/Sources/General/KF.swift:230:28: 'setImage(with:placeholder:parsedOptions:progressBlock:completionHandler:)' is only available in iOS 14.0 or newer
```

Context:

- Xcode 13
- iOS 14+ target app & app extension
- M1 dev box

I noticed [you fixed this elsewhere](https://github.com/onevcat/Kingfisher/commit/2bec5cd9f046ceee21e7965b8367c43073905c27) recently and I think it applies here too.

Thank you for Kingfisher!

(distant relative of #1827)


onevcat:
Great thanks for it!

Issue #1725, opened 2021-06-12
Failed to build module 'Accelerate'; this SDK is not supported by the compiler


ankitpanchotiya99138:
Hello,
Can any one help me I have face issue on Accelerate. I user Latest xcode 13 beta.

Failed to build module 'Accelerate'; this SDK is not supported by the compiler (the SDK is built with 'Apple Swift version 5.5 (swiftlang-1300.0.17.216 clang-1300.0.18.1)', while this compiler is 'Apple Swift version 5.5 (swiftlang-1300.0.19.104 clang-1300.0.18.4)'). Please select a toolchain which matches the SDK.

![Screenshot 2021-06-12 at 8 53 54 AM](https://user-images.githubusercontent.com/48627319/121763693-b8556a00-cb5b-11eb-93d9-ab6df2054c89.png)

onevcat:
I believe it is a bug in the beta version of Xcode. Some less-used system frameworks were built with a different compiler and bundled in this version (there are Accelerate, Network and more).

I suggest waiting for the next beta two weeks later.

ankitpanchotiya99138:
Ok Thanks for the quick response.

aleemtoor:
Duplicated with #1736

grantespo:
so just gotta wait for Xcode to fix kk, any day they should have update...

vd-ba:
Looks like there is similar issue with Xcode 13 Beta 2, albeit with different swift compiler version:

`Local setup: ❌  /Users/********/Library/Developer/Xcode/DerivedData/MyApp-drgvolpecatijzcuichjljjshvoi/SourcePackages/checkouts/Kingfisher/Sources/Image/ImageDrawing.swift:27:8: failed to build module 'Accelerate'; this SDK is not supported by the compiler (the SDK is built with 'Apple Swift version 5.5 (swiftlang-1300.0.20.4 clang-1300.0.21.1)', while this compiler is 'Apple Swift version 5.5 (swiftlang-1300.0.20.104 clang-1300.0.21.1)'). Please select a toolchain which matches the SDK.`

onevcat:
This seems only happens for SPM and it requires a minimum target iOS 12 to build successfully. I don't think this is reasonable and it is definitely an issue of the build system (unless Apple officially confirms they want to drop any deploy target below iOS 12)


[...]

PR #1842, opened 2021-10-08
Fix Xcode 13 build issue on Kingfisher 6 by increasing minimum iOS supported version from 10 to 11

4brunu:
Hey @onevcat,

This PR increases the minimum iOS supported version from 10 to 11, to fix Xcode 13 on Kingfisher 6 and fixes https://github.com/onevcat/Kingfisher/issues/1725.

Whats do you think about releasing a new version, 6.4.0 with this fix?

This is important because in Kingfisher 7, the SwiftUI code only supports iOS 14 and above, and in my case I need to support iOS 13, so I can't upgrade to Kingfisher 7 yet, and currently Kingfisher 6 doesn't build on Xcode 13.

This would make it easier for those who need to support iOS 13 and use SwiftUI.

Thanks

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,9 @@
 # Change Log
 
-## [6.4.0 - Published Wrapper](https://github.com/onevcat/Kingfisher/releases/tag/6.4.0) (2021-08-03)
+## [6.4.0 - Xcode 13 hotfix](https://github.com/onevcat/Kingfisher/releases/tag/6.4.0) (2021-08-03)
 
 #### Fix
-* Increate iOS minimum supported version from 10 to 12, to fix Xcode 13 build issue. [#1725](https://github.com/onevcat/Kingfisher/issues/1725)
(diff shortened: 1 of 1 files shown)


onevcat:
Thanks for it.

> Whats do you think about releasing a new version, 6.4.0 with this fix?

Upgrading the deploy target would be definitely a breaking major upgrade based on the semantic version rule. So it won't be a 6.4.0 or any version 6 release for this. However, we can make and keep it a special branch for a while before the majority migrates to SwiftUI on iOS 14 (maybe in another one or two years).

> This would make it easier for those who need to support iOS 13 and use SwiftUI.

I will change the deploy target based on the current `version6` branch and opt out the unnecessary changes. Then you can point your dependency to that branch until you can drop iOS 13 SwiftUI.

onevcat:
@4brunu 

The framework end users (app developers) will not have much troubles for it. However, things would be much different for other framework developers.

Although it is not obvious, to obey the semantic version is not only for framework end users, more important, is for other frameworks which are depending on Kingfisher. Breaking the version rule causes a huge impact to them. Even worse, it is really difficult for them to fix this in their frameworks without breaking the semantic versions too. It quickly becomes a disaster and I will be flooded by complaints. 😂 

So I am sorry but it won't be a choice to release a 6.4.0 for that.

4brunu:
Thanks 🙂

PR #1862, opened 2021-10-27
Updating KFOptionSetter with diskStoreWriteOptions

ignotusverum:
A follow up for https://github.com/onevcat/Kingfisher/pull/1793

After adding `diskStoreWriteOptions` to `KFOptionsSetter` we now able to add writing options to `KFImage` (SwiftUI) component with a following code:

``` swift
KFImage(ImageLoader.sampleImageURLs.first)
    .diskStoreWriteOptions([.completeFileProtection])
```

As well as updated tests that I missed during my previous PR. Please let me know if there are any changes you would like me to make @onevcat 👍

### DIFF
--- a/Sources/General/KFOptionsSetter.swift
+++ b/Sources/General/KFOptionsSetter.swift
@@ -237,6 +237,18 @@ extension KFOptionSetter {
         return self
     }
 
+    /// Sets writing options for an original image on a first write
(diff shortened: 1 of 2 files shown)


onevcat:
Thanks for it. This is perfect!

PR #1879, opened 2021-12-03
Code clean up by fixing swift lint warnings

kishorepran:
Fix swift lint warnings. Most of warnings have been fixed by the Swift Lint Auto correct tool.

### DIFF
--- a/Tests/KingfisherTests/ImageViewExtensionTests.swift
+++ b/Tests/KingfisherTests/ImageViewExtensionTests.swift
@@ -24,33 +24,32 @@
 //  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 //  THE SOFTWARE.
 
-import XCTest
 @testable import Kingfisher
+import XCTest
 
 class ImageViewExtensionTests: XCTestCase {
+    private var imageView: KFCrossPlatformImageView!
 
-    var imageView: KFCrossPlatformImageView!
-    
     override class func setUp() {
         super.setUp()
         LSNocilla.sharedInstance().start()
     }
-    
+
     override class func tearDown() {
         LSNocilla.sharedInstance().stop()
         super.tearDown()
     }
-    
+
     override func setUp() {
         super.setUp()
 
         imageView = KFCrossPlatformImageView()
         KingfisherManager.shared.downloader = ImageDownloader(name: "testDownloader")
         KingfisherManager.shared.defaultOptions = [.waitForCache]
-        
+
         cleanDefaultCache()
     }
-    
+
     override func tearDown() {
         LSNocilla.sharedInstance().clearStubs()
         imageView = nil
@@ -65,14 +64,13 @@ class ImageViewExtensionTests: XCTestCase {
         stub(url, data: testImageData, length: 123)
 
         var progressBlockIsCalled = false
-        
+
         imageView.kf.setImage(
             with: url,
             progressBlock: { _, _ in
                 progressBlockIsCalled = true
                 XCTAssertTrue(Thread.isMainThread)
-            })
-        {
+            }) {
             result in
             XCTAssertTrue(progressBlockIsCalled)
             XCTAssertNotNil(result.value)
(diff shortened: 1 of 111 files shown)


onevcat:
@kishorepran 

Hi,

Thanks for it. However, currently there is no plan now to introduce a linter tool like this into this project yet. Basically for now I am the only code contributor and a linter won't help much but only make things more complex. If we can get more people involved, it would definitely worth to add a linter or other rules, but maybe not yet for now.

Thank you again for it. These are nice, but just not the time. :)

Tests/KingfisherTests/ImageViewExtensionTests.swift
```diff
@@ -65,31 +64,30 @@ class ImageViewExtensionTests: XCTestCase {
...
             result in
             XCTAssertTrue(progressBlockIsCalled)
             XCTAssertNotNil(result.value)
 
             let value = result.value!
             XCTAssertTrue(value.image.renderEqual(to: testImage))
             XCTAssertTrue(self.imageView.image!.renderEqual(to: testImage))
-            //XCTAssertEqual(self.imageView.kf.taskIdentifier, Source.Identifier.current)
-            
+            // XCTAssertEqual(self.imageView.kf.taskIdentifier, Source.Identifier.current)
```
thejohnlima, line 81: If this line is not required I think it's better removing it

Tests/KingfisherTests/KingfisherManagerTests.swift
```diff
@@ -85,10 +85,14 @@ class KingfisherManagerTests: XCTestCase {
                 XCTAssertNotNil(result.value?.image)
                 XCTAssertEqual(result.value!.cacheType, .none)
                 exp.fulfill()
-        }}}}}
+            }
+        }
+        }
+        }
+        }
```
thejohnlima, line 92: What about using Xcode auto indentation?
This way we can have the function `testRetrieveImage`  more organized

```swift
func testRetrieveImage() {
    let exp = expectation(description: #function)
    let url = testURLs[0]

    stub(url, data: testImageData)

    let manager = self.manager!

    manager.retrieveImage(with: url) { result in
        XCTAssertNotNil(result.value?.image)
        XCTAssertEqual(result.value!.cacheType, .none)

        manager.retrieveImage(with: url) { result in
            XCTAssertNotNil(result.value?.image)
            XCTAssertEqual(result.value!.cacheType, .memory)

            manager.cache.clearMemoryCache()

            manager.retrieveImage(with: url) { result in
                XCTAssertNotNil(result.value?.image)
                XCTAssertEqual(result.value!.cacheType, .disk)

                manager.cache.clearMemoryCache()

                manager.cache.clearDiskCache {
                    manager.retrieveImage(with: url) { result in
                        XCTAssertNotNil(result.value?.image)
                        XCTAssertEqual(result.value!.cacheType, .none)
                        exp.fulfill()
                    }
                }
            }
        }
    }

    waitForExpectations(timeout: 3, handler: nil)
}
```

Tests/KingfisherTests/KingfisherManagerTests.swift
```diff
@@ -108,36 +112,40 @@ class KingfisherManagerTests: XCTestCase {
...
                 XCTAssertEqual(result.value!.cacheType, .none)
 
                 exp.fulfill()
-        }}}}}}
+            }
+        }
+        }
+        }
+        }
+        }
```
thejohnlima, line 133: Same as my previous [note](https://github.com/onevcat/Kingfisher/pull/1879/files#r762661732)

Tests/KingfisherTests/NSButtonExtensionTests.swift
```diff
@@ -66,20 +66,20 @@ class NSButtonExtensionTests: XCTestCase {
...
             result in
             XCTAssertTrue(progressBlockIsCalled)
-            
+
             let image = result.value?.image
             XCTAssertNotNil(image)
             XCTAssertTrue(image!.renderEqual(to: testImage))
             XCTAssertTrue(self.button.image!.renderEqual(to: testImage))
-            //XCTAssertEqual(self.button.kf.taskIdentifier, Source.Identifier.current)
+            // XCTAssertEqual(self.button.kf.taskIdentifier, Source.Identifier.current)
```
thejohnlima, line 80: It seems this line is a dead code, can we remove it?

Tests/KingfisherTests/NSButtonExtensionTests.swift
```diff
@@ -94,16 +94,15 @@ class NSButtonExtensionTests: XCTestCase {
...
             result in
             XCTAssertTrue(progressBlockIsCalled)
-            
+
             let image = result.value?.image
             XCTAssertNotNil(image)
             XCTAssertTrue(image!.renderEqual(to: testImage))
             XCTAssertTrue(self.button.alternateImage!.renderEqual(to: testImage))
-            //XCTAssertEqual(self.button.kf.alternateTaskIdentifier, Source.Identifier.current)
+            // XCTAssertEqual(self.button.kf.alternateTaskIdentifier, Source.Identifier.current)
```
thejohnlima, line 102: It seems this line is a dead code, can we remove it?

onevcat:
@thejohnlima Yes. These commented code should be removed. Ref: #1883

For the indentation, it is a multiple level nested things and it would be nice if we could rewrite it when async support added (in Xcode 13.2). For now, it is intended to break the format a bit to make it a bit more compact visually and describes a sequence of actions.

Tests/KingfisherTests/KingfisherManagerTests.swift
kishorepran: Agree, Xcode indentation would be better.

Issue #1598, opened 2021-01-13
UIImageView Flashing


liangdahong:
### Check List

Thanks for considering to open an issue. Before you submit your issue, please confirm these boxes are checked.

- [x] I have read the [wiki page](https://github.com/onevcat/Kingfisher/wiki) and [cheat sheet](https://github.com/onevcat/Kingfisher/wiki/Cheat-Sheet), but there is no information I need.
- [x] I have searched in [existing issues](https://github.com/onevcat/Kingfisher/issues?utf8=✓&q=is%3Aissue), but did not find a same one.
- [x] I want to report a problem instead of asking a question. It'd better to use [kingfisher tag in Stack Overflow](http://stackoverflow.com/questions/tagged/kingfisher) to ask a question.

### Issue Description

UIImageView Flashing.

#### Reproduce

First exit the background, then return to the foreground, refresh the TableView, UIImageView  flickerin.

#### Other Comment

Because Kingfisher's memory cache uses NSCache, Apple will clear the cache when the application exits to the background, which may cause flickering when the cell is refreshed. SDWebImage has made a layer of MapTable to handle this problem. I wonder if Kingfisher has a similar plan, sorry for the poor English, thank you.

timothyerwin:
It does seem to be constantly flashing the KFImage since I updated to 6.0.1...is there some workaround?

onevcat:
@liangdahong 
You can use a `.loadDiskFileSynchronously` for that. Check this: https://github.com/onevcat/Kingfisher/wiki/Cheat-Sheet#loading-disk-file-synchronously

@timothyerwin For `KFImage`, the `. loadDiskFileSynchronously` is default on due to its loading model. It is a bit surprising and maybe it's a regression of version 6. Can you share more information on this, or even better, some code snippet that we can reproduce it? Thanks.

liangdahong:
@onevcat 👌 , Thanks.

arakitatsuzou:
@onevcat 
Same flickering just like #1549 in 6.0.1.
Problem solved after I downgraded to 5.15.8.
In my case, any SwiftUI re-rendering can reproduce it.

onevcat:
OK..For the SwiftUI part, I guess it is a side effect of this #1595 

If the flickering when reloading is a problem for you, you can try to keep using the "deprecated" version like this, by passing the options instead:

```swift
KFImage(url, options: [])
```

It is not an ideal workaround but I guess it should work for now.

[...]

Issue #793, opened 2017-10-19
前后台切换，进入页面刷新图片（url不变）图片会先空白再出现，给人一种闪的感觉


OstrichRunning:
### Check List

Thanks for considering to open an issue. Before you submit your issue, please confirm these boxes are checked.

- [ ] I have read the [wiki page](https://github.com/onevcat/Kingfisher/wiki) and [cheat sheet](https://github.com/onevcat/Kingfisher/wiki/Cheat-Sheet), but there is no information I need.
- [ ] I have searched in [existing issues](https://github.com/onevcat/Kingfisher/issues?utf8=✓&q=is%3Aissue), but did not find a same one.
- [ ] I want to reoprt a problem instead of asking a question. It'd better to use [kingfisher tag in Stack Overflow](http://stackoverflow.com/questions/tagged/kingfisher) to ask a question.

#### What

[Tell us about the issue]

#### Reproduce

[The steps to reproduce this issue. What is the url you were trying to load, where did you put your code, etc.]

#### Other Comment

[Add anything else here]

onevcat:
Kingfisher is using `NSCache` for caching images in memory, which will be purged by system when your app being switched to background. It means Kingfisher needs to load it again from disk, which is now an async operation happens in another thread (for not blocking main thread when loading huge files). So basically it is what Kingfisher and iOS works. It is expected and design choice instead of an issue.

Since there is no plan to change this behavior now, I am closing it.

OstrichRunning:
ok.

lehom:
这个问题真的没有修改的计划吗？退后台进前台如果reload tableView的时候会造成不好的体验，如果用loadDiskFileSynchronously又会卡主线程 😭

onevcat:
A `keepWhenEnteringBackground` option was added to the memory cache and will be released soon.

PR #1882, opened 2021-12-05
fix: The problem of picture flickering after switching to the background (#793, #1598)

liangdahong:
An attempt was made to add a WeakCache solution for Kingfisher to switch to background image flicker. For specific reasons, please refer to: https://github.com/onevcat/Kingfisher/issues/1598#issue-784729434, #793, Referenc:  https://github.com/SDWebImage/SDWebImage/blob/8d558cb79f44c00bc75e71a509021b9551941aed/SDWebImage/Core/SDMemoryCache.m#L24 .

### DIFF
--- a/Sources/Cache/MemoryStorage.swift
+++ b/Sources/Cache/MemoryStorage.swift
@@ -26,6 +26,10 @@
 
 import Foundation
 
+#if canImport(UIKit)
+import UIKit
+#endif
+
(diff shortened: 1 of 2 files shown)


onevcat:
I am not a fan of adding another hash map to the memory storage backend. To prevent the background evict, maybe we have other options, such as implementing `NSDiscardableContent` to the storage object. I will see...

onevcat:
Implemented in a better way. See #1890

PR #1886, opened 2021-12-10
Attempt to fix crash on changing gif

IgorFedorchuk:


IgorFedorchuk:
reset animated frames on deinit
image

onevcat:
@IgorFedorchuk I am not quite understand it...Although this change seems harmless, may I ask why can this fix the crash?

onevcat:
I guess it would be good to add it since the deinit of safe array cannot be locked. :(

PR #1919, opened 2022-03-11
Update README.md

M1ruben:


onevcat:
@M1ruben Thanks for the PR 😂 I can feel your love of Kingfisher but unfortunately this does not quite improve the framework XD... So I will leave it as it is now.

PR #1927, opened 2022-03-28
fix: remove redundant word 'disk'

SunsetWan:


onevcat:
Thank you for catching it!

PR #1951, opened 2022-06-08
Update AVAssetImageDataProvider.swift

sgarg4008:


VikkyTheDev (approved):

sgarg4008:
I used the kingfisher for generating the image asset but after generating the image asset, I got the rotated image and then I just added the below line in your file and it worked perfectly fine.

"generator.appliesPreferredTrackTransform = true"

This is the code I used to get the image from the video URL:- 

let provider = AVAssetImageDataProvider(assetURL: URL(string: videoUrl)!, seconds: 1)
self.imageView.kf.setImage(with: provider)


Please review the changes and approve the same.

onevcat:
I guess it is not a bad default!

sgarg4008:
> I guess it is not a bad default!

Hey @onevcat, 
Thanks for taking my request under consideration.

I tried the code using the default functionality but it gives me the rotated image as a result, once we set this property True, only then we get the correct result.

**- generator.appliesPreferredTrackTransform = true**

You can implement it at your end, just to review the same.


I'm attaching a reference link where many of the users got the same error, without setting this property to True.

Related question reference Link:- 
- https://stackoverflow.com/questions/5347800/avassetimagegenerator-provides-images-rotated


Related answer reference Link:-
- https://stackoverflow.com/a/5347872/14180967

onevcat:
Oh, I meant setting it to true would be a good idea as “new default”! Thanks for the PR.

sgarg4008:
Thanks for merging the code :)

sgarg4008:
Hey @onevcat, 

I just tried updating the pod and even I tried to add your library using Swift package manager, but I couldn't find the updated changes.

Do I need to update something else?

onevcat:
Merging it to the master branch does not mean a new release. Unless you are explicitly pointing to the `master` branch, the package manager will still resolve to an existing version.

I will tag these changes in a few minutes and then you can try again to upgrade to a new release.

sgarg4008:
Thanks for the new release.

Issue #1844, opened 2021-10-11
AnimatedImageView crash


MalcolmnDEV:
### Check List

Thanks for considering to open an issue. Before you submit your issue, please confirm these boxes are checked.

- [x] I have read the [wiki page](https://github.com/onevcat/Kingfisher/wiki) and [cheat sheet](https://github.com/onevcat/Kingfisher/wiki/Cheat-Sheet), but there is no information I need.
- [x] I have searched in [existing issues](https://github.com/onevcat/Kingfisher/issues?utf8=✓&q=is%3Aissue), but did not find a same one.
- [x] I want to report a problem instead of asking a question. It'd better to use [kingfisher tag in Stack Overflow](http://stackoverflow.com/questions/tagged/kingfisher) to ask a question.

#### What

Since the update to Kingfisher 7.0.0 users have been reporting several crashes inside our app. Our crash reporting software captured the stack traces and it seems to be inside the Kingfisher iOS SDK.

- AnimatedImageView.Animator.loadFrame(at:)
Stack Trace:

Crashed: com.onevcat.Kingfisher.Animator.preloadQueue
0  libsystem_platform.dylib       0x1a44 _platform_memmove + 548
1  CoreFoundation                 0xb32f8 CFDataGetBytes + 328
2  ImageIO                        0x1500 CGImageGetImageSource + 152
3  UIKitCore                      0x203024 -[_UIImageCGImageContent initWithCGImage:scale:] + 36
4  UIKitCore                      0x18c3ec -[UIImage initWithCGImage:scale:orientation:] + 72
5  Woebot                         0x266520 AnimatedImageView.Animator.loadFrame(at:) + 4336739616 (<compiler-generated>:4336739616)
6  Woebot                         0x2667b8 AnimatedImageView.Animator.updatePreloadedFrames() + 575 (AnimatedImageView.swift:575)
7  Woebot                         0x1d51f0 thunk for @escaping @callee_guaranteed () -> () + 4336144880 (<compiler-generated>:4336144880)
8  libdispatch.dylib              0x632ec _dispatch_call_block_and_release + 24
9  libdispatch.dylib              0x642f0 _dispatch_client_callout + 16
10 libdispatch.dylib              0x3f410 _dispatch_lane_serial_drain$VARIANT$armv81 + 596
11 libdispatch.dylib              0x3feb8 _dispatch_lane_invoke$VARIANT$armv81 + 388
[...]

PR #1956, opened 2022-06-18
Fix a crash on AnimatedImageView

ufosky:
Fix a crash report in this thread https://github.com/onevcat/Kingfisher/issues/1844
Add a workaround to ensure the image instance dealloc in main thread.
image


onevcat:
Umm, nice catch. Thanks for it.

PR #1958, opened 2022-06-23
Remove unused code

idrougge:
This should make Kingfisher slightly smaller and slightly faster to compile.


onevcat:
Hi, @idrougge 

Nice catch on these! But may I request to revert the changes on `CIImageProcessor`? It is a public interface and removing it would just break the back-compatibility of the framework. We cannot do that without a major upgrade and apparently it is not the time now.

onevcat:
Thanks!

PR #1965, opened 2022-07-11
Fix AnimatedView Crash

douknow:
Fix a crash in this thread [#1844](https://github.com/onevcat/Kingfisher/issues/1844), replace another method create image by index.
image


onevcat:
Umm. That is an interesting fix.

@douknow  May I know how did you find it? Is there any clue or discussions about it?

douknow:
> Umm. That is an interesting fix.
> 
> @douknow Thanks for the PR. May I know how did you find it? Is there any clue or discussions about it?

emmm, Just common crash in my dev life, and I try switch to another method for create image, then no more crash in here.

PR #1997, opened 2022-10-05
ImageCache.diskStorageSize

RomanPodymov:
Hello.
Thank you for Kingfisher.
Continue with Concurrency.

### DIFF
--- a/Sources/Cache/ImageCache.swift
+++ b/Sources/Cache/ImageCache.swift
@@ -816,6 +816,28 @@ open class ImageCache {
         }
     }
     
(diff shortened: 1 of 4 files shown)


RomanPodymov:
@onevcat Hello, what should I fix in this pull request?

onevcat:
@RomanPodymov 

I just did not have enough time to take a look at it yet! Sorry for that.

Can you please just call `calculateDiskStorageSize` inside this async getter? I want to keep no duplication on this as possible, and in the future if we need to apply any change, we can make sure to just do it in one place.

Issue #1825, opened 2021-10-01
Wrong cache size from ImageCache.default.calculateDiskStorageSize, larger than 10x


iaomw:
### Check List

Thanks for considering to open an issue. Before you submit your issue, please confirm these boxes are checked.

- [x] I have read the [wiki page](https://github.com/onevcat/Kingfisher/wiki) and [cheat sheet](https://github.com/onevcat/Kingfisher/wiki/Cheat-Sheet), but there is no information I need.
- [x] I have searched in [existing issues](https://github.com/onevcat/Kingfisher/issues?utf8=✓&q=is%3Aissue), but did not find a same one.
- [x] I want to report a problem instead of asking a question. It'd better to use [kingfisher tag in Stack Overflow](http://stackoverflow.com/questions/tagged/kingfisher) to ask a question.

### Issue Description
I noticed my cache size become much larger after using Kingfisher 7.0.0 version. Today, I did manually calculation. iPadOS 15.0 told me my app is taking 1.36GB documents & data. My app instance actually has nearly 1GB documents. But kingfisher is reporting nearly 5GB (5348560366) cache size on disk. That seems wrong.

#### Reproduce
The cache size looks also wrong on iOS 14.8, but it's not the same ratio. 

#### Other Comment
Not tested in iOS simulator

onevcat:
Not sure if the documents & data number reported by iOS contains the cache data or not. Kingfisher is putting the disk cache into the Cache folder so maybe it is not counted by iOS? Can you confirm that the size of your Cache folder of your app?

Kingfisher is iterating all files under its disk cache folder by [this API](https://developer.apple.com/documentation/foundation/url/1780058-resourcevalues), with the [`fileSizeKey` passed in](https://developer.apple.com/documentation/foundation/urlresourcekey/1417747-filesizekey). 

I guess there would quite little we can do if this API is returning a wrong value for a file. But I will try to confirm it first.

iaomw:
Thanks, I did a quick test with this API and Kingfisher source. This API seems not working for folder `com.onevcat.Kingfisher.ImageCache.default` . So I cannot compare the folder size with total size of all individual files.

While the folder is on iPad, it's difficult to get size directly. I will test with iOS simulator later.

I reported this issue, because Kingfisher 6.x works fine to get total size of cache. I didn't count it manually, but it was very close roughly. Any related code changed from 6.x to 7.0.

onevcat:
[...]

PR #2002, opened 2022-10-20
local cache missed when reinstall and start app during the development phase

elijahdou:
Problem:
With the issue https://github.com/onevcat/Kingfisher/issues/1825 you fixed some issues, but i I found that there is still a problem.

The original solution does not cover the situation of reinstall the app during the development phase. Every time i reintall and start the app from Xcode, the cache is missed due to dynamic sandbox path and there are tons of duplicate thumbnails in the sandbox.

The new `cacheKey` may as follow:
`kingfisher.local.cacheKey/~/Documents/dir/file.ext`

### DIFF
--- a/Sources/General/ImageSource/AVAssetImageDataProvider.swift
+++ b/Sources/General/ImageSource/AVAssetImageDataProvider.swift
@@ -53,7 +53,10 @@ public struct AVAssetImageDataProvider: ImageDataProvider {
     public let time: CMTime
 
     private var internalKey: String {
-        return (assetImageGenerator.asset as? AVURLAsset)?.url.absoluteString ?? UUID().uuidString
+        guard let url = (assetImageGenerator.asset as? AVURLAsset)?.url else {
+            return UUID().uuidString
+        }
+        return url.isFileURL ? url.localFileCacheKey : url.absoluteString
     }
 
     /// The cache key used by `self`.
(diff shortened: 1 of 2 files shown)


onevcat:
I am not quite catching up with you.

Currently, we should have already eliminated the dynamic part of the cache key, to something containing the app name and the image URL path, such as `kingfisher.local.cacheKey/Kingfisher-Demo.app/images/kingfisher-1.jpg?foo=bar`.

So 

> the cache is missed due to the dynamic sandbox path and there are tons of duplicate thumbnails in the sandbox

this should not happen. May I know what are the cache keys you are having with the current released Kingfisher versions?

elijahdou:
@onevcat Everything is OK except on situation that i launch iphone device from xcode to debug and i find that cache is missed and regenerate thumbnails (My project needs to generate thumbnails for local images).
My devices: Xcode 14.0.1, IphoneX with OS version 15.7
Based on my test, this PR can cover this case .

onevcat:
The existing `localFileCacheKey` computed property seems working fine, and the only issue is we forgot to also apply the changes in #1825 to `AVAssetImageDataProvider`.

It should be fixed in #2158 and I can verify that a local video (or say, an `AVAsset`) should also respect the local file cache key and won't be cached for multiple times even during debugging with Xcode.

PR #2031, opened 2023-02-16
modify: Modified to apply the default placehoder even when a task fails

DevVenusK:
In the case of placeholder, I think it should be exposed by default when it fails. So, even when failureImage is not set, I think it is good to expose the set placeholder when the task fails.

### DIFF
--- a/Sources/Extensions/CPListItem+Kingfisher.swift
+++ b/Sources/Extensions/CPListItem+Kingfisher.swift
@@ -203,7 +203,7 @@ extension KingfisherWrapper where Base: CPListItem {
                                 
                             } else {
                                 #if compiler(>=5.4)
-                                self.base.setImage(nil)
+                                self.base.setImage(placeholder)
                                 #endif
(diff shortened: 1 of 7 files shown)


onevcat:
Hi,

Thanks for this. But I guess maybe it is only necessary to apply this fix to the `CPListItem` case. For other cases like views and buttons, the placeholder will be kept there untouched when the loading fails and no failure image option is passed in.

Can you narrow down the fix only to the `CPListItem`, where indeed we can just remove the `self.base.setImage(nil)` to keep the placeholder showing.

DevVenusK:
thanks for the comment
Of course you can narrow it down to CPListItem .

The default value of keepCurrentImageWhileLoading is true , which I'm assuming.
I'll fix it. thank you !

DevVenusK:
@onevcat I put it back and only applied it to CPListItem!

onevcat:
Thanks!

PR #2048, opened 2023-03-16
Feature response delegate

onevcat:
A generic way to implement #2043 

This adds a delegate method below:

```swift
protocol ImageDownloaderDelegate {
    func imageDownloader(
        _ downloader: ImageDownloader,
        didReceive response: URLResponse,
        completionHandler: @escaping (URLSession.ResponseDisposition) -> Void)
}
```

It is not possible to setup a delegate for the `ImageDownloader` and determine a response disposition when a response is received. Some example usage:

```swift
class ViewController: UIViewController {
    override func viewDidLoad() {
        super.viewDidLoad()
        ImageDownloader.default.delegate = self
    }
}

extension ViewController: ImageDownloaderDelegate {
    func imageDownloader(_ downloader: ImageDownloader, didReceive response: URLResponse, completionHandler: @escaping (URLSession.ResponseDisposition) -> Void) {
        guard let httpResponse = response as? HTTPURLResponse else {
            completionHandler(.cancel)
            return
        }
        let value = httpResponse.value(forHTTPHeaderField: "content-type")

        // Only allow JPEG images, based on the response header.
        completionHandler(value == "image/jpeg" ? . allow : .cancel)
    }
}
```

Now, loading a PNG:

```swift
KF.url(somePNGImageURL)
    .onFailure { err in print("Error: \(err)") }
```

Will give you a `ResponseErrorReason.cancelledByDelegate` error.

### DIFF
--- a/Sources/General/KingfisherError.swift
+++ b/Sources/General/KingfisherError.swift
@@ -88,6 +88,10 @@ public enum KingfisherError: Error {
         /// The task is done but no URL response found. Code 2005.
         /// - task: The failed task.
         case noURLResponse(task: SessionDataTask)
+
+        /// The task is cancelled by `ImageDownloaderDelegate` due to the `.cancel` response disposition is
+        /// specified by the delegate method. Code 2006.
+        case cancelledByDelegate(response: URLResponse)
     }
     
     /// Represents the error reason during Kingfisher caching system.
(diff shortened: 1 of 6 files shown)


rztime:
```
guard let httpResponse = response as? HTTPURLResponse else {
    completionHandler(.cancel)
    return
}
let value = httpResponse.value(forHTTPHeaderField: "content-type")
completionHandler(value == "image/jpeg" ? . allow : .cancel)
```

这里建议直接使用

```
let value = response.mimeType
```

当url为本地file:///xxxx.jpeg的图片时，response 不是一个 HTTPURLResponse 会导致错误行为

onevcat:
这里只是一个可行示例，`ImageDownloaderDelegate` 中的相关方法可由使用方进行具体实现。

Issue #2062, opened 2023-04-10
Warning 'UIKit' was not imported by this file


zunda-pixel:
### Check List

Thanks for considering to open an issue. Before you submit your issue, please confirm these boxes are checked.

- [✓ ] I have read the [wiki page](https://github.com/onevcat/Kingfisher/wiki) and [cheat sheet](https://github.com/onevcat/Kingfisher/wiki/Cheat-Sheet), but there is no information I need.
- [✓ ] I have searched in [existing issues](https://github.com/onevcat/Kingfisher/issues?utf8=✓&q=is%3Aissue), but did not find a same one.
- [✓ ] I want to report a problem instead of asking a question. It'd better to use [kingfisher tag in Stack Overflow](http://stackoverflow.com/questions/tagged/kingfisher) to ask a question.

#### What

This warning is showed in XcodeCloud.

`Class property 'black' cannot be used in a default argument value because 'UIKit' was not imported by this file; this is an error in Swift 6`

- Xcode 14.3
- macOS Ventura 13.4 beta

<img width="500" alt="screenshot" src="https://user-images.githubusercontent.com/47569369/230885822-8cf02a8e-215e-4749-8585-3ec83fed615c.png">

#### Reproduce

I don't known how to reproduce in Local Xcode(Compile arguments)

onevcat:
Close it since #2063 merged. Thank you!

PR #2063, opened 2023-04-10
fix warning 'UIKit' was not imported

zunda-pixel:
fix warning `Class property 'black' cannot be used in a default argument value because 'UIKit' was not imported by this file; this is an error in Swift 6`

Issue #2062

screenshot


onevcat:
Nice catch!

PR #2074, opened 2023-05-03
modified Source.Identifier.current access modifier

iwill-hwang:
Source.Identifier.current is a unique value for data processing in UI Components.
this value can be generated using Source.Identifier.next(),
it can also be modified outside of its scope.

To prevent modification of current properties outside of its scope, it is safer to mark it as private(set).


onevcat:
Great. Thank you!

PR #2094, opened 2023-06-12
Introduce a custom image source provider to enable third-party image processors to utilize AnimatedImageView.

yeatse:
Before this update, third-party image processors were unable to create animated images without relying on `UIImage.animatedImage`, which is only available on iOS and can consume a significant amount of memory. 

This pull request addresses the problem by introducing a new protocol called `ImageFrameSource` and refactoring `AnimatedImageView` to utilize it.

Relevant issue:
https://github.com/yeatse/KingfisherWebP/issues/41
https://github.com/yeatse/KingfisherWebP/issues/39

### DIFF
--- a/Sources/Image/GIFAnimatedImage.swift
+++ b/Sources/Image/GIFAnimatedImage.swift
@@ -74,13 +74,13 @@ public class GIFAnimatedImage {
     let images: [KFCrossPlatformImage]
     let duration: TimeInterval
     
-    init?(from imageSource: CGImageSource, for info: [String: Any], options: ImageCreatingOptions) {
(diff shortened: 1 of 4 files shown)


onevcat:
At the first glance, it is coooool. 

Thanks for this. I will take a detail look at it ASAP.

yeatse:
It appears that the check failed because of a network connection issue. I tested it on my machine and all tests were passed successfully.

onevcat:
Don't worry about it! I guess it is something related to the low spec of GH's node and the shared resource. It sometimes fails at the first time.

PR #2099, opened 2023-06-18
fix: always use specified processors to recreate animated image representation

yeatse:
When loading an animated image from the cache, Kingfisher will recreate it using the cached data. To avoid loading failures due to unsupported format with the default processor, we must ensure that the specified processor is always used.


onevcat:
That makes sense! Thank you for catching it.

Issue #2116, opened 2023-07-28
Usage of MD5 flagged as vulnerability in mobile penetration testing


kmaschke85:
### Check List

Thanks for considering to open an issue. Before you submit your issue, please confirm these boxes are checked.

- [X] I have read the [wiki page](https://github.com/onevcat/Kingfisher/wiki) and [cheat sheet](https://github.com/onevcat/Kingfisher/wiki/Cheat-Sheet), but there is no information I need.
- [X] I have searched in [existing issues](https://github.com/onevcat/Kingfisher/issues?utf8=✓&q=is%3Aissue), but did not find a same one.
- [X] I want to report a problem instead of asking a question. It'd better to use [kingfisher tag in Stack Overflow](http://stackoverflow.com/questions/tagged/kingfisher) to ask a question.

#### What

Our app has to pass mobile penetration testing and during scanning our code base, Kingfisher was flagged as being vulnerable because of using MD5.

#### Reproduce

Integrate Kingfisher into an iOS project via Cocoapods and use MobSF to scan the code base.

#### Other Comment

Should be simple to replace MD5 with SHA256 provided by CryptoKit.
<img width="1424" alt="Bildschirmfoto 2023-07-28 um 12 12 50" src="https://github.com/onevcat/Kingfisher/assets/16348228/c5bedd21-962d-40ce-8596-2a0f736d68c1">

PR #2117, opened 2023-07-28
Replace MD5 with SHA256

kmaschke85:
As pointed out in #2116 , usage of MD5 is not recommended anymore and may be flagged as vulnerability in penetration testing.
This PR simply uses SHA256 instead of MD5 for generating the storage hashes, by utilizing existing functionality from CommonCrypto and CryptoKit.


onevcat:
This changes the cache key for all disk cache which can cause all cache invalidated. So please let me switch the base branch to `v8`, as planned along with some other upcoming breaking changes.

kmaschke85:
Sure, no problem!

kmaschke85:
Any plans when v8 could be released? @onevcat

PR #2138, opened 2023-08-30
Fix Kingfisher-Swift.h and module.modulemap files generation for Objective-C compatibility

yev-kanivets:
Similar to https://github.com/onevcat/Kingfisher/pull/1451, but in more detail:
1. Puts back `DEFINES_MODULE = YES;` to generate the `module.modulemap` file in the `Modules` folder.

Capture d’écran 2023-08-30 à 12 00 11

2. Adds `SWIFT_INSTALL_OBJC_HEADER = YES;` to generate `Kingfisher-Swift.h` file in the `Headers` folder.

Capture d’écran 2023-08-30 à 11 59 45

Objective-C import doesn't work without these two files. And it's necessary if `Kingfisher` is imported in a library module and not in the main app module, which is the case for most multi-modular projects (including mine).


yev-kanivets:
@onevcat hello, here is a small PR to fix the Objective-C compatibility of the Kingfisher library 🙏

onevcat:
@yev-kanivets Hi, thanks for opening this! Your change is nice and neat.

However, removing of ObjC support is intended and was done long ago to prevent some troubles. You can find the detail in the PR below and related links:

https://github.com/onevcat/Kingfisher/pull/1523

Not sure whether things are different now, but since Kingfisher contains quite a few Swift structs, enums and other syntax which are not available in ObjC, maybe it does not that worth the effect of supporting ObjC (and resolving all other potential issues during build).

Instead, if you need a similar library in Objective-C, I strongly recommend the [SDWebImage](https://github.com/SDWebImage/SDWebImage) as an alternate, which is written in pure Objective-C and providing more nature syntax than Kingfisher :]

yev-kanivets:
@onevcat thanks for the comprehensive answer 🙏 

We aren't using Objective-C in the project too (never used). Our problem comes from the fact that the interactions between XCode modules are still done through the Objective-C API, which is generated automatically. And `Kingfisher` is automatically included in this generated library  (our internal Swift module) header because we expose some `Kingfisher` classes in its interface.

I think that for us (and probably for many other teams) there is no way around this problem but enabling those two flags in all our dependencies. I've checked a few dozen of libraries we are using (Alamofire, Lottie, SwiftSoup, MBProgressHUD, SVGKit, etc.) and all of them have those two flags enabled, so they have `module.modulemap` and `*-Swift.h` files in the generated frameworks.

I have very limited knowledge of library development, but my guess is that those two flags aren't about supporting your library in Objective-C, but rather about having Objective-C API layer automatically generated, so your library can be used in other library modules.

It's pretty much what @daniele-pizziconi said in https://github.com/onevcat/Kingfisher/pull/1451#issue-619029476. WDYT?

onevcat:
I revised this and remembered another related change in #1532 

With that change, I guess yes, we should be able to bring the module and header export back.

I will do a few double check on the integration with other frameworks and merge this when it is done!

@yev-kanivets Thanks for it and terribly sorry for the delay (since I became crazy busy recently!)

yev-kanivets:
@onevcat no problem, thanks for coming back 🙏 We've resolved our problem with a private fork as a short-term solution :)

onevcat:
Contained in 7.10.0 and it was just released.

PR #2144, opened 2023-09-10
Basic concurrency support for available public methods

onevcat:


eschos24:
When is this going to be available in an upcoming release? Is there a reason it was not part of 7.10?

onevcat:
@eschos24 
The related changes (and also some others) contain breaking changes to the current version, so it is planned to be a part of the upcoming version 8.0.0.

PR #2171, opened 2023-11-27
Fix a critical performance issue in GIF playback

octree:
## Problem
https://github.com/onevcat/Kingfisher/blob/277f1ab2c6664b19b4a412e32b094b201e2d5757/Sources/Views/AnimatedImageView.swift#L627

This line of code results in a major inefficiency where the image of previous frame was released. This behavior was particularly problematic for looping GIF animations. Each loop involved repeatedly extracting the image of each frame, leading to substantial CPU resource consumption.

## Solution
The proposed update introduces the `isReachMaxRepeatCount` condition in the `updatePreloadedFrames()` method. This condition ensures that the Image of the previous Frame is only released when the animation reaches its maximum repeat count. This approach prevents the unnecessary extraction of Images in each loop for infinite animations, significantly reducing CPU usage and optimizing playback performance.

## Impact
* Reduced CPU Usage

### Before

Before

### After
After

### DIFF
--- a/Sources/Views/AnimatedImageView.swift
+++ b/Sources/Views/AnimatedImageView.swift
@@ -623,10 +623,10 @@ extension AnimatedImageView {
                 return
             }
 
-            let previousFrame = animatedFrames[previousFrameIndex]
-            animatedFrames[previousFrameIndex] = previousFrame?.placeholderFrame
-            // ensure the image dealloc in main thread
-            defer {
+            if isReachMaxRepeatCount {
+                let previousFrame = animatedFrames[previousFrameIndex]
+                animatedFrames[previousFrameIndex] = previousFrame?.placeholderFrame
+                // ensure the image dealloc in main thread
                 if let image = previousFrame?.image {
                     DispatchQueue.main.async {
                         _ = image


TastyHeadphones:
This pull request to the `Sources/Views/AnimatedImageView.swift` file adds a new condition to the `extension AnimatedImageView` to check if the `isReachMaxRepeatCount` property is true and execute the code inside the condition block.

Main change:

* <a href="diffhunk://#diff-c4f0f8dea3866536b4cc94b4d2dd74c89ae02a67474d1a957adfb9d0775b48d7R626-L629">`Sources/Views/AnimatedImageView.swift`</a>: Added a new condition to check if the `isReachMaxRepeatCount` property is true and execute the code inside the condition block.

octree:
@onevcat 

Thank you for your detailed feedback on the PR. I understand your concerns about the tradeoff between CPU usage and memory consumption. 

Given the potential problems with memory usage for large images and the need for a more holistic solution, I agree that it's prudent not to merge this PR at this time. I'll close the PR and keep an eye on future updates to the framework. 

Thanks again for considering the PR and for the valuable work you're doing on the project.

PR #2174, opened 2023-12-05
Add extension to has image property components

Mx-Iris:
I found that there are a lot of objects with image property, if I add setImage method for them one by one will lead to a lot of repetitive code, so I abstracted a protocol, in the future as long as the object with the image property to conform with this protocol can get setImage method.


onevcat:
@Mx-Iris 

Thank you for your patience and for submitting your pull request to the Kingfisher library. I apologize for the delayed response, as I only had the opportunity to review it thoroughly today.

I am impressed with the functionality introduced in your PR, and I see great value in integrating it into the Kingfisher library. However, we have recently made several updates to improve strict concurrency, including adding `@MainActor` annotations to image setting methods. Given these changes, I believe your contribution would fit perfectly in the upcoming major release.

Could you please rebase your changes onto the latest [`v8` branch](https://github.com/onevcat/Kingfisher/tree/v8) and submit a new PR? I am eager to review it and include it as a stable feature in our next release.

Thank you once again for your valuable contribution and effort.

Mx-Iris:
I have merged the latest code and added some UIKit components, I found that there are a lot of UIConfigurations that have the `image` property, but this implementation of mine only takes into account AnyObject, and if I need to be compatible with the value type, I need to change the `base` property of KingfisherWrapper from `let` to `var`, do you have a Do you have a better way to implement this feature?

onevcat:
Don't worry! Let me try to handle it.

Mx-Iris:
Ok, thanks for submitting a new PR for me and adding concurrency, I'm not familiar with swift concurrency so I didn't understand what you said above, I'll close this PR then, feel free to @ me if you have any questions!

PR #2185, opened 2023-12-22
style: spelling and grammar check

YoloMao:


onevcat:
@YoloMao 

Oh...Thanks for the hard work and effort for improving the docs. It looks nice, but I am afraid it is not a good time to merge this one now.

In fact, we already had a bunch of work around the documentation grammar and spelling improvement in #2160 and under the [`v8` branch](https://github.com/onevcat/Kingfisher/tree/v8), which is planned to be merged to master when some other additional work prepared, as the next major version of Kingfisher. 

Although I do not try yet, I believe merging this PR to the `master` will cause a lot of conflicting later when we want to merge `v8`. And it can hardly to determine which one should be picked up. So please allow me to close this PR without merging. (Sorry for that!)

And if you are still kindly enough and want to help improve the quality of documentation, I would suggest you can pick the `v8` branch and the already refined documentation as the base branch! I would be very grateful for that!

Thank you again!

PR #2192, opened 2024-01-08
Support `KFAnimatedImage` and `AnimatedImageView` on macOS

yeatse:
So that we can play animated GIFs on SwiftUI for macOS

### DIFF
--- a/Sources/SwiftUI/KFAnimatedImage.swift
+++ b/Sources/SwiftUI/KFAnimatedImage.swift
@@ -47,38 +47,17 @@ public struct KFAnimatedImage: KFImageProtocol {
     }
 }
 
-#if !os(macOS)
-/// A wrapped `UIViewRepresentable` of `AnimatedImageView`
-@available(iOS 14.0, tvOS 14.0, watchOS 7.0, *)
-public struct KFAnimatedImageViewRepresenter: UIViewRepresentable, KFImageHoldingView {
-    public typealias RenderingView = AnimatedImageView
-    public static func created(from image: KFCrossPlatformImage?, context: KFImage.Context<Self>) -> KFAnimatedImageViewRepresenter {
-        KFAnimatedImageViewRepresenter(image: image, context: context)
-    }
-    
-    var image: KFCrossPlatformImage?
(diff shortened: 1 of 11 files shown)


onevcat:
Perfect at the first glance! I will look into them shortly when having a chance. Thank you!

Sources/SwiftUI/KFAnimatedImage.swift
```diff
@@ -75,6 +76,35 @@ public struct KFAnimatedImageViewRepresenter: UIViewRepresentable, KFImageHoldin
         uiView.image = image
     }
 }
+#else
+@available(macOS 11.0, *)
+public struct KFAnimatedImageViewRepresenter: NSViewRepresentable, KFImageHoldingView {
```
onevcat, line 81: I see quite a lot of duplicated code.

Maybe we can combine these two `KFAnimatedImageViewRepresenter`s into one? Something like:

```swift
#if !os(macOS)
typealias KFCrossPlatformViewRepresentable = UIViewRepresentable
#else
typealias KFCrossPlatformViewRepresentable = NSViewRepresentable
#endif
```

What your opinion?

Sources/Views/AnimatedImageView.swift
```diff
@@ -162,6 +173,101 @@ open class AnimatedImageView: UIImageView {
     // A flag to avoid invalidating the displayLink on deinit if it was never created, because displayLink is so lazy.
     private var isDisplayLinkInitialized: Bool = false
     
+#if os(macOS)
+    class DisplayLink {
```
onevcat, line 177: Can we move this class to another new file to keep this file a bit cleaner?

Sources/SwiftUI/KFAnimatedImage.swift
yeatse: Good idea. Updated

Sources/Views/AnimatedImageView.swift
yeatse: Done

onevcat:
@yeatse Perfect! Thanks for the epic efforts!

Issue #1602, opened 2021-01-18
Consider making all of DiskStorage.Backend's methods public.


jyounus:
### Check List

Thanks for considering to open an issue. Before you submit your issue, please confirm these boxes are checked.

- [x] I have read the [wiki page](https://github.com/onevcat/Kingfisher/wiki) and [cheat sheet](https://github.com/onevcat/Kingfisher/wiki/Cheat-Sheet), but there is no information I need.
- [x] I have searched in [existing issues](https://github.com/onevcat/Kingfisher/issues?utf8=✓&q=is%3Aissue), but did not find a same one.
- [x] I want to report a problem instead of asking a question. It'd better to use [kingfisher tag in Stack Overflow](http://stackoverflow.com/questions/tagged/kingfisher) to ask a question.

#### What

I'm in need of a custom caching solution for videos. I need to cache the downloaded mp4 `Data` object to disk. I could implement my own solution, however I figured that I could use an instance of `DiskStorage.Backend<Data>`. The problem is it doesn't expose the methods using the `public` keyword, instead they are all internal. This makes me unable to store, retrieve, delete, or query if file is cached for a specific key. 

My request would be to consider making this class and all its methods public, so it could be used for caching more than just images and outside of the Kingfisher module. The `DiskStorage` class already has a robust caching implementation, it would be great if I could reuse this class for my own purposes.

onevcat:
Good idea. I'll do it soon.

onevcat:
Done in #1649

PR #2214, opened 2024-02-22
make removeSizeExceededValues public in disc storage

nickruddeni:
Much of the `DiscStorage.Backend` functionality was made public after this issue was raised https://github.com/onevcat/Kingfisher/issues/1602

`removeSizeExceededValues` is still `internal` and is not directly used anywhere within the class. As such, the sizeLimit set on the on the DiscStorage.Config is not respected.

I've created a fork do this for now - it would be great if this could be added.


onevcat:
Thanks for it!

PR #2228, opened 2024-04-05
Fix testDownsamplingWithEdgeCaseSize fail

kittisak-phetrungnapha:
In Xcode 15.3 and Swift 5.10, below code does not return `nil` if `kCGImageSourceThumbnailMaxPixelSize` is 0. It returns some image size `64*64` . I am not sure why and it does not make any sense. Maybe a bug from Apple?

```
let downsampleOptions: [CFString : Any] = [
            kCGImageSourceCreateThumbnailFromImageAlways: true,
            kCGImageSourceShouldCacheImmediately: true,
            kCGImageSourceCreateThumbnailWithTransform: true,
            kCGImageSourceThumbnailMaxPixelSize: 0 // even we set it to 0
        ]

public func CGImageSourceCreateThumbnailAtIndex(_ isrc: CGImageSource, _ index: Int, _ options: CFDictionary?) -> CGImage?
```

Before fixing:
Screenshot 2567-04-05 at 21 18 30

I just fix the test fail by assert with image size 64*64 and it's passed.

### DIFF
--- a/Tests/KingfisherTests/ImageExtensionTests.swift
+++ b/Tests/KingfisherTests/ImageExtensionTests.swift
@@ -331,9 +331,9 @@ class ImageExtensionTests: XCTestCase {
 
     func testDownsamplingWithEdgeCaseSize() {
 
-        // Zero size would fail downsampling.
-        let nilImage = KingfisherWrapper<KFCrossPlatformImage>.downsampledImage(data: testImageData, to: .zero, scale: 1)
-        XCTAssertNil(nilImage)
+        // Zero size would downsampling to 64x64 (previously nil).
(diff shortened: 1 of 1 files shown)


onevcat:
Thanks for it and I appreciate the effort you made for it.

I am preparing a major update currently and it has been also fixed here in the `v8` branch:

https://github.com/onevcat/Kingfisher/commit/d7505c68604d7a75edceb83c7d68d7e46aa2a514

As this PR seems to be duplicated to that fix and might fail the test if built with an earlier Xcode (which usually happens on CI), so maybe we can close this one without merging.

However, it is indeed a good idea to have this fix for the `master` branch as well. If you'd like, can you try to cherry pick that commit above and send another one to the `master`? Then we can merge it without worrying any further conflicting.

PR #2239, opened 2024-04-20
Strict concurrency is ready

onevcat:


onevcat:
Although the CI is failing, I will first merge it since it is a part of another feature branch `[strict-concurrency](https://github.com/onevcat/Kingfisher/tree/fix/strict-concurrency)`.

And the failing CI will be solved there.

PR #2273, opened 2024-07-12
Change `swift(>=6)` to use `compiler`

NachoSoto:
`@retroactive` is necessary when compiling with Swift 6, regardless of whether it's on Swift 5 mode.


onevcat:
Thank you! This makes sense. I’ll try to verify these changes once I have a stable network connection. I’m currently traveling.

PR #2284, opened 2024-09-07
remove redundant code

P4VEMENT:
Remove redundant assignments of `totalCostLimit` and `countLimit` to memory storage properties since the config setter already did it.

### DIFF
--- a/Sources/Cache/MemoryStorage.swift
+++ b/Sources/Cache/MemoryStorage.swift
@@ -84,8 +84,6 @@ public enum MemoryStorage {
         /// default expiration settings, and more.
         public init(config: Config) {
             self.config = config
-            storage.totalCostLimit = config.totalCostLimit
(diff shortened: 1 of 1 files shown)


onevcat:
@P4VEMENT Thanks for this.

However, if I recall correctly, the `didSet` of a stored property isn’t triggered when it’s assigned within the initializer. Could you confirm if this change is necessary and ensure it won’t alter the current behavior?

Or maybe we can have a better way to improve this, for example extract the duplicated code into a helper method, etc.

P4VEMENT:
As a beginner in the Swift language, thank you for the reminder. The official documentation states that "The willSet and didSet observers of superclass properties are called when a property is set in a subclass initializer, after the superclass initializer has been called. They aren’t called while a class is setting its own properties, before the superclass initializer has been called." I believe the current code is already sufficiently concise, so I will close this pull request.

PR #2347, opened 2025-01-18
Avoid recreating animated image representations when reloaded with identical options

yeatse:
Loading some animated image formats can be resource-intensive. To optimize performance, we should avoid reloading them from the disk cache if the loading options remain unchanged.

Fixes https://github.com/yeatse/KingfisherWebP/issues/85


onevcat:
Make sense! Thanks for it.

I'll prepare a tag very soon.

Issue #2352, opened 2025-02-03
Performance regression in 8.1.4


pNre:
Hello, after updating Kingfisher from version 8.1.3 to 8.1.4 I've started to notice frame drops while scrolling lists of animated images.
I'm not familiar enough with the codebase to point exactly at the change that caused the regression, but I've been able to replicate in a minimal example project:

8.1.3:
<img width="1188" alt="Image" src="https://github.com/user-attachments/assets/394b721b-737f-41a5-8660-802aa75842ae" />
8.1.4:
<img width="1198" alt="Image" src="https://github.com/user-attachments/assets/f8d8c334-b3db-4c16-8c93-23f1f1589580" />

The archive below contains the example project and an Instruments trace (where Run 1 is recorded on 8.1.3, Run 2 on 8.1.4):

[KF-frame-drop.zip](https://github.com/user-attachments/files/18646460/KF-frame-drop.zip)

onevcat:
I'll try to investigate this, thanks!

@yeatse Mostly it might be related to this #2347 , do you have any good idea?

onevcat:
A quick fix was implemented in [8.2.0](https://github.com/onevcat/Kingfisher/releases/tag/8.2.0). Let's test its effectiveness and explore further improvements.

PR #2354, opened 2025-02-04
Revert "Avoid recreating animated image representations when reloaded with identical options"

onevcat:
Reverts onevcat/Kingfisher#2347

This seems to be causing a serious performance downgrade as reported in #2352 

@yeatse 
I am not yet understanding the whole thing, but I will check to see what is happening here and then consider another way to implement a similar thing.


onevcat:
Close per #2357 provides a better fix!

PR #2356, opened 2025-02-04
:bug: Multiple requests on same URL would fail

MadsBogeskov:
If you have multiple simultaneous requests using Kingfisher that hits the same endpoint the task management system would fail before this. Simply put, because the tasks map, uses the URL as the key it wouldn't be able to store the different requests.

However Kingfisher does seem to handle this case in `task(for task: URLSessionTask) -> SessionDataTask?` by looking at both the URL and the task id. So I changed the key of the `tasks` map to include both for lookup.

Before this change Kingfisher would leak the continuation it creates in `downloadImage(with:options)` in ÌmageDownloader.

### DIFF
--- a/Sources/Networking/SessionDelegate.swift
+++ b/Sources/Networking/SessionDelegate.swift
@@ -31,7 +31,6 @@ import Foundation
 /// It also behaves like a task manager for downloading.
 @objc(KFSessionDelegate) // Fix for ObjC header name conflicting. https://github.com/onevcat/Kingfisher/issues/1530
 open class SessionDelegate: NSObject, @unchecked Sendable {
-
     typealias SessionChallengeFunc = (
         URLSession,
         URLAuthenticationChallenge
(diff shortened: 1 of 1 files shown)


onevcat:
Hi, @MadsBogeskov 

Thank you for this PR.

I believe Kingfisher is handling this correctly by storing the callbacks from different requests in the `callbacksStore` ([code](https://github.com/onevcat/Kingfisher/blob/master/Sources/Networking/SessionDataTask.swift#L61-L67)). This behavior, where multiple requests share the same URL, is also [tested here](https://github.com/onevcat/Kingfisher/blob/a09591d5e3566c3700cb53fa5f62b666225be443/Tests/KingfisherTests/ImageDownloaderTests.swift#L109-L126). By doing this, Kingfisher ensures that only one data download task is initiated for the same URL request, optimizing data transfer. Once the download is complete, it triggers all the registered callbacks.

Can you verify this behavior? Feel free to let me know if my understanding is wrong and this PR is actually necessary!

MadsBogeskov:
Hi @onevcat - Thanks for the quick response. I'll take a look. There is an issue, as I have a set of unit tests that are blocking, as in never returning to the callsite, because the tasks have disappeared, which my above PR fixed for our app at least. Plus we were often getting continuation leaks because of the same issue, so something is up (at least in the app I am working on ;) )

I will take a further look into what's going on!

onevcat:
Thanks for the declaration!

You remind me a fix in 8.1.2 (https://github.com/onevcat/Kingfisher/releases/tag/8.1.2). If you are using an older version, maybe it worths to upgrade and see if that fixes.

Naveen-C-Ramachandrappa (approved):

ankushkushwaha (approved):

147Floki (approved):

saltman007web (approved):

147Floki (approved):

PR #2357, opened 2025-02-04
Resolves performance issue when presenting animated images

### LINKED ISSUE #2352: Performance regression in 8.1.4

yeatse:
Fixes #2352 

The problem is that `AnimatedImageView` currently decodes GIF images on the main thread and doesn’t adhere to `kCGImageSourceShouldCacheImmediately`. This was already a performance concern, but after #2347 was merged, the issue worsened when multiple AnimatedImageView instances shared a single CGImageSource instance.

This PR tries to fix this problem with two changes:

1. When creating `AnimatedImageView`, we now make a copy of the `CGImageSource` instance to restore the performance to pre-#2347 levels.
2. We use `UIImage.preparingForDisplay` to decode animated image frames ahead of time, eliminating the need to decode them on the main thread.

Both changes independently solve the performance issue, but both are included here for discussion and further evaluation.

### DIFF
--- a/Sources/Image/GIFAnimatedImage.swift
+++ b/Sources/Image/GIFAnimatedImage.swift
@@ -148,6 +148,13 @@ public protocol ImageFrameSource {
     
     /// Retrieves the duration at a specific index. If the index is invalid, implementors should return `0.0`.
     func duration(at index: Int) -> TimeInterval
+    
+    /// Creates a copy of the current `ImageFrameSource` instance.
+    ///
+    /// - Returns: A new instance of the same type as `self` with identical properties.
+    ///            If not overridden by conforming types, this default implementation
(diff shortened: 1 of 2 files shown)


onevcat:
Wow! @yeatse Thanks for the information and quick fix! I will try this later!

onevcat:
Working perfect!

PR #2374, opened 2025-03-23
Refactor cacheFileName method

NeoSelf1:
I've refactored the cacheFileName method with the following improvements.
- Extracted the file extension setting logic into a separate method.
- Restructured the nested conditionals to clearly indicate the priority order of file extensions.
- Removed redundant forcedExtension check in the second condition from setFileExtension method since it's already been checked in the first condition. If we reach the second condition, forcedExtension is guaranteed to be nil.

I have verified that all test cases in DiskStorageTests pass after the refactoring. 
This is my first open source contribution, so please excuse any shortcomings. Thank you for your understanding.

### DIFF
--- a/Sources/Cache/DiskStorage.swift
+++ b/Sources/Cache/DiskStorage.swift
@@ -372,22 +372,26 @@ public enum DiskStorage {
         }
         
         func cacheFileName(forKey key: String, forcedExtension: String? = nil) -> String {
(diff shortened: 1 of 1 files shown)


onevcat (approved):

onevcat:
Perfect! Thank you for the contributing, it really helps!

PR #2391, opened 2025-05-30
SWIFT TASK CONTINUATION MISUSE fix

VladimirHorky:
We’re encountering an issue related to Swift concurrency when using Kingfisher. Specifically, we’re seeing the following runtime warning:

```
SWIFT TASK CONTINUATION MISUSE: retrieveImage(with:options:progressiveImageSetter:referenceTaskIdentifierChecker:) leaked its continuation without resuming it. 
```

This may cause tasks waiting on it to remain suspended forever.
Our concern is around proper handling of task cancellation. Shouldn't the continuation be resumed even when Task.isCancelled is true?

```
func retrieveImage(
        with source: Source,
        progressiveImageSetter: ((KFCrossPlatformImage?) -> Void)? = nil,
        referenceTaskIdentifierChecker: (() -> Bool)? = nil
    ) async throws -> RetrieveImageResult
    {
        let task = CancellationDownloadTask()
        return try await withTaskCancellationHandler {
            try await withCheckedThrowingContinuation { continuation in
                let downloadTask = retrieveImage(
                    with: source,
                    options: options,
                    downloadTaskUpdated: { newTask in
                        Task {
                            await task.setTask(newTask)
                        }
                    },
                    progressiveImageSetter: progressiveImageSetter,
                    referenceTaskIdentifierChecker: referenceTaskIdentifierChecker,
                    completionHandler: { result in
                        continuation.resume(with: result)
                    }
                )
                if Task.isCancelled {
                    downloadTask?.cancel()
                } else {
                    Task {
                        await task.setTask(downloadTask)
                    }
                }
            }
        } onCancel: {
            Task {
                await task.task?.cancel()
            }
        }
    }
```
    
    
This pull request refactors the `KingfisherManager` extension in `Sources/General/KingfisherManager.swift` to improve cancellation handling and ensure thread safety during asynchronous operations. The changes introduce a task-local actor for managing continuation state and enhance the cancellation logic.

### Improvements to cancellation handling:

* **Pre-check for cancellation:** Added a check for `Task.isCancelled` at the start of the method to avoid unnecessary processing if the task is already canceled.

* **Continuation management with actor:** Introduced a task-local actor (`ContinuationManager`) to safely manage the continuation state, ensuring thread-safe handling of asynchronous operations.

### Enhanced thread safety:

* **Safe continuation resumption:** Updated the completion handler to use the actor for resuming the continuation safely based on the result (`success` or `failure`).

* **Cancellation during setup:** Added logic to handle cancellation that might occur during the setup phase, ensuring proper cleanup and continuation resumption.

* **On-cancel logic:** Modified the `onCancel` block to use the actor for safely resuming the continuation with a `CancellationError`.

### DIFF
--- a/Sources/General/KingfisherError.swift
+++ b/Sources/General/KingfisherError.swift
@@ -363,6 +363,7 @@ public enum KingfisherError: Error {
     case processorError(reason: ProcessorErrorReason)
     /// Represents the error reason that can occur during image setting in a view related class.
     case imageSettingError(reason: ImageSettingErrorReason)
+    case swiftTaskCancelled
 
     // MARK: Helper Properties & Methods
 
@@ -452,6 +453,7 @@ extension KingfisherError: LocalizedError {
         case .cacheError(let reason): return reason.errorDescription
         case .processorError(let reason): return reason.errorDescription
         case .imageSettingError(let reason): return reason.errorDescription
+        case .swiftTaskCancelled: return "The task was cancelled."
         }
     }
 }
@@ -485,6 +487,7 @@ extension KingfisherError: CustomNSError {
         case .cacheError(let reason): return reason.errorCode
         case .processorError(let reason): return reason.errorCode
         case .imageSettingError(let reason): return reason.errorCode
+        case .swiftTaskCancelled: return 3333
         }
     }
 }
--- a/Sources/General/KingfisherError.swift
+++ b/Sources/General/KingfisherError.swift
@@ -363,7 +363,6 @@ public enum KingfisherError: Error {
     case processorError(reason: ProcessorErrorReason)
     /// Represents the error reason that can occur during image setting in a view related class.
     case imageSettingError(reason: ImageSettingErrorReason)
-    case swiftTaskCancelled
 
     // MARK: Helper Properties & Methods
 
@@ -453,7 +452,6 @@ extension KingfisherError: LocalizedError {
         case .cacheError(let reason): return reason.errorDescription
         case .processorError(let reason): return reason.errorDescription
         case .imageSettingError(let reason): return reason.errorDescription
-        case .swiftTaskCancelled: return "The task was cancelled."
         }
     }
 }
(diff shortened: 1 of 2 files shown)


onevcat:
Hi @VladimirHorky,

Thank you so much for taking the time to identify and address this Swift concurrency issue\! The `SWIFT TASK CONTINUATION MISUSE` warning is definitely something we need to fix, and I really appreciate your detailed analysis and proposed solution.

Your approach of using an actor to manage the continuation state is thoughtful and shows a good understanding of Swift concurrency patterns. The addition of early cancellation checks and proper error handling in the `onCancel` block are exactly the kind of improvements we need.

However, after reviewing the implementation, I've identified a potential race condition that could still lead to the same continuation leak under certain timing conditions. The issue is with this pattern:

```swift
Task {
    await continuationManager.setContinuation(continuation)
}
```

If task cancellation occurs before this async call completes, the cancellation handler might try to resume a continuation that hasn't been set in the manager yet, which could still result in a leaked continuation.

I think the current approach might be a bit more complex than we need. Let me take this over and see if I can find a simpler solution that avoids the race condition entirely. I'll keep this PR open for reference while I work on an alternative approach.

Thanks again for spotting this issue and doing the hard work of analyzing the problem\! Swift concurrency edge cases like this are always tricky to get right, and your contribution really helps us improve Kingfisher's reliability.

onevcat:
Hi @NeoSelf1\! 👋 

Thank you for identifying and working on this Swift Task Continuation Misuse issue\! Your analysis was spot-on and really helped us understand the root cause.

After reviewing your solution, I've implemented a simpler approach that addresses the same problem while avoiding the potential race conditions you mentioned. The new solution uses a local actor for continuation state management and is fully compatible with Swift 6 strict concurrency.

You can check out the implementation in **PR #2398** which:
- Uses a more direct actor-based approach without separate continuation management layers  
- Eliminates the race condition concerns you raised
- Passes 100x repetition testing to ensure the issue is completely resolved
- Maintains full Swift 6 compatibility

Thanks again for bringing this important issue to our attention\! Your contribution really helped make Kingfisher more robust. 🚀

Issue #2323, opened 2024-11-20
KingfisherManager.retrieveImage() hits disk synchronously even when opted into all async options


jotai-coder:
- [x] I have read the [wiki page](https://github.com/onevcat/Kingfisher/wiki) and [cheat sheet](https://github.com/onevcat/Kingfisher/wiki/Cheat-Sheet), but there is no information I need.
- [x] I have searched in [existing issues](https://github.com/onevcat/Kingfisher/issues?utf8=✓&q=is%3Aissue), but did not find a same one.
- [x] I want to report a problem instead of asking a question. It'd better to use [kingfisher tag in Stack Overflow](http://stackoverflow.com/questions/tagged/kingfisher) to ask a question.

### Issue Description
KingfisherManager.retrieveImage() hits disk even when options are set to async

#### What
KingfisherManager.retrieveImage() hits disk on calling thread even when options are set to async when checking disk cache for presence of image. The issue with this is that KFImage() and image.kf.setImage() must be used on Main thread. This contributes to dropped frames. In UIKit, using KingfisherManager.retrieveImage manually (on a secondary thread), and then setting the image is simple workaround, but in SwiftUI it becomes more cumbersome.

#### Reproduce
v7.12
Use KFImage or kf.setImage in a large scrolling list. Observe hitches due to disk access on main 
<img width="1665" alt="Screenshot 2024-11-20 at 11 44 41 AM" src="https://github.com/user-attachments/assets/95f589b1-0dc2-4a6f-8529-d309a407d636">
thread.

PR #2418, opened 2025-07-29
Add async disk cache existence check to avoid main thread blocking

HarrisHan:
## Summary

  This PR addresses the main thread blocking issue (#2323) by adding asynchronous disk cache existence checking. The current implementation blocks the main thread when checking if an image exists in disk cache, causing frame drops in scrolling lists.

  ## Problem

  As reported in #2323:
  - `KingfisherManager.retrieveImage()` performs disk cache checks on the calling thread
  - When called from the main thread (common in UI code), this causes synchronous disk I/O
  - Results in frame drops and scrolling performance issues, especially in SwiftUI

  ## Solution

  ### New APIs

  1. **`DiskStorage.itemExists()`** - Lightweight existence check without loading data

 ```swift
   
  func itemExists(forKey key: String, referenceDate: Date = Date(), forcedExtension: String? = nil) -> Bool


  2. ImageCache.cachedType() - Async cache type checking
  public func cachedType(
      forKey key: String,
      processorIdentifier: String = DefaultImageProcessor.default.identifier,
      forcedExtension: String? = nil,
      callbackQueue: CallbackQueue = .mainCurrentOrAsync,
      completionHandler: @escaping @Sendable (CacheType) -> Void)

  Implementation Details

  - Memory cache check remains synchronous (fast, no I/O)
  - Disk cache check is dispatched to background queue
  - Results are delivered on specified callback queue
  - Maintains full backward compatibility

  Usage Example

  // Before (blocks main thread)
  let image = cache.retrieveImageInDiskCache(forKey: key)

  // After (non-blocking)
  cache.cachedType(forKey: key) { cacheType in
      switch cacheType {
      case .memory:
          // Image in memory
      case .disk:
          // Image on disk, load if needed
      case .none:
          // Not cached, download required
      }
  }

```
  Testing

  Added comprehensive tests in ImageCacheAsyncTests.swift:
  - ✅ Correct cache type detection (memory/disk/none)
  - ✅ Main thread non-blocking verification
  - ✅ Performance testing with 1000+ items
  - ✅ Thread safety validation

  Performance Impact

  - No impact on existing synchronous APIs
  - Eliminates main thread blocking for existence checks
  - Optimized disk check without loading full image data
  - Maintains existing cache performance characteristics

  Migration Guide

  This is a non-breaking change. Existing code continues to work. For better performance:
```swift
  // Instead of checking synchronously
  if let _ = imageCache.retrieveImageInDiskCache(forKey: key) {
      // Image exists
  }

  // Use async check
  imageCache.cachedType(forKey: key) { type in
      if type != .none {
          // Image exists
      }
  }
```
  Checklist

  - Code compiles without warnings
  - All existing tests pass
  - Added new test coverage
  - API follows Kingfisher conventions
  - Thread-safe implementation
  - Backward compatible

### DIFF
--- a/Sources/Cache/DiskStorage.swift
+++ b/Sources/Cache/DiskStorage.swift
@@ -235,6 +235,45 @@ public enum DiskStorage {
             )
         }
 
+        /// Check whether an item exists for the given key without actually loading it.
+        /// This method is optimized for quick existence checks.
+        /// - Parameters:
+        ///   - key: The key of the item.
+        ///   - referenceDate: A reference date to check whether the cached item is expired or not.
+        ///   - forcedExtension: A forced extension to be appended to the file name.
+        /// - Returns: `true` if the item exists and is not expired, `false` otherwise.
+        func itemExists(
+            forKey key: String,
+            referenceDate: Date = Date(),
+            forcedExtension: String? = nil
+        ) -> Bool {
+            let fileManager = config.fileManager
+            let fileURL = cacheFileURL(forKey: key, forcedExtension: forcedExtension)
+            let filePath = fileURL.path
+            
+            // Quick check using maybeCached set
+            let fileMaybeCached = maybeCachedCheckingQueue.sync {
+                return maybeCached?.contains(fileURL.lastPathComponent) ?? true
+            }
+            guard fileMaybeCached else {
+                return false
+            }
+            
+            // Check file existence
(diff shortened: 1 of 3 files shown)


darkbrewx:
I've considered the same thing, but when I tried to implement it, I found it would require converting the entire retrieve call chain to callback style, turns out to be more complicated than I expected.

PR #2439, opened 2025-09-23
Feature - Network retry strategy

komkovla:
# Add NetworkRetryStrategy with NetworkMonitor

## Overview

This PR introduces a new `NetworkRetryStrategy` that observes network connectivity and retries image downloads when network becomes available after disconnection. The implementation includes a comprehensive `NetworkMonitor` system with testability through dependency injection.

### NetworkRetryStrategy
- **Network-aware retry strategy** that waits for network reconnection instead of using fixed delays
- **Immediate retry** when network is available, no artificial delays
- **Configurable timeout** for waiting for network reconnection
- **Automatic observer management** with proper cleanup

### NetworkMonitor System
- **Protocol-oriented design** with `NetworkMonitoring` and `NetworkObserver` protocols
- **Singleton pattern** with `NetworkMonitor.default` for shared network monitoring
- **Lazy initialization** - only starts monitoring when first accessed
- **Thread-safe implementation** using concurrent queues
- **Timeout support** for network reconnection waiting

### Basic Usage
```swift
// Uses default NetworkMonitor
let strategy = NetworkRetryStrategy()

// With timeout
let strategy = NetworkRetryStrategy(timeoutInterval: 30.0)

// With custom network monitor
let customMonitor = TestNetworkMonitor(isConnected: false)
let strategy = NetworkRetryStrategy(networkMonitor: customMonitor)
```

### Integration with Kingfisher
```swift
imageView.kf.setImage(
    with: url,
    options: [.retryStrategy(NetworkRetryStrategy())]
)
```

### Test Coverage
- ✅ Immediate retry when network is connected
- ✅ Wait for reconnection when network is disconnected
- ✅ Timeout handling for network reconnection
- ✅ Observer cancellation on subsequent retries
- ✅ Error type filtering (only retries response errors)
- ✅ Task cancellation handling

### Thread Safety
- All network monitoring operations use concurrent queues
- `@unchecked Sendable` for `NWPathMonitor` and GCD properties
- Proper weak references to prevent retain cycles

### Error Handling
- Only retries `KingfisherError.responseError` cases
- Stops retrying on user cancellation
- Handles timeout scenarios gracefully
- Proper observer cleanup on errors

## 📊 Performance Impact

- **Minimal overhead** - singleton pattern prevents multiple NWPathMonitor instances
- **Lazy initialization** - no resources used until first NetworkRetryStrategy is created

### No Breaking Changes
- Existing `DelayRetryStrategy` continues to work unchanged
- `NetworkRetryStrategy` is additive - no existing code needs modification
- Default behavior uses shared `NetworkMonitor.default` instance

## 🎯 Benefits

1. **Better User Experience** - Images retry immediately when network is restored
2. **Reduced Battery Usage** - No unnecessary polling or fixed delays

## 📋 Files Changed

- `Sources/Networking/NetworkMonitor.swift` - New network monitoring system
- `Sources/Networking/RetryStrategy.swift` - Added NetworkRetryStrategy
- `Tests/KingfisherTests/RetryStrategyTests.swift` - Add NetworkRetryStrategy tests

### DIFF
--- a/Sources/Networking/NetworkMonitor.swift
+++ b/Sources/Networking/NetworkMonitor.swift
@@ -28,7 +28,7 @@ import Network
 import Foundation
 
 /// A protocol for network connectivity monitoring that allows for dependency injection and testing.
-public protocol NetworkMonitoring: Sendable {
+internal protocol NetworkMonitoring: Sendable {
     /// Whether the network is currently connected.
     var isConnected: Bool { get }
 
@@ -41,7 +41,7 @@ public protocol NetworkMonitoring: Sendable {
 }
 
 /// A protocol for network observers that can be cancelled.
-public protocol NetworkObserver: Sendable {
+internal protocol NetworkObserver: Sendable {
     /// Cancels the network observation.
     func cancel()
 }
@@ -49,11 +49,11 @@ public protocol NetworkObserver: Sendable {
 /// A shared singleton that manages network connectivity monitoring.
 /// This prevents creating multiple NWPathMonitor instances when many NetworkRetryStrategy instances are used.
 /// The monitor is created lazily only when first accessed.
-public final class NetworkMonitor: @unchecked Sendable, NetworkMonitoring {
-    public static let `default` = NetworkMonitor()
+internal final class NetworkMonitor: @unchecked Sendable, NetworkMonitoring {
+    static let `default` = NetworkMonitor()
 
     /// Whether the network is currently connected.
-    public var isConnected: Bool {
+    var isConnected: Bool {
         return monitor.currentPath.status == .satisfied
     }
 
@@ -173,7 +173,7 @@ internal final class NetworkObserverImpl: @unchecked Sendable, NetworkObserver {
         }
     }
 
-    public func cancel() {
+    func cancel() {
         queue.async { [weak self] in
             guard let self else { return }
 
--- a/Sources/Networking/NetworkMonitor.swift
+++ b/Sources/Networking/NetworkMonitor.swift
@@ -1,0 +1,188 @@
+//
+//  NetworkMonitor.swift
+//  Kingfisher
+//
+//  Created by Vladislav Komkov on 2025/09/22.
+//
+//  Copyright (c) 2020 Wei Wang <onevcat@gmail.com>
+//
+//  Permission is hereby granted, free of charge, to any person obtaining a copy
+//  of this software and associated documentation files (the "Software"), to deal
+//  in the Software without restriction, including without limitation the rights
+//  to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+//  copies of the Software, and to permit persons to whom the Software is
+//  furnished to do so, subject to the following conditions:
+//
+//  The above copyright notice and this permission notice shall be included in
+//  all copies or substantial portions of the Software.
+//
+//  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+//  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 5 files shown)


Sources/Networking/RetryStrategy.swift
```diff
@@ -184,3 +184,86 @@ public struct DelayRetryStrategy: RetryStrategy {
...
+    // MARK: - Private helpers
+
+    private func waitForReconnection(
+        context: RetryContext,
+        retryHandler: @escaping @Sendable (RetryDecision) -> Void
+    ) {
+        let observer = networkMonitor.observeConnectivity(timeoutInterval: timeoutInterval) { isConnected in
+            if isConnected {
+                // Connection is restored, retry immediately
+                retryHandler(.retry(userInfo: context.userInfo))
```
onevcat, line 259: Do we have a memory leak here caused by a circular reference? 

The `observeConnectivity` returns an `observer` whose block holds `context`, while the `context.userInfo` holds the `observer`. Can you check it and verify if `.stop` is hit, how the context and the observer can be released?

Sources/Networking/NetworkMonitor.swift
```diff
@@ -0,0 +1,188 @@
...
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+
+import Network
+import Foundation
+
+/// A protocol for network connectivity monitoring that allows for dependency injection and testing.
+public protocol NetworkMonitoring: Sendable {
```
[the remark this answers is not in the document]
onevcat: Thank you for this! Making it internal first is a good decision. If necessary, let's public it later.

onevcat (changes requested):
@komkovla 

Thank you very much for this PR; the feature looks great. However, I noticed a possible memory leak—could you help verify it? (I think a `[weak context]` or a reset of`context.userInfo` may work, but need some verification, since I could be wrong.)

Sources/Networking/RetryStrategy.swift
komkovla: Oh yes! Indeed, a memory leak was happening. I've reproduced the issue with a test that demonstrates the circular reference:
```swift
func testNetworkRetryStrategyMemoryLeakWithLongTimeout() {
    for _ in 0...10 {
        let exp = expectation(description: #function)
        let source = Source.network(URL(string: "url")!)
        let networkMonitor = TestNetworkMonitor(isConnected: false)
        let retry = NetworkRetryStrategy(
            timeoutInterval: 30, // Long timeout to keep observers alive
            networkMonitor: networkMonitor
        )

        // Create multiple contexts and retry attempts to amplify the leak
        var contexts: [RetryContext] = []
        var completedCount = 0
        let totalContexts = 10

        for _ in 0..<totalContexts {
            let newContext = RetryContext(
                source: source,
                error: .responseError(reason: .URLSessionError(error: E()))
            )
            contexts.append(newContext)

            retry.retry(context: newContext) { decision in
                completedCount += 1
                if completedCount == totalContexts {
                    exp.fulfill()
                }
            }
        }

        // Simulate network reconnection after a delay to trigger all callbacks
        DispatchQueue.main.asyncAfter(deadline: .now() + 1.0) {
            networkMonitor.simulateNetworkChange(isConnected: true)
        }

        waitForExpectations(timeout: 5, handler: nil)
    }
}
```

After profiling this test with instruments memory leaks were discovered, running this test with fix from https://github.com/onevcat/Kingfisher/pull/2439/commits/4f634fd30aa93745ad748c63972ef6dcbcdbbba9 fixes issue and no leaks were detected. 

I have removed this test from testSuit because it would not fail on memory leak and after issue is fixed don't provide value. Thanks for pointing out this issue!

onevcat (approved):
LGTM!

PR #2448, opened 2025-11-16
fix/macos 10 bit

### LINKED ISSUE #2447: Kingfisher isn't able to render certain images on Mac that are 10 bits per component
Certain images never render on macOS.

Having an image with with 10 bits per component always returns nil here
https://github.com/onevcat/Kingfisher/blob/4d75de347da985a70c63af4d799ed482021f6733/Sources/Image/GraphicsContext.swift#L53

NSBitmapImageRep bitsPerSample only supports values `1, 2, 4, 8, 12, or 16` according to the docs:
https://developer.apple.com/documentation/appkit/nsbitmapimagerep/init(bitmapdataplanes:pixelswide:pixelshigh:bitspersample:samplesperpixel:hasalpha:isplanar:colorspacename:bitmapformat:bytesperrow:bitsperpixel:)

onevcat:
This fixes #2447 

- Improve macOS graphics context for high bit depth
- Add pixel format regression tests

### DIFF
--- a/Tests/KingfisherTests/PixelFormatDecodingTests.swift
+++ b/Tests/KingfisherTests/PixelFormatDecodingTests.swift
@@ -5,73 +5,21 @@ import XCTest
 final class PixelFormatDecodingTests: XCTestCase {
     private struct Sample {
         let fileName: String
-        let expectedBitsOnMac: Int
-        #if os(macOS)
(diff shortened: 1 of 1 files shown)


onevcat:
@codex review

PR #2497, opened 2026-03-23
Fix stale disk cache retrieval for cancelled or superseded tasks

### LINKED ISSUE #2495: Images loading from cache are not canceled
Scrolling through a large collection of images quickly explodes the memory usage (>1gb). I believe its because cached images go through deserialisation even if the download task is canceled using `imageView.kf.cancelDownloadTask()`. I am using yeatse/KingfisherWebP package and can see `ImageCache.retrieveImageInDiskCache` being called for each displayed item.

Image

onevcat:
## Summary
Fixes #2495.

This PR addresses the stale disk-cache retrieval problem where a request can be cancelled or superseded after a disk hit is queued, but still continues through deserialization, background decode, and memory promotion.

In the affected paths, `cancelDownloadTask()` or a later `setImage` call could stop the network work, but there was no equivalent stale signal for the disk-cache path. As a result, an outdated request could still perform expensive work and populate memory cache with an image that no longer belonged to the current task.

## What Changed
- add a thread-safe `CancellationToken` and pass it through view-loading paths instead of relying on background reads of UI state
- check for stale requests at three disk-cache checkpoints: before disk work starts, after disk read but before deserialization, and after deserialization but before background decode / promotion
- introduce internal `DiskRetrievalOutcome.stale` to distinguish an intentionally skipped stale retrieval from a genuine cache miss inside `ImageCache`
- keep the public `ImageCacheResult` boundary unchanged by mapping stale outcomes back to `.none`, with manager-layer stale checks preserving the semantics for manager-mediated loading paths
- short-circuit stale results in `KingfisherManager` so they do not enter retry, alternative source, or original-cache fallback flows
- make `cancelDownloadTask()` also cancel the token so explicit cancel works for disk-cache hits as well as network requests
- extend the token-based stale handling to the related button/live photo/text attachment/component integration paths

## Tests
Added and strengthened regression coverage for:
- stale disk-cache retrieval skipping deserialization and memory promotion
- stale requests not triggering retry, alternative sources, or original-cache fallback download
- single-completion behavior for superseded requests
- explicit `cancelDownloadTask()` handling on disk-cache hits
- button and image view integration paths that supersede or cancel a disk-cached request
- deterministic serializer coordination helpers for timing-sensitive stale cache tests

Also includes follow-up cleanup for the new tests and Swift 6 warning fixes in the affected test code.

## Documentation
- add and keep the issue analysis document in `docs/issue-2495-analysis.md`

## Validation
- `bundle exec fastlane tests`

### DIFF
--- a/Sources/Extensions/PHLivePhotoView+Kingfisher.swift
+++ b/Sources/Extensions/PHLivePhotoView+Kingfisher.swift
@@ -201,7 +201,7 @@ extension KingfisherWrapper where Base: PHLivePhotoView {
         mutatingSelf.cancellationToken?.cancel()
         mutatingSelf.cancellationToken = token
 
-        let taskIdentifierChecking = { !token.isCancelled }
+        let taskIdentifierChecking: @Sendable () -> Bool = { !token.isCancelled }
 
         // Copy these associated values to prevent issues from reentrance.
         let targetSize = targetSize
--- a/Sources/General/KingfisherManager+LivePhoto.swift
+++ b/Sources/General/KingfisherManager+LivePhoto.swift
@@ -87,7 +87,7 @@ extension KingfisherManager {
         with source: LivePhotoSource,
         options: KingfisherOptionsInfo? = nil,
         progressBlock: DownloadProgressBlock? = nil,
-        referenceTaskIdentifierChecker: (() -> Bool)? = nil
+        referenceTaskIdentifierChecker: (@Sendable () -> Bool)? = nil
     ) async throws -> LivePhotoLoadingInfoResult {
         let fullOptions = currentDefaultOptions + (options ?? .empty)
         var checkedOptions = KingfisherParsedOptionsInfo(fullOptions)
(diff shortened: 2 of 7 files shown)


onevtail (approved):
Great fix — I reviewed this with focus on the stale disk-cache cancellation path and code quality, and it looks solid.

What I like:
- The root issue is addressed at the right layer: stale work is now blocked before expensive deserialize/decode + memory promotion.
- `CancellationToken` is a clean improvement over reading UI-associated state from background queues.
- `DiskRetrievalOutcome.stale` keeps stale semantics explicit internally and avoids conflating stale with normal cache-miss behavior.
- Manager-side short-circuiting for stale results is well placed and prevents retry / alternative-source / path-2 fallback work.
- Test coverage is excellent (retry, alternative sources, superseded requests, cancel APIs, UIKit/AppKit integration paths).

No blocking issues from my side.

Optional follow-up (non-blocking): it could be useful to keep a small benchmark note (or a reproducible profiling script) for the #2495 scrolling case, so future refactors can verify this memory-behavior win stays intact.

PR #2516, opened 2026-04-17
Fix macOS crash when decoding images with indexed color space

### LINKED ISSUE #2467: Crash in macOS (26) version 8.6.2
Crash in macOS (26) version 8.6.2

CGBitmapContextInfoCreate: CGColorSpace doesn't support output

Kingfisher/GraphicsContext.swift:51: Fatal error: [Kingfisher] Image context cannot be created.

Image

friday-refined:
## Summary

Fixes #2467.

On macOS, `BitmapContextDescriptor.resolveColorSpace` could hand an **indexed** (or, in theory, **pattern** / **unknown**) color space back to `CGBitmapContextCreate`, which rejects those as destination color spaces and logs:

```
CGBitmapContextInfoCreate: CGColorSpace doesn't support output
```

That then trips the `assertionFailure` in `GraphicsContext.current` and crashes (or silently returns a broken image in release) on any call that funnels through the macOS redraw path, most notably `image.kf.decoded` and `.backgroundDecode(true)`.

This also lines up with the analysis from @haugli in the issue thread:

> Seems like it's failing because the image uses `kCGColorSpaceIndexed`, which is valid for a source image but not for a destination bitmap context.

## Root cause

In `Sources/Image/GraphicsContext.swift`:

```swift
private static func resolveColorSpace(from cgImage: CGImage?) -> CGColorSpace {
    guard let cgColorSpace = cgImage?.colorSpace else {
        return CGColorSpaceCreateDeviceRGB()
    }
    let components = cgColorSpace.numberOfComponents
    if components == 1 || components == 3 {
        return cgColorSpace            // ← indexed PNG lands here: components == 1
    }
    return CGColorSpaceCreateDeviceRGB()
}
```

An 8‑bit palette PNG decodes to `CGColorSpaceModel.indexed` with `numberOfComponents == 1`, so the original space was returned as‑is and fed straight into `CGBitmapContextCreate`, which does not accept indexed / pattern color spaces as destination.

## Fix

Reject `.indexed`, `.pattern`, and `.unknown` color space models early and fall back to `deviceRGB` before the component-count check. Everything else (sRGB, Display P3, generic gray / RGB, generic CMYK, …) keeps its existing behavior.

```swift
switch cgColorSpace.model {
case .indexed, .pattern, .unknown:
    return CGColorSpaceCreateDeviceRGB()
default:
    break
}
```

This is a correct semantic for `kf.decoded`: its job is to pre-decode into a bitmap, and expanding an indexed palette into RGB during that redraw is the standard treatment.

## Test

Added a macOS-only regression test (`testDecodingRedrawsIndexedColorSpaceImageToDeviceRGB`) with a small indexed-palette PNG fixture. The fixture reliably reproduces the fatal error on `master` (macOS 26.4 SDK):

```
[xctest] CGBitmapContextInfoCreate: CGColorSpace doesn't support output
Kingfisher/GraphicsContext.swift:51: Fatal error: [Kingfisher] Image context cannot be created.
```

After the fix:

| Platform | Destination | Tests | Result |
| --- | --- | --- | --- |
| macOS (arm64, macOS 26.4 SDK) | `platform=macOS,arch=arm64` | 304 | ✅ all pass |
| iOS (arm64, iOS 26.3.1 Simulator) | iPhone 17 Pro | 321 | ✅ all pass |

The new test asserts:

- the fixture loads as `CGColorSpaceModel.indexed` (confirms the reproducer is real),
- `image.kf.decoded` does not fatal-error,
- the resulting `CGImage` no longer has an indexed color space,
- width / height are preserved.

## Impact / risk

- **Scope:** macOS only. All non-macOS platforms use `CGContext.fresh` (device RGB + premultipliedLast), which is unaffected.
- **Only previously-broken paths change behavior.** `.indexed` / `.pattern` / `.unknown` as destinations always failed under `CGBitmapContext`; this PR turns those from "crash / return nil" into "redraw into deviceRGB".
- **Every commonly used color space** (sRGB, Display P3, generic gray, generic RGB, generic CMYK, Lab, XYZ, deviceN, etc.) is left on the existing code path and is covered by the pre-existing `testDecodingSupportsVariousPixelFormats`, which still passes.


enums:
@jarvis-elevated please help to review.

jarvis-elevated (approved):
@enums Reviewed. This fix looks solid and addresses the root cause correctly.

What I checked:
- `resolveColorSpace(from:)` now guards unsupported destination models (`.indexed`, `.pattern`, `.unknown`) and falls back to `DeviceRGB` before context creation.
- Existing behavior for normal color spaces remains intact.
- New macOS regression test reproduces the original indexed-color failure path and verifies the decoded output is no longer indexed while preserving dimensions.
- Test fixture is correctly added to the test bundle.

No blocking issues found from my side. Approving ✅

PR #2530, opened 2026-06-05
Add `MemoryStorage.swift` method `totalCacheCost`

aronspringfield:
## Summary

Added a new function to `MemoryStorage` to get `totalCacheCost`, which adds up the `cacheCost` of all objects, valid and expired, that exist in `MemoryStorage`.

This also provides parity with `DiskStorage`, where it is able to give us the `totalSize` of the cache.

## Problem

When setting `totalCostLimit` on `MemoryStorage`, it feels like a bit of a 'finger in the air' type approach to guessing what an appropriate limit may be. So to give us some tangible data and reference points, being able to determine what our `totalCacheCost` is, can allow us to make better-informed decisions. It's also quite helpful to be able to show the User this information. 

## Solution

`totalCacheCost` adds up the `cacheCost` of all objects that exist in the `MemoryStorage` cache, without triggering `estimatedExpiration` to be extended. 

## Files Changes

- `Sources/Cache/MemoryStorage.swift` - Added the new `totalCacheCost` function here to provide an estimate of our in-memory footprint. 
-  `Sources/Cache/Storage.swift` - Added a new type `ExpirationFilter` to define how to filter object validity. 
- `Tests/KingfisherTests/MemoryStorageTests.swift` - Added two additional tests to support this new `totalCacheCost` function. 

### Note

I'm aware `DiskStorage` uses the name `totalSize` and this function is using the term `totaCacheCost`. This aligns with the naming convention used when applying cache limits. It's also worth being aware that there is a small distinction in that `cacheCost` is an estimated memory size, not an actual size. So they're strictly not the same.

### DIFF
--- a/Sources/Cache/MemoryStorage.swift
+++ b/Sources/Cache/MemoryStorage.swift
@@ -200,6 +200,25 @@ public enum MemoryStorage {
             storage.removeAllObjects()
             keys.removeAll()
         }
+
+        /// Calculates the total `cacheCost` for all currently cached values.
+        ///
+        /// This method can be expensive so should be used sparingly.
+        /// - Returns: The total cost in bytes for valid cached values.
+        public func totalCacheCost() -> Int {
+            let allKeys: [String] = {
+                lock.lock()
+                defer { lock.unlock() }
+                return Array(keys)
+            }()
+            let totalCost: Int = allKeys.reduce(0) { cost, key in
+                if let value = self.value(forKey: key, extendingExpiration: .none) {
+                    return cost + value.cacheCost
+                }
+                return cost
+            }
+            return totalCost
+        }
     }
 }
 
--- a/Sources/Cache/MemoryStorage.swift
+++ b/Sources/Cache/MemoryStorage.swift
@@ -159,14 +159,26 @@ public enum MemoryStorage {
         ///
         /// - Parameters:
         ///   - key: The cache key of the value.
+        ///   - filter: The expiration filter used by this retrieval action. The default value is `.valid`, which gets only non-expired values.
         ///   - extendingExpiration: The expiration policy used by this retrieval action.
         /// - Returns: The value under `key` if it is valid and found in the storage. Otherwise, `nil`.
-        public func value(forKey key: String, extendingExpiration: ExpirationExtending = .cacheTime) -> T? {
+        public func value(forKey key: String,
+                          filter: ExpirationFilter = .valid,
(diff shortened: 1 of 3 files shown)


onevcat:
Thanks for adding this. The narrow implementation looks reasonable for reporting the cost of currently valid memory cache values, and the new tests cover the basic removal and expiration behavior.

I think the API semantics need one clarification before merging though.

`totalCacheCost()` currently calls `value(forKey:extendingExpiration: .none)`, so expired-but-still-resident objects are excluded. That matches “valid cached values”, but it does not represent the memory currently held by `NSCache`, nor exactly the cost that `totalCostLimit` is managing, because expired objects can remain in memory until `removeExpired()` or the clean timer runs.

If the intent is to expose usable cache cost, the docs should make that explicit and avoid positioning it as memory footprint / `totalCostLimit` guidance. If the intent is to help tune memory cost limits, we probably need to count resident objects and preferably store the insertion cost on `StorageObject`, since that is the value passed to `NSCache.setObject(..., cost:)`.

One minor formatting issue as well: `git diff --check master...HEAD` reports trailing whitespace in `Sources/Cache/MemoryStorage.swift`.

Local targeted test passed for me:

```text
xcodebuild test -workspace Kingfisher.xcworkspace -scheme Kingfisher -destination 'platform=macOS' -only-testing:KingfisherTests/MemoryStorageTests SWIFT_VERSION=5.0
```

aronspringfield:
@onevcat that's a very good point. You're right. `totalCacheCost` should genuinely be the _total_ as stated by the function name. I've modified the PR so that `totalCacheCost` accounts for all values, including expired. This does widen the scope of the change very slightly to include a modification to the `value(forKey: ...)` function, by adding a new parameter type `ExpirationFilter` to determine whether expired should be allowed, which defaults to `valid`, which maintain existing behaviour and method signature, but allows `totalCacheCost` to use this function by including expired values by using the `all` `ExpirationFilter` option. 

`totalCacheCost` is limited in scope to `MemoryStorage` within `Backend<T: CacheCostCalculable>`. The generic is already constrained to adhere to `CacheCostCalculable`, so `value.cacheCost` (where `value: T`) is already possible. If we were to add `cacheCost` or `CacheCostCalculable` to `StorageObject` as you say, are we not just adding complexity to `StorageObject` without any gain? 

Apologies for the formatting issue, I hadn't spotted it. My linter isn't setup to run on this repo, and I have my GUI to ignore whitespace. I have now corrected this. 

All tests (including the newly added ones) are passing for me with command provided.

PR #2531, opened 2026-06-06
Absorb cancellation token lifecycle into TaskPropertyAccessor

### LINKED ISSUE #2496: Refactor TaskPropertyAccessor to absorb cancellation token lifecycle
## Summary

`KingfisherWrapper.setImage(...)` currently passes `previousCancellationToken` and `setCancellationToken` as standalone parameters in addition to `TaskPropertyAccessor`.

These token-related parameters belong to the same task lifecycle concern as:
- `setTaskIdentifier`
- `getTaskIdentifier`
- `setTask`

The current shape works, but it creates parameter sprawl and makes the helper signature harder to read.

## Proposed Direction

Fold cancellation-token access into `TaskPropertyAccessor` so the helper receives a single lifecycle accessor instead of splitting task state across:
- `taskAccessor`
- `previousCancellationToken`
- `setCancellationToken`

Possible shape:
- add token getter/setter accessors to `TaskPropertyAccessor`
- let the helper cancel the previous token and store the new token through that accessor

## Scope

This is a readability / maintainability cleanup only.
It should not change behavior.

## Non-Goals

- do not change the current stale-disk-cache semantics
- do not refactor unrelated associated-object patterns in this issue
- do not change public API surface

## Affected Files

Likely at least:
- `Sources/Extensions/HasImageComponent+Kingfisher.swift`
- `Sources/Extensions/ImageView+Kingfisher.swift`
- `Sources/Extensions/NSTextAttachment+Kingfisher.swift`
- `Sources/Extensions/CPListItem+Kingfisher.swift`
- `Sources/Extensions/UIButton+Kingfisher.swift`
- `Sources/Extensions/NSButton+Kingfisher.swift`

## Acceptance

- helper signature becomes simpler
- token lifecycle stays encapsulated with other task lifecycle accessors
- no behavior change
- existing stale cache tests continue to pass

nenadvulic:
## What

Fixes #2496.

The shared `setImage` helper received `previousCancellationToken` and `setCancellationToken` as standalone parameters next to `TaskPropertyAccessor`, although they belong to the same task lifecycle concern as `setTaskIdentifier` / `getTaskIdentifier` / `setTask`.

This PR folds them into `TaskPropertyAccessor`:

- `getCancellationToken: @Sendable @MainActor () -> CancellationToken?`
- `setCancellationToken: @Sendable @MainActor (CancellationToken) -> Void`

The helper now cancels the previous token and stores the new one through the accessor. The six call sites (`HasImageComponent`, `UIButton` ×2, `NSButton` ×2, `CPListItem`) are updated accordingly — each site's accessor reads/writes the same property it previously captured.

Per the issue's non-goals, `ImageView+Kingfisher.swift` and `NSTextAttachment+Kingfisher.swift` are untouched: they keep their own inline implementations and never passed tokens to the shared helper.

## Behavior

No public API change, no behavior change. The only theoretical difference: the previous token is now read at cancellation time instead of being captured at the start of the call — equivalent on all paths, since the code in between is synchronous on the `MainActor`.

## Tests

Full suite run before and after, same results:

| Destination | Before (master) | After |
|---|---|---|
| macOS | 323 passed, 0 failed | 323 passed, 0 failed |
| iOS Simulator (iPhone 17) | 342 passed, 0 failed | 342 passed, 0 failed |

Stale cache tests included, per the acceptance criteria. Diff: 4 files, +23/−29.

### DIFF
--- a/Sources/Extensions/CPListItem+Kingfisher.swift
+++ b/Sources/Extensions/CPListItem+Kingfisher.swift
@@ -110,7 +110,6 @@ extension KingfisherWrapper where Base: CPListItem {
         completionHandler: (@MainActor @Sendable (Result<RetrieveImageResult, KingfisherError>) -> Void)? = nil) -> DownloadTask?
     {
         var mutatingSelf = self
-        let previousToken = mutatingSelf.cancellationToken
         return setImage(
             with: source,
             imageAccessor: ImagePropertyAccessor(
@@ -137,10 +136,10 @@ extension KingfisherWrapper where Base: CPListItem {
             taskAccessor: TaskPropertyAccessor(
                 setTaskIdentifier: { mutatingSelf.taskIdentifier = $0 },
                 getTaskIdentifier: { mutatingSelf.taskIdentifier },
-                setTask: { mutatingSelf.imageTask = $0 }
+                setTask: { mutatingSelf.imageTask = $0 },
+                getCancellationToken: { mutatingSelf.cancellationToken },
+                setCancellationToken: { mutatingSelf.cancellationToken = $0 }
             ),
-            previousCancellationToken: previousToken,
-            setCancellationToken: { mutatingSelf.cancellationToken = $0 },
             placeholder: placeholder,
             parsedOptions: parsedOptions,
             progressBlock: progressBlock,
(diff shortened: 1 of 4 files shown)


onevcat:
Amazing work! Thank you!

PR #2532, opened 2026-06-09
Add `MemoryStorage.totalCacheCost()` for in-memory cost reporting

onevcat:
## Summary

Adds `MemoryStorage.Backend.totalCacheCost()`, which reports the total estimated in-memory `cacheCost` of the memory cache. This gives users tangible data to reason about an appropriate `totalCostLimit`, and provides parity with `DiskStorage.totalSize`.

This builds on top of @aronspringfield's work in #2530 — both of their commits are preserved here. Thanks for the contribution! 🙏

## Background

In #2530 review we landed on the correct semantics: `totalCacheCost()` should report the *total* cost of resident objects, **including expired-but-not-yet-evicted** ones, since those still occupy memory and still count against what `NSCache` manages via `totalCostLimit`.

The first iteration implemented this by adding a public `ExpirationFilter` parameter to `value(forKey:)`. This final commit keeps the same semantics but avoids the public API churn:

- `ExpirationFilter` / the new `filter:` parameter were only ever used with `.all` (the `.expired` case was dead outside its own tests), so the public `value(forKey:)` signature is left unchanged.
- It also removes a subtle side effect: with the new `extendingExpiration: .cacheTime` default, reading an item via `.all`/`.expired` would silently *resurrect* an expired object by extending its expiration.

`totalCacheCost()` now sums `cacheCost` directly from the resident storage objects under the existing key set. Expired items are included; items already evicted by the system are excluded (they no longer occupy memory).

## Files Changed

- `Sources/Cache/MemoryStorage.swift` — adds `totalCacheCost()`.
- `Tests/KingfisherTests/MemoryStorageTests.swift` — covers empty / populated / removal / expired-inclusion / evicted-exclusion cases.

## Note

`DiskStorage` uses `totalSize` (actual bytes) while this uses `totalCacheCost` to match the naming of the memory cost limit (`totalCostLimit`). `cacheCost` is an *estimated* memory size, not an exact byte count, so the two are intentionally not identical.

### DIFF
--- a/Tests/KingfisherTests/MemoryStorageTests.swift
+++ b/Tests/KingfisherTests/MemoryStorageTests.swift
@@ -232,14 +232,14 @@ class MemoryStorageTests: XCTestCase {
         waitForExpectations(timeout: 3, handler: nil)
     }
 
-    func testTotalCacheCostIgnoresExpiredValues() {
+    func testTotalCacheCostIncludesExpiredValues() {
         let exp = expectation(description: #function)
 
         storage.store(value: 1, forKey: "1", expiration: .seconds(0.1))
         storage.store(value: 2, forKey: "2")
 
         delay(0.2) {
-            XCTAssertEqual(self.storage.totalCacheCost(), 1)
+            XCTAssertEqual(self.storage.totalCacheCost(), 2)
             XCTAssertNotNil(self.storage.storage.object(forKey: "1"))
             exp.fulfill()
         }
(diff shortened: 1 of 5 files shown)


Tests/KingfisherTests/MemoryStorageTests.swift
```diff
@@ -241,6 +271,19 @@ class MemoryStorageTests: XCTestCase {
...
+        let storage = MemoryStorage.Backend<Int>(config: config)
+
+        // Storing three items with a `totalCostLimit` of 2 forces the system to evict at least one of them.
+        storage.store(value: 1, forKey: "1")
+        storage.store(value: 1, forKey: "2")
+        storage.store(value: 1, forKey: "3")
+
+        // Evicted items no longer occupy memory, so they must not be counted toward the total cost.
+        XCTAssertLessThanOrEqual(storage.totalCacheCost(), 2)
+    }
```
aronspringfield, line 284: Perhaps this works differently during tests, but with my understanding of NSCache, it is not guaranteed that the cache will evict an object immediately when a limit is exceeded. The limits are treated as _advisory_ constraints to the cache. 

Although if this is passing, especially in a synchronous manner, then I imagine it will always pass. That is until Apple potentially change the underlying behaviour that stays in-line with the documented expectation.

aronspringfield (approved):

aronspringfield (approved):

Tests/KingfisherTests/MemoryStorageTests.swift
onevcat: That's right! Thanks for catching it.

aronspringfield (approved):

PR #2540, opened 2026-06-29
Fix broken disk caching when the image URL path contains '@'

### LINKED ISSUE #2301: autoExtAfterHashedFileName breaks caching for some URLs
### Check List

#### What

I wanted to use `DiskStorage.config.autoExtAfterHashedFileName` in order to have a file ready for usage as `url` in
```swift
try await PHPhotoLibrary.shared().performChanges { @Sendable in
    let request = PHAssetCreationRequest.forAsset()
    request.addResource(with: .photo, fileURL: url, options: nil)
}
```
but the `autoExtAfterHashedFileName` setting just looks to fully break caching in my case. It does happen with some URLs only. Without this option, I have to add a copy of the file after disk cache is ready (because PH API wants a file path with extension).

#### Reproduce

Note: removing the precondition allows the image to display, so it's not a matter of bad URL.

```swift
import SwiftUI
import Kingfisher

@main
struct autoExtAfterHashedFileNameApp: App {
    init() {
        ImageCache.default.diskStorage.config.autoExtAfterHashedFileName = true
    }
    
    // Works
//    let url = URL(string: "https://user-images.githubusercontent.com/1812216/106253726-271ed000-6218-11eb-98e0-c9c681925770.png")!
    // Doesn't work
    let url = URL(string: "https://t.furaffinity.net/57373197@300-1720981878.jpg")!
    
    var body: some Scene {
        WindowGroup {
            KFImage(url)
                .waitForCache()
                .onSuccess { result in
                    precondition(result.cacheType != .none)
                }
        }
    }
}
```

#### Other Comment

[Add anything else here]

devzahirul:
## What

Fixes #2301 — with `DiskStorage.config.autoExtAfterHashedFileName = true`, disk caching silently breaks for image URLs whose path contains an `@`.

## Root cause

When `autoExtAfterHashedFileName` is on, the cache file name is `<sha256>.<ext>`, where `<ext>` comes from `String.kf.ext` (`String+SHA256.swift`):

```swift
var ext: String? {
    guard let firstSeg = base.split(separator: "@").first else { return nil }
    var ext = ""
    if let index = firstSeg.lastIndex(of: ".") {
        let extRange = firstSeg.index(index, offsetBy: 1)..<firstSeg.endIndex
        ext = String(firstSeg[extRange])
    }
    return ext.count > 0 ? ext : nil
}
```

The `@` split exists to drop the processor identifier — processed cache keys are formed as `originalKey@processorIdentifier` (`ImageCache.swift:1470`). But when the **URL path itself contains `@`** — e.g. `https://t.furaffinity.net/57373197@300-1720981878.jpg` — `split(separator: "@").first` returns `https://t.furaffinity.net/57373197`, and the last `.` is the one in `.net`, so the extracted extension becomes **`net/57373197`**.

That value contains a path separator. `cacheFileName` then returns `<hash>.net/57373197`, and `cacheFileURL` (`DiskStorage.swift:378-380`) resolves it into a nonexistent `<hash>.net/` sub-directory. `FileManager` cannot create the file there, so the write fails, every subsequent lookup misses, and `result.cacheType` is always `.none` (the reporter's `precondition(result.cacheType != .none)` trips).

## Fix

Validate the extracted extension before using it: accept only a non-empty, short, alphanumeric string; otherwise return `nil` (no extension), which restores correct caching. Behavior for normal and processed keys is unchanged.

```swift
guard !ext.isEmpty,
      ext.count <= 20,
      ext.allSatisfy({ $0.isLetter || $0.isNumber })
else { return nil }
return ext
```

## Tests

`StringExtensionTests`:
- `testStringExt` — basic extraction, processor-identifier stripping (`image.png@round-corner` → `png`), and the no-extension cases.
- `testStringExtRejectsValueWithPathSeparator` — regression for #2301: `…/57373197@300-1720981878.jpg` must not yield an extension containing `/`.

## Verification (macOS)

```bash
  -destination 'platform=macOS' -only-testing:KingfisherTests/StringExtensionTests
```

- **Before the fix:** `testStringExtRejectsValueWithPathSeparator` fails — `XCTAssertNil failed: "net/57373197"`.
- **After the fix:** all `StringExtensionTests` pass.

## Note / scope

Fully *recovering* the real extension for URLs whose path contains `@` is ambiguous (the same `@` also separates the appended processor identifier), so this change deliberately scopes to the actual defect: never emit an extension that corrupts the cache path. For such URLs the cache file simply has no extension (the same outcome as `autoExtAfterHashedFileName = false`), and caching works again.

## CI

### DIFF
--- a/Sources/Utility/String+SHA256.swift
+++ b/Sources/Utility/String+SHA256.swift
@@ -54,6 +54,22 @@ extension KingfisherWrapper where Base == String {
             let extRange = firstSeg.index(index, offsetBy: 1)..<firstSeg.endIndex
             ext = String(firstSeg[extRange])
         }
-        return ext.count > 0 ? ext : nil
+
+        // The result is appended to the hashed cache file name as a file extension
+        // (e.g. `<hash>.png`). The text after the last `.` is not guaranteed to be a valid
+        // extension: when the original URL path itself contains `@` (such as
+        // `.../57373197@300-1720981878.jpg`), the `@`-split above yields `.../57373197`, and
+        // this extracts `net/57373197`. A value containing a path separator turns the cache
+        // file name into a nonexistent sub-directory path, so the file can never be written
+        // and disk caching silently breaks. See https://github.com/onevcat/Kingfisher/issues/2301
+        //
+        // Only accept a plausible file extension: non-empty, short, and alphanumeric.
+        guard !ext.isEmpty,
+              ext.count <= 20,
+              ext.allSatisfy({ $0.isLetter || $0.isNumber })
+        else {
+            return nil
+        }
+        return ext
     }
 }
--- a/Tests/KingfisherTests/StringExtensionTests.swift
+++ b/Tests/KingfisherTests/StringExtensionTests.swift
@@ -14,4 +14,31 @@ class StringExtensionTests: XCTestCase {
         let s = "hello"
         XCTAssertEqual(s.kf.sha256, "2cf24dba5fb0a30e26e83b2ac5b9e29e1b161e5c1fa7425e73043362938b9824")
     }
+
+    func testStringExt() {
+        // Basic extension extraction from a key.
+        XCTAssertEqual("https://example.com/image.png".kf.ext, "png")
+        XCTAssertEqual("https://example.com/image.JPG".kf.ext, "JPG")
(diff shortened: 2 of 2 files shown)


onevcat:
@devzahirul Thank you for the fix and the clear investigation on this! Sorry I forgot to leave a note when merging — this is already in master and will be included in the next release. Appreciate your contribution! 🎉

PR #2541, opened 2026-06-29
Fix data race on RetrievingContext.propagationErrors

devzahirul:
## What

Fixes a data race on `RetrievingContext` (the per-request retrieval state in `KingfisherManager`).

`RetrievingContext` is `@unchecked Sendable` and deliberately guards its mutable `_options` with a serial `propertyQueue`:

```swift
private var _options: KingfisherParsedOptionsInfo
var options: KingfisherParsedOptionsInfo {
    get { propertyQueue.sync { _options } }
    set { propertyQueue.sync { _options = newValue } }
}
```

But `propagationErrors` was left as a plain, unsynchronized `var`:

```swift
var propagationErrors: [PropagationError] = []

@discardableResult
func appendError(_ error: KingfisherError, to source: Source) -> [PropagationError] {
    let item = PropagationError(source: source, error: error)
    propagationErrors.append(item)   // ← no synchronization
    return propagationErrors
}
```

`appendError(_:to:)` is called from `failCurrentSource` (`KingfisherManager.swift:359-390`), which runs inside `@Sendable` download completion handlers — i.e. on the downloader's callback queues. Across alternative-source failovers, successive handlers can run on different threads, and `failCurrentSource` also *reads* `propagationErrors`. Concurrent append/read on a Swift `Array` with no synchronization is a data race: Thread Sanitizer reports a **`Swift access race`**, and appends can be lost or crash.

## Fix

Route `propagationErrors` through the existing `propertyQueue`, exactly as `_options` already is. Also make `popAlternativeSource()` a single atomic read-modify-write under the queue — its previous "get via `options`, mutate, set via `options`" was synchronized per access but **not atomic**, so two concurrent failovers could pop the same alternative source.

```swift
private var _propagationErrors: [PropagationError] = []
var propagationErrors: [PropagationError] { propertyQueue.sync { _propagationErrors } }

func popAlternativeSource() -> Source? {
    propertyQueue.sync {
        guard var alternativeSources = _options.alternativeSources, !alternativeSources.isEmpty else { return nil }
        let nextSource = alternativeSources.removeFirst()
        _options.alternativeSources = alternativeSources
        return nextSource
    }
}

@discardableResult
func appendError(_ error: KingfisherError, to source: Source) -> [PropagationError] {
    let item = PropagationError(source: source, error: error)
    return propertyQueue.sync {
        _propagationErrors.append(item)
        return _propagationErrors
    }
}
```

No public API or behavior change — only synchronization. (Accessing `_options` directly inside `popAlternativeSource` keeps the whole RMW in one `propertyQueue.sync` and avoids a re-entrant deadlock.)

## Tests

`KingfisherManagerTests.testContextAppendErrorIsThreadSafe` hammers `appendError` from 8 concurrent workers (250 each) while reading `propagationErrors`, then asserts all 2000 appends were recorded.

## Verification (macOS, Thread Sanitizer)

```bash
  -only-testing:KingfisherTests/KingfisherManagerTests/testContextAppendErrorIsThreadSafe \
  -enableThreadSanitizer YES
```

- **Before the fix:** `WARNING: ThreadSanitizer: Swift access race` → `** TEST FAILED **`.
- **After the fix:** passes with no sanitizer report; the existing `testContextRemovingAlternativeSource` still passes too.

## Why it matters

This is the same class of `@unchecked Sendable` hardening as the recent `fix-network-observer-cancel-race` / `fix-alternative-source-task-update-order` work, on the alternative-source failover path. The class already declares thread-safety intent via `propertyQueue`; this closes the two spots that were left out.

## CI

### DIFF
--- a/Sources/General/KingfisherManager.swift
+++ b/Sources/General/KingfisherManager.swift
@@ -364,8 +364,7 @@ public class KingfisherManager: @unchecked Sendable {
             }
             // When low data mode constrained error, retry with the low data mode source instead of use alternative on fly.
             guard !error.isLowDataModeConstrained else {
-                if let source = retrievingContext.options.lowDataModeSource {
-                    retrievingContext.options.lowDataModeSource = nil
+                if let source = retrievingContext.takeLowDataModeSource() {
                     startNewRetrieveTask(with: source, retryContext: retryContext, downloadTaskUpdated: downloadTaskUpdated)
                 } else {
                     // This should not happen.
@@ -1240,6 +1239,18 @@ class RetrievingContext<SourceType>: @unchecked Sendable {
             let nextSource = alternativeSources.removeFirst()
             _options.alternativeSources = alternativeSources
             return nextSource
+        }
+    }
+
+    func takeLowDataModeSource() -> Source? {
+        // Read and clear `lowDataModeSource` in a single critical section. Clearing it through the
+        // `options` accessor would be a get-modify-set of the whole options value, which can write
+        // back a stale snapshot and restore an alternative source that a concurrent
+        // `popAlternativeSource()` already consumed.
+        propertyQueue.sync {
+            guard let source = _options.lowDataModeSource else { return nil }
+            _options.lowDataModeSource = nil
+            return source
         }
     }
 
--- a/Sources/General/KingfisherManager.swift
+++ b/Sources/General/KingfisherManager.swift
@@ -1218,7 +1218,11 @@ class RetrievingContext<SourceType>: @unchecked Sendable {
     }
 
     let originalSource: SourceType
-    var propagationErrors: [PropagationError] = []
+
+    private var _propagationErrors: [PropagationError] = []
+    var propagationErrors: [PropagationError] {
+        propertyQueue.sync { _propagationErrors }
+    }
 
     init(options: KingfisherParsedOptionsInfo, originalSource: SourceType) {
         self.originalSource = originalSource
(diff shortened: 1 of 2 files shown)


onevpaw:
One remaining race may still undo a concurrent pop: the low-data-mode branch clears `retrievingContext.options.lowDataModeSource` via a value-type get–modify–set. It can snapshot old alternatives, allow `popAlternativeSource()` to remove one under `propertyQueue`, then write the stale snapshot back and reintroduce that source.

A context-local `takeLowDataModeSource()` that reads and clears the value within the same `propertyQueue` critical section would close this gap.

It would also be useful to add a concurrent `popAlternativeSource()` regression test, asserting each alternative is returned exactly once and the final value is `nil`; the current concurrent coverage exercises append/read but not the newly atomic pop behavior.

— an assistant to @onevcat

onevpaw (changes requested):
As @onevcat's assistant, I found a remaining race in the low-data-mode path that can restore an alternative source already removed by another callback.

`retrievingContext.options.lowDataModeSource = nil` performs a get–modify–set on the whole `KingfisherParsedOptionsInfo` struct. If it reads options containing alternative source `A`, another callback pops `A`, and then this assignment writes its stale copy back, `A` is reintroduced.

Could this be replaced with an atomic `RetrievingContext` operation that reads and clears `_options.lowDataModeSource` within the same `propertyQueue.sync` critical section? A concurrent regression test covering low-data source extraction racing with `popAlternativeSource()` would also help ensure removed alternatives cannot return.

onevtail (dismissed):
As @onevcat's assistant, I found one remaining lost-update race in the low-data-mode path.

`retrievingContext.options.lowDataModeSource = nil` reads and writes back the entire value-type options snapshot. This can overwrite an intervening atomic `popAlternativeSource()` mutation:

1. The low-data-mode branch reads an older options snapshot.
2. Another callback pops an alternative source.
3. The low-data-mode branch writes its snapshot back with only `lowDataModeSource` cleared, restoring the already-consumed alternative source.

Could this be changed to an atomic `takeLowDataModeSource()` operation under `propertyQueue`, rather than mutating through `options`? Please also add concurrency coverage for this interleaving, alongside the concurrent alternative-source pop case.

onevtail (dismissed):
Reviewing this as @onevcat’s assistant, I found one remaining concurrency issue that needs to be addressed:

`failCurrentSource` reads `options.lowDataModeSource` and then clears it through separate synchronized accesses. Because clearing the computed property performs a read-modify-write of `options`, it can write back an older options snapshot after a concurrent `popAlternativeSource()` has removed an alternative. That can reintroduce an already-consumed source and allow it to be selected twice.

Could this be made one operation under `propertyQueue`—for example, `takeLowDataModeSource()` that reads `_options.lowDataModeSource`, clears it, and returns it within one `sync` block—and covered with a concurrent regression test involving low-data fallback and `popAlternativeSource()`?

onevtail (commented):
As onevcat's assistant, I reviewed the exact head `c992a3155c2b339281896aeec625edd8144d277f`.

The concurrency changes are consistent with the existing `propertyQueue` synchronization and address the previously identified stale-options writeback path. Focused macOS TSan coverage and the `KingfisherManagerTests` suite passed locally.

One validation note: `swift test` exits because this package has no SwiftPM test target, and `Package.swift` is unchanged by this PR. The visible CI is still pending, so this is a non-approving review comment rather than a merge signal.

onevcat:
Nice fix! @devzahirul Thank you for catching it!

Issue #2543, opened 2026-06-29
Crash in SessionDataTask.mutableData copy (Data(_mutableData) → __DataStorage.init)   on iOS 26.5 — regression surfaced by the #2524 fix


MANEG-SEON-A:
## Summary

  After upgrading to **8.10.0** (which includes #2524, "Fix SessionDataTask mutable data COW sharing", the fix for #2523), we are seeing a **new** production crash on  **iOS 26.5** in the very copy that #2524 introduced.

  #2524 changed the getter to:

  ```swift
  public var mutableData: Data {
      lock.lock()
      defer { lock.unlock() }
      return Data(_mutableData)   // SessionDataTask.swift:48
  }
  ```

  The trap in #2523 was on the **write side** (`Data.append` → `_DataStorage.ensureUniqueBufferReference`). With 8.10.0 it has **moved to the copy side**: constructing `Data(_mutableData)` now traps inside Foundation's  `__DataStorage.init(bytes:length:)`. So the same iOS-26 `SessionDataTask`  instability appears to remain, just at a different site.

  ## Environment

  - **Kingfisher**: 8.10.0
  - **OS**: iOS 26.5.0 only. No events on earlier iOS.
  - **Devices**: iPhone 14, iPhone 15 (not device-specific)
  - **Free memory at crash**: 67 MB and 370 MB on two separate events — **not memory-pressure dependent**.
  - **Usage**: standard `kf.setImage` with `DownsamplingImageProcessor`. **No  progressive decoding** option. Only customizations:   `ImageDownloader.default.downloadTimeout = 20`, `ImageCache` size limits. No custom  `ImageDownloaderDelegate`.
  - Source: production Crashlytics.

  ## Stack trace (representative)

  ```
  Crashed: com.apple.NSURLSession-delegate
  EXC_BREAKPOINT

  0   Foundation      __DataStorage.init(bytes:length:) + 300
  2   App             closure #1 in Data.init<A>(_:)
  3   App             SessionDataTask.swift:48   closure #6 in ImageDownloader.setupSessionHandler()
  4   App             Delegate.swift:91          closure #1 in  Delegate.delegate<A>(on:block:)
  5   App             Delegate.swift:141         Delegate<>.call(_:)
  6   App             SessionDelegate.swift:218  SessionDelegate.urlSession(_:task:didCompleteWithError:)
  7   App            @objc SessionDelegate.urlSession(_:task:didCompleteWithError:)
[...]

PR #2545, opened 2026-06-29
Avoid generic Data initializer when copying mutableData (iOS 26 crash, #2543)

devzahirul:
Candidate fix for #2543.

## Root cause

`SessionDataTask.mutableData` (after #2524) returns a defensive copy via `Data(_mutableData)`:

```swift
public var mutableData: Data {
    lock.lock()
    defer { lock.unlock() }
    return Data(_mutableData)   // SessionDataTask.swift
}
```

Because `Data` is a `Sequence<UInt8>`, `Data(_mutableData)` resolves to the **generic** `Data.init<S: Sequence>(_:)` initializer — visible in the crash stack from #2543:

```
0  Foundation  __DataStorage.init(bytes:length:) + 300
2  App         closure #1 in Data.init<A>(_:)            ← generic Sequence init
3  App         SessionDataTask.swift:48                  ← mutableData getter
...
6  App         SessionDelegate.urlSession(_:task:didCompleteWithError:)
```

On iOS 26.x that generic initializer can trap (`EXC_BREAKPOINT`) inside `__DataStorage.init(bytes:length:)` while copying a `Data` that was accumulated through many `append`s. This is the new crash #2524 introduced (the report distinguishes it from the older append-side crash, which is untouched here).

## Fix

Build the standalone copy without the generic sequence initializer: allocate the destination up front with `Data(count:)` and copy the bytes in with `copyBytes(to:)`.

```swift
let count = _mutableData.count
guard count > 0 else { return Data() }
var copy = Data(count: count)
copy.withUnsafeMutableBytes { destination in
    _ = _mutableData.copyBytes(to: destination)
}
return copy
```

This goes through a different, stable path (`Data(count:)` zero-fill allocation + `memcpy`) and keeps the copy independent of `_mutableData`, preserving the copy-on-write isolation #2524 intended.

## Tests

`testSessionDataTaskMutableDataIsCorrectIndependentCopy` asserts `mutableData` returns the correct accumulated bytes and that a previously returned value is unaffected by later appends. The existing `testDownloadedDataCouldBeModified[WithTask]` (which read `mutableData` through the delegate path) still pass.

## ⚠️ Verification status — please read

I could **not reproduce the crash** — it is iOS 26.5-only and I don't have that environment. This change is a *reasoned route-around* of the generic initializer that appears in the report's stack trace; it is verified only for **functional correctness and copy-independence on macOS**, not against the actual iOS-26.5 trap.

The reporter offered to validate a candidate fix against production traffic — that confirmation should gate merge. To test, point the dependency at this branch:

```swift
.package(url: "https://github.com/devzahirul/Kingfisher.git", branch: "fix/mutabledata-ios26-copy")
```

Scoped to crash **B** (the copy site) from the report; the pre-existing append-side crash **A** is a separate issue and not addressed here.

### DIFF
--- a/Sources/Networking/SessionDataTask.swift
+++ b/Sources/Networking/SessionDataTask.swift
@@ -45,7 +45,19 @@ public class SessionDataTask: @unchecked Sendable {
     public var mutableData: Data {
         lock.lock()
         defer { lock.unlock() }
-        return Data(_mutableData)
+        // Return a standalone copy that does not share copy-on-write storage with `_mutableData`
+        // (the intent of #2524) — but build it WITHOUT `Data(_mutableData)`. Because `Data` is a
+        // `Sequence<UInt8>`, `Data(_mutableData)` resolves to the generic `Data.init<S: Sequence>(_:)`,
+        // which on iOS 26.x can trap inside `__DataStorage.init(bytes:length:)` while copying data that
+        // was accumulated through many `append`s (issue #2543). Allocating the destination up front and
+        // copying the bytes in goes through a different, stable path.
+        let count = _mutableData.count
+        guard count > 0 else { return Data() }
+        var copy = Data(count: count)
+        copy.withUnsafeMutableBytes { destination in
+            _ = _mutableData.copyBytes(to: destination)
+        }
+        return copy
     }
 
     var mutableDataCount: Int {
--- a/Tests/KingfisherTests/ImageDownloaderTests.swift
+++ b/Tests/KingfisherTests/ImageDownloaderTests.swift
@@ -125,6 +125,30 @@ class ImageDownloaderTests: XCTestCase {
         waitForExpectations(timeout: 5, handler: nil)
     }
     
+    // Verifies the `mutableData` getter (issue #2543): it returns the accumulated bytes correctly and
+    // as an independent copy — appending more data afterward must not mutate a previously returned
+    // value (the copy-on-write isolation introduced by #2524) — without using the generic `Data(_:)`
+    // sequence initializer that traps on iOS 26.x.
+    func testSessionDataTaskMutableDataIsCorrectIndependentCopy() {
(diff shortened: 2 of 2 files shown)


onevcat:
Thank you for taking the time to put a candidate fix together, and especially for being upfront about the verification status — that honesty made this much easier to evaluate.

After investigating #2543 I have to decline this one, because the evidence shows the root cause is elsewhere. I reproduced the crash locally on macOS 26.5 (same-generation Foundation as iOS 26.5) by fault-injecting the allocator (DYLD-interposed `malloc`/`calloc` failing large requests) during a real `ImageDownloader` download:

- Current code (`Data(_mutableData)`): traps at `__DataStorage.init(bytes:length:) + 300` — exactly the frame in #2543, including the offset.
- **This PR's replacement (`Data(count:)` + `copyBytes`): traps identically under the same condition**, just at the sibling symbol `__DataStorage.init(length:) + 364`.

The trap is the allocation-failure path (`brk #1` after a failed `malloc`/`calloc`) that all `__DataStorage` allocation routes share — it isn't specific to the generic `Sequence` initializer. Any construct that allocates a full-size buffer will crash the same way when the device can't satisfy the allocation; this change would move the crash signature rather than remove it. As a side note, `Data(count:)` also zero-fills and, in my measurements, rounds the capacity up (a 200 MiB request allocated 250 MiB), so peak memory behavior would actually get slightly worse.

The direction that does remove the crash site is to not allocate at completion at all: #2547 makes the internal completion and progressive paths use a zero-copy COW share (safe, since no append can occur after completion), while keeping the public `mutableData` snapshot semantics. The underlying condition in #2543 is OOM-class — see my comment there for the full analysis.

Closing in favor of #2547. Thanks again for the contribution and for scoping it carefully to crash B — that framing was useful even though the mechanism turned out to be allocation failure.

PR #2550, opened 2026-07-08
Avoid empty KFImage layout before image loads

### LINKED ISSUE #2533: Image view takes too much space before image has been loaded
SwiftUI Image view takes too much space if it's loading or having error. I want to have dynamic size for image. So it will be as big as loaded picture or as big as fits with aspectRatio fit. For loading and error I set max height. But view itself is still bigger and on screenshots can see white empty space.

Environment:
Kingfisher 8.10.0
Xcode  26.5 (17F42)
Swift 6.2
SwiftUI
iOS 26

Code:
```Swift
import Kingfisher
import SwiftUI

struct LazyImageWithKingFisher: View {
	let imageUrl: String

	var body: some View {
		KFImage(URL(string: imageUrl))
			.resizable()
			.placeholder {
				Color.gray
					.frame(maxHeight: 200)
					.border(.blue)
			}
			.onFailureView {
				Color.red
					.frame(maxHeight: 200)
					.border(.blue)
			}
			.clipShape(
				RoundedRectangle(cornerRadius: 12)
			)
			.aspectRatio(contentMode: .fit)
			.border(.red)
	}
}

#Preview {
	VStack(spacing: 0) {
		LazyImageWithKingFisher(
			imageUrl: "https://images.unsplash.com/photo-1506744038136-46273834b3fb"
		)
		LazyImageWithKingFisher(
			imageUrl: "https://httpbin.org/delay/10"
		)
		Spacer()
	}
}
```
Screenshots:
Image
Image

wodud1107:
## Summary

Fixes #2533.

`KFImageRenderer` now skips the image branch until `loadedImage` is available.

With the sample from #2533, `KFImage` can show a `placeholder` or `onFailureView` while an empty image branch is still in the view hierarchy. That image is invisible, but SwiftUI still includes it when it calculates layout. This can make the outer `KFImage` taller than the placeholder or failure view.

This lets the placeholder or failure view define the layout until Kingfisher has an image to render.

## Reproduction

I used the sample from #2533:

```swift
import Kingfisher
import SwiftUI

struct LazyImageWithKingFisher: View {
    let imageUrl: String

    var body: some View {
        KFImage(URL(string: imageUrl))
            .resizable()
            .placeholder {
                Color.gray
                    .frame(maxHeight: 200)
                    .border(.blue)
            }
            .onFailureView {
                Color.red
                    .frame(maxHeight: 200)
                    .border(.blue)
            }
            .clipShape(
                RoundedRectangle(cornerRadius: 12)
            )
            .aspectRatio(contentMode: .fit)
            .border(.red)
    }
}

#Preview {
    VStack(spacing: 0) {
        LazyImageWithKingFisher(
            imageUrl: "https://images.unsplash.com/photo-1506744038136-46273834b3fb"
        )
        LazyImageWithKingFisher(
            imageUrl: "https://httpbin.org/delay/10"
        )
        Spacer()
    }
}
```

Before:

Before

After:

After

## Verification

I verified the sample from #2533 manually in Xcode Preview.

I also added regression tests for the `placeholder` and `onFailureView` layout paths:

- `KFImageRendererTests.testPlaceholderDefinesLayoutWhenLoadedImageIsNil`
- `KFImageRendererTests.testFailureViewDefinesLayoutWhenLoadedImageIsNil`

I verified that these tests fail before the renderer change and pass after it.

I checked:

- Loading state with `placeholder`
- Failure state with `onFailureView`
- Successful image loading
- The same sample without adding `.frame(maxHeight:)` to the whole `KFImage`

### DIFF
--- a/Kingfisher.xcodeproj/project.pbxproj
+++ b/Kingfisher.xcodeproj/project.pbxproj
@@ -29,6 +29,7 @@
 		38D5D4A12C5C757E00BF1D01 /* ImageDataProviderCancellationTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 38D5D4A22C5C757E00BF1D01 /* ImageDataProviderCancellationTests.swift */; };
 		3ADE9AF92A73CD69009A86CA /* String+SHA256.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3ADE9AF82A73CD69009A86CA /* String+SHA256.swift */; };
 		4B0C35B22FFF5F4C00BBDEA1 /* KFImageRendererTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4B0C35B12FFF5F4C00BBDEA1 /* KFImageRendererTests.swift */; };
+		4B0C57F03008A31100BBDEA1 /* ImageBinderTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4B0C57EF3008A30A00BBDEA1 /* ImageBinderTests.swift */; };
 		4B10480D216F157000300C61 /* ImageDataProcessor.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4B10480C216F157000300C61 /* ImageDataProcessor.swift */; };
 		4B46CC5F217449C600D90C4A /* MemoryStorage.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4B46CC5E217449C600D90C4A /* MemoryStorage.swift */; };
 		4B46CC64217449E000D90C4A /* Storage.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4B46CC63217449E000D90C4A /* Storage.swift */; };
@@ -192,6 +193,7 @@
 		38D5D4A22C5C757E00BF1D01 /* ImageDataProviderCancellationTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ImageDataProviderCancellationTests.swift; sourceTree = "<group>"; };
 		3ADE9AF82A73CD69009A86CA /* String+SHA256.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = "String+SHA256.swift"; sourceTree = "<group>"; };
 		4B0C35B12FFF5F4C00BBDEA1 /* KFImageRendererTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = KFImageRendererTests.swift; sourceTree = "<group>"; };
+		4B0C57EF3008A30A00BBDEA1 /* ImageBinderTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ImageBinderTests.swift; sourceTree = "<group>"; };
 		4B10480C216F157000300C61 /* ImageDataProcessor.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ImageDataProcessor.swift; sourceTree = "<group>"; };
 		4B164ACE1B8D554200768EC6 /* CFNetwork.framework */ = {isa = PBXFileReference; lastKnownFileType = wrapper.framework; name = CFNetwork.framework; path = System/Library/Frameworks/CFNetwork.framework; sourceTree = SDKROOT; };
 		4B3E714D1B01FEB200F5AAED /* WatchKit.framework */ = {isa = PBXFileReference; lastKnownFileType = wrapper.framework; name = WatchKit.framework; path = System/Library/Frameworks/WatchKit.framework; sourceTree = SDKROOT; };
@@ -535,6 +537,7 @@
 				D12E0C491C47F23500AC98AD /* Info.plist */,
 				D12E0C441C47F23500AC98AD /* dancing-banana.gif */,
 				D1D2C3291C70A3230018F2F9 /* single-frame.gif */,
+				4B0C57EF3008A30A00BBDEA1 /* ImageBinderTests.swift */,
 				D12E0C451C47F23500AC98AD /* ImageCacheTests.swift */,
 				D1DC4B401D60996D00DFDFAA /* StringExtensionTests.swift */,
 				D12E0C461C47F23500AC98AD /* ImageDownloaderTests.swift */,
@@ -1022,6 +1025,7 @@
 				D16FEA4D23078C63006E67D5 /* LSMatcher.m in Sources */,
 				4B8351C8217066580081EED8 /* StubHelpers.swift in Sources */,
 				D1F1F6FF24625EC600910725 /* RetryStrategyTests.swift in Sources */,
+				4B0C57F03008A31100BBDEA1 /* ImageBinderTests.swift in Sources */,
 				D1DC4B411D60996D00DFDFAA /* StringExtensionTests.swift in Sources */,
 				D1BFED95222ACC6B009330C8 /* ImageProcessorTests.swift in Sources */,
 				D16FEA5223078C63006E67D5 /* LSHTTPRequestDSLRepresentation.m in Sources */,
--- a/Kingfisher.xcodeproj/project.pbxproj
+++ b/Kingfisher.xcodeproj/project.pbxproj
@@ -17,18 +17,18 @@
 		2F6023D252BE169414B01D8B /* CancellationToken.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1FAE088D96F2BF3FACCBD355 /* CancellationToken.swift */; };
 		388F37382B4D9CDB0089705C /* DisplayLink.swift in Sources */ = {isa = PBXBuildFile; fileRef = 388F37372B4D9CDB0089705C /* DisplayLink.swift */; };
 		38D5D3A32C5C757E00BF1D01 /* PixelFormatDecodingTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 38D5D3A42C5C757E00BF1D01 /* PixelFormatDecodingTests.swift */; };
-		38D5D4A12C5C757E00BF1D01 /* ImageDataProviderCancellationTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 38D5D4A22C5C757E00BF1D01 /* ImageDataProviderCancellationTests.swift */; };
 		38D5D3C12C5C7A1800BF1D01 /* gradient-8b-srgb-opaque.png in Resources */ = {isa = PBXBuildFile; fileRef = 38D5D3AC2C5C784700BF1D01 /* gradient-8b-srgb-opaque.png */; };
 		38D5D3C22C5C7A1800BF1D01 /* gradient-8b-srgb-alpha.png in Resources */ = {isa = PBXBuildFile; fileRef = 38D5D3AD2C5C784700BF1D01 /* gradient-8b-srgb-alpha.png */; };
 		38D5D3C32C5C7A1800BF1D01 /* gradient-8b-displayp3-alpha.png in Resources */ = {isa = PBXBuildFile; fileRef = 38D5D3AE2C5C784700BF1D01 /* gradient-8b-displayp3-alpha.png */; };
 		38D5D3C42C5C7A1800BF1D01 /* gradient-8b-gray.png in Resources */ = {isa = PBXBuildFile; fileRef = 38D5D3AF2C5C784700BF1D01 /* gradient-8b-gray.png */; };
-		6F62C780F5433101ADA3A9E4 /* gradient-8b-indexed.png in Resources */ = {isa = PBXBuildFile; fileRef = 5B961EF75D5B87CED8EF5951 /* gradient-8b-indexed.png */; };
 		38D5D3C52C5C7A1800BF1D01 /* gradient-10b-srgb-opaque.heic in Resources */ = {isa = PBXBuildFile; fileRef = 38D5D3B02C5C784700BF1D01 /* gradient-10b-srgb-opaque.heic */; };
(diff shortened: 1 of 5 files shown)


onevclaw:
Thanks for the fix. I checked the renderer change against #2533, and the direction looks sound: skipping the image branch while `loadedImage` is `nil` lets `placeholder` / `onFailureView` own the SwiftUI layout before load or after failure.

I also ran the focused renderer tests locally and they passed:

```sh
xcodebuild test -project Kingfisher.xcodeproj -scheme Kingfisher -destination 'id=F446A7EB-334C-428F-AC38-8B1791B33299' -only-testing:KingfisherTests/KFImageRendererTests
```

One concern before merge: this changes the lifecycle of the image branch. For `.fade(duration:)`, `ImageBinder` currently calls `markLoaded(sendChangeEvent: true)` inside `withAnimation` before assigning `loadedImage`. Previously the image branch already existed with opacity `0` and could animate to `1`; with this PR, the real image branch may be inserted only after `loaded` is already `true`, so the fade could become an immediate insertion. `loadTransition(...)` is probably less exposed, but it is also worth verifying because it relies on the transaction around `loaded` state and `loadedImage` assignment.

Could you please add a small regression test or manual confirmation for `.fade(duration:)`, and ideally `loadTransition(...)`, on a non-cached success path? Another possible direction is to make the real `loadedImage` enter the view tree while `loaded` is still `false`, then flip `loaded` to `true` inside the animation transaction.

Minor: `git diff --check` reports trailing whitespace in `Tests/KingfisherTests/KFImageRendererTests.swift` around line 80.

— onevclaw, acting on behalf of @onevcat

wodud1107:
Thanks for pointing this out.

I also removed the trailing whitespace from `KFImageRendererTests.swift` and confirmed that `git diff --check` is clean.

With the renderer change alone, `loadedImage` and `loaded` were updated in the same render pass. That meant the image could be inserted in its final state, so `.fade` and `loadTransition` did not animate.

I now set `loadedImage` first, then schedule `markLoaded(sendChangeEvent: true)` on the next main-queue task. This gives SwiftUI a chance to render the initial state before starting the transition.

I checked a non-cached successful load with `forceRefresh()`. The first GIF is with the renderer change only, and the second is with the binder update. Both `.fade(duration: 2)` and `.loadTransition(.opacity, animation: .linear(duration: 2))` now animate as expected.

| Case | Before: renderer change only | After: renderer and binder update |
| --- | --- | --- |
| `.fade(duration:)` | before_fade | after_fade |
| `.loadTransition(...)` | before_loadTransition | after_loadTransition |

<details>
<summary>Verification code</summary>

```swift
import Kingfisher
import SwiftUI

struct TransitionVerificationView: View {
    enum Mode {
        case fade
        case loadTransition
    }

    let mode: Mode
    @State private var reloadID = UUID()

    private let imageURL = URL(
        string: "https://images.unsplash.com/photo-1506744038136-46273834b3fb?auto=format&fit=crop&w=1200&q=80"
    )!

    var body: some View {
        VStack(spacing: 16) {
            image
                .id(reloadID)

            Button("Reload") {
                reloadID = UUID()
            }
        }
        .padding()
    }

    @ViewBuilder
    private var image: some View {
        switch mode {
        case .fade:
            KFImage(imageURL)
                .resizable()
                .placeholder {
                    Color.gray
                }
                .forceRefresh()
                .fade(duration: 2)
                .aspectRatio(contentMode: .fit)
                .frame(height: 240)

        case .loadTransition:
            KFImage(imageURL)
                .resizable()
                .placeholder {
                    Color.gray
                }
                .forceRefresh()
                .loadTransition(.opacity, animation: .linear(duration: 2))
                .aspectRatio(contentMode: .fit)
                .frame(height: 240)
        }
    }
}

#Preview("Fade") {
    TransitionVerificationView(mode: .fade)
}

#Preview("Load transition") {
    TransitionVerificationView(mode: .loadTransition)
}
```

</details>

onevcat:
Thanks for the PR! The approach and the tests are solid, and I verified the main animation paths on a simulator with frame-by-frame analysis of screen recordings — the results there are good:

- `.fade(duration:)`: smooth, full-duration interpolation (~60 frames over 1s), no blank-frame artifacts between placeholder and image.
- `loadTransition` (fade / slide / blur): all animate correctly, and the slide-in over the placeholder is frame-perfect.
- Progressive JPEG and the placeholder/failure layout fix for #2533 all behave as expected.

However, I found one regression that I think needs to be addressed before merging:

## Regression: external SwiftUI `.transition` on `KFImage` no longer animates

Repro (demo app, `TransitionViewDemo`, image already in cache):

```swift
withAnimation { showDetails.toggle() }
// ...
if showDetails {
    KFImage(url).transition(.slide)
}
```

- **master**: the image slides in from the leading edge as expected (frame analysis shows a continuous ~10-frame sweep).
- **this PR**: the image pops in at its final position in a single frame. The button label's cross-fade in the same recording still animates, so the `withAnimation` transaction itself runs — only the KFImage transition is lost.

### Videos

Side-by-side, 4× slow motion — **left: this PR (image pops in) / right: master (image slides in)**:

https://github.com/user-attachments/assets/efbfc206-e761-4f75-8938-2f8a42bdef7d

Full-speed originals — master (expected):

https://github.com/user-attachments/assets/2cfde0c1-f132-4133-87c5-95e583169732

this PR (regression):

https://github.com/user-attachments/assets/0b0f13be-f02d-46a3-9799-279e7537b0d7

## Why this happens

On master, the image branch (`renderedImage().opacity(...)`) exists in the view tree from the moment `KFImage` is inserted. The external `.transition` attaches to that insertion, and when the (cached) image arrives a beat later it is only a content update — the in-flight transition keeps going and the image rides it.

With this PR, at insertion time `loadedImage == nil`, so the tree contains only the placeholder branch (`Color.clear` here). When the cache hit lands, the renderer performs a structural branch swap (placeholder branch out, `renderedImage()` in) in an update that is not part of the original animated transaction — and the external transition is skipped entirely. Any usage that relies on the KFImage container keeping a stable view identity across the load is affected; a fast cache hit makes it the most visible.

## Suggested direction

Keep the view structure isomorphic before/after loading and make the empty image branch layout-neutral by value instead of removing it structurally, e.g.:

```swift
renderedImage()
    .opacity(binder.loaded ? 1.0 : 0.0)
    .frame(width: binder.loadedImage == nil ? 0 : nil,
           height: binder.loadedImage == nil ? 0 : nil)
```

Changing modifier values doesn't change view identity, so external transitions survive, while the empty image still stops participating in layout (preserving your fix for #2533). It would also be great to cover this case with a regression test similar to the ones you added — the combination "external `.transition` + cache hit" is exactly the blind spot here.

Happy to discuss alternatives if the zero-frame approach runs into edge cases (e.g. with `aspectRatio`).

wodud1107:
Thanks for the detailed review and the suggested direction! It helped me reproduce the cache-hit insertion issue in the demo app and narrow down the fix.

The renderer now keeps the image branch in the hierarchy, but gives it a zero layout size until an image is available. That keeps the external transition working without affecting the placeholder or failure-view layout.

I also updated the failure layout test. It could previously measure a transient state, so it now waits for the failure callback before taking the measurement.

As an additional local check, I ran the focused layout tests with `.aspectRatio(contentMode: .fill)` in place of `.fit`. With the zero-size branch, both the placeholder and failure views still settled at 200 pt.

I also verified the cache-hit path in the demo app after the update. The external .transition(.slide) now animates as expected:

https://github.com/user-attachments/assets/663239f0-3579-4d27-8c7d-49ecc7a9a4f9

onevclaw (commented):
As @onevcat's assistant, I found no blocking issues in `bd861614a2cd27216e30a8383e44584715a0d29e`.

The zero-sized retained image correctly keeps `KFImage`'s structure available for external SwiftUI transitions while allowing the placeholder or failure view to determine layout. The split `loadedImage`/`loaded` updates also restore the initial state needed for `.fade` and `loadTransition` animations.

The focused renderer tests and broader platform test/build coverage passed locally. I would leave approval to a follow-up review once the currently pending CI checks complete.

onevclaw (changes requested):
Reviewing this on behalf of @onevcat: the zero-sized image branch looks like a sound way to preserve the view identity needed by external transitions while allowing placeholder and failure content to determine layout.

Before this is ready, please add regression coverage for the transition paths affected by the state sequencing. `loadedImage` is assigned before `loaded` is set on a later main-queue task; during that interval, the `loadTransition` path can have neither the fallback nor the transitioned image rendered if SwiftUI commits an update between those changes. The new tests cover placeholder and failure layout, but do not exercise a non-cached successful load with `.fade` / `.loadTransition` or a cached image with an external `KFImage.transition(.slide)`.

Please add coverage that verifies those paths complete without an empty intermediate state, and include the external-transition case that this implementation is intended to preserve. Tightening the new layout assertions around the expected maximum height would also make the layout regression tests more effective.

wodud1107:
Thanks for the follow-up. I added tests for both transition regressions.

For `.fade` and `.loadTransition`, the suite now runs non-cached loads and also checks the intermediate state where `loadedImage` is available but `loaded` is still false. The placeholder remains in the hierarchy during that state.

For external transitions, one test verifies that the image branch exists before loading. Another inserts a memory-cached `KFImage` from a parent conditional with `.transition(.slide)`.

I also tightened the placeholder and failure layout assertions to expect 200 pt.

Verified with:

```sh
xcodebuild test -project Kingfisher.xcodeproj -scheme Kingfisher -destination 'id=6BDF2DFD-86DC-4FF7-AED4-1A4909867F34' -only-testing:KingfisherTests/KFImageRendererTests
```

onevtail (changes requested):
Review from onevcat's assistant:

1. **The fade preparation frame still participates in layout.**

`loadedImage` is published before `loaded` is set. In that intermediate frame, the renderer restores the image branch's intrinsic size even though its opacity is zero, while the placeholder is still present. Since opacity does not affect layout, a `.resizable().aspectRatio(.fit)` image can change the container height from a 200pt placeholder before the fade begins—the layout jump this change is intended to avoid.

Please keep the image branch layout-neutral until the image is actually renderable, and add a non-cached `.fade` regression test that asserts the placeholder geometry during this intermediate state.

2. **`onSuccess` can run before `markLoaded` with a custom callback queue.**

The state update and success callback are dispatched independently. When `callbackQueue` is a custom background queue, the state task can enqueue `markLoaded` back onto the main queue while the separately dispatched success callback runs first. This changes the previous ordering and makes the completion state observable before the image is marked loaded.

Please serialize these operations, or otherwise explicitly guarantee that `markLoaded` happens before `onSuccess`, and cover the ordering with a custom callback-queue test.

wodud1107:
Thanks for the detailed review. I updated the renderer and binder based on the two cases you pointed out.

The image branch now stays at zero size until both `loadedImage` and `loaded` are set. That keeps the placeholder or failure view from losing its layout during the preparation step.

For `.fade` and `loadTransition`, `onSuccess` now runs after `markLoaded` in the scheduled main-queue task. The non-transition path still delivers its callback asynchronously.

I added regression tests for the prepared fade layout, the failure view with `loadTransition`, and custom callback queues for both animated paths. I also added cleanup for the named cache used by the external-transition test.

Verified with:

```sh
  -destination 'id=6BDF2DFD-86DC-4FF7-AED4-1A4909867F34' \
  -only-testing:KingfisherTests/KFImageRendererTests \
  -only-testing:KingfisherTests/ImageBinderTests

xcodebuild test -workspace Kingfisher.xcworkspace -scheme Kingfisher \
  -destination 'platform=macOS'
```

onevtail (changes requested):
Reviewing this as @onevcat's assistant, I found two changes needed before this is ready:

- `Tests/KingfisherTests/ImageBinderTests.swift:64` contains trailing whitespace, causing `git diff --check master...HEAD` to fail. Please remove it.
- The external `.transition(.slide)` regression test in `KFImageRendererTests.swift` only verifies a cache hit and the success callback. It would still pass if the image appeared directly in its final state, so it does not verify the transition behavior this PR is intended to preserve. Please add a deterministic assertion for an observable intermediate transition state (or an equivalent snapshot/inspection-based check).

Also, `swift test` currently exits with “no tests found” because the package has no test target. This appears to predate the PR, but it means the passing Xcode test runs are the relevant validation rather than SwiftPM test discovery.

wodud1107:
Thanks for pointing this out. I removed the trailing whitespace and updated the cache-hit external-transition test.

The test now checks that the image branch receives the animation transaction from the surrounding `withAnimation` when `KFImage` is inserted. With the earlier structural branch, the cache result inserted the image branch later, outside that transaction.

I used a transaction probe instead of a timed snapshot because it checks the cause of the regression without depending on a particular animation frame.

I reran the focused renderer and binder tests in Xcode.

onevtail (commented):
As @onevcat's assistant, I reviewed the current head and found no blocking issues in the SwiftUI loading and transition changes.

The prepared-versus-loaded state preserves placeholder and failure layout until rendering is ready while keeping a stable image-view identity for caller-supplied SwiftUI transitions. Deferring `markLoaded` also allows the internal `loadTransition` to animate correctly. The targeted tests cover the intermediate state, layout behavior, callback ordering, and external transaction propagation.

The exact head passed the iOS, tvOS, and macOS test suites locally, along with a watchOS build. Leaving this as a comment rather than an approval because the visible CI is still pending.

onevcat:
@wodud1107 Thank you for the excellent work on this fix, and for the patient, thorough iterations across several rounds of review — the zero-frame layout approach for #2533, the transition preservation, and the regression test coverage all landed in great shape.

Your commits are now in `master` via #2555 (which included this branch unchanged, so this PR is marked as merged with your full authorship). #2555 added one follow-up commit on top, addressing a fade scaling side effect discovered in a later review pass: the zero frame needed to be released one render pass *before* the animated `markLoaded`, otherwise SwiftUI interpolates the image frame from zero inside the fade transaction and every fade-in scales up from the center.

The fix will be part of the next release. Thanks again for the contribution! 🙏

Issue #2440, opened 2025-09-24
Crash in Release builds when using `KFImage` extension modifiers that mutate `context` (`placeholder` / `cancelOnDisappear`)


greatsk55:
Environment
-----------
- Kingfisher version: 8.5.0
- Swift version: 5.5
- Xcode version: 16.2
- Build configuration: Release (-O)

## Summary
`KFImage` extension methods that mutate `context` via a `var image = self; image.context = ...; return image` pattern crash only in Release builds (does not reproduce in Debug). The crash appears tied to optimized builds and likely stems from unsafe in-place mutation of `KFImage`/its `context`, escaping closures + `AnyView` type-erasure, or optimizer-induced undefined behavior.

## Steps to reproduce (minimal)
1. Add an extension on `KFImage` like the following and use it in a SwiftUI view:

```swift
extension KFImage {
  public func kfPlaceholder<P: View>(@ViewBuilder _ content: @escaping (Progress) -> P) -> Self {
      var image = self
      image.context.placeholder = { progress in AnyView(content(progress)) }
      return image
  }
  public func kfCancelOnDisappear(_ flag: Bool) -> Self {
      var image = self
      image.context.cancelOnDisappear = flag
      return image
  }
}
```

2. Build the app using Release configuration (Optimize for Speed) and run on device or TestFlight.
3. Use the modifiers above while loading images. The app crashes during view composition / image lifecycle.

## Expected behavior
Modifiers should safely set placeholder/failure/cancel flags without crashing in Release builds.

## Actual behavior
- Crash only in Release (optimized) builds.
- No crash in Debug builds.
- Removing the in-place mutation or delegating to the protocol-level API removes the crash.

## Observed clues
- Crash is optimizer-dependent (Release-only).
- In-place mutation of `self.context` with `var image = self` is correlated with failure.
- Escaping `ViewBuilder`-based closures and `AnyView` used as placeholder are involved.
- Stack traces vary; crash often occurs when view is created/destroyed or closure executes.

## Analysis — plausible causes
1. **Unsafe in-place mutation / copy-on-write aliasing**
   - `KFImage` is a `View` (value type). Mutating a nested `context` that may be a reference type or has COW semantics can create aliasing or use-after-free under -O optimizations.

2. **Escaping closures + type-erasure lifetime**
   - Storing an escaping closure that returns `AnyView` into internal context could violate lifetime assumptions when optimized.

3. **Thread-safety / data race**
[...]

PR #2551, opened 2026-07-10
Fix SwiftUI image context sharing

EmanTahir888:
## Summary
Fixes SwiftUI image modifiers sharing the same mutable `KFImage.Context` reference between copied `KFImage` / `KFAnimatedImage` values.

This is related to #2440, where Release builds can crash when SwiftUI modifiers mutate `context`.

## Problem
`KFImage` and `KFAnimatedImage` are SwiftUI `View` value types, but their `context` is a reference type. Modifier methods returned new view values while mutating the same underlying `Context`, so derived values could unexpectedly share state:

```swift
let image = KFImage(url)
let a = image.placeholder { Text("Loading") }
let b = image.cancelOnDisappear(true)
```

The same issue also applied to generic `KFOptionSetter` modifiers such as `forceRefresh`, `memoryCacheExpiration`, `setProcessor`, `requestModifier`, `retry`, and `imageModifier`, because they mutate through `KFImageProtocol.options`.

Callback modifiers (`onSuccess`, `onFailure`, `onProgress`) had the same sharing issue because delegates lived on the shared context.

## Fix
- Add a `copyForMutation()` hook to `KFOptionSetter`.
- Make `KFImageProtocol` copy its `Context` before any generic option mutation.
- Copy context before SwiftUI-specific context mutations such as placeholders, transitions, and disappear behavior.
- Copy delegate instances when copying `Context`, so derived values no longer overwrite/share callback state.
- Keep the default `KFOptionSetter.copyForMutation()` as identity so class-based `KF.Builder` preserves existing behavior.

## Tests
Added regression coverage for:
- SwiftUI-specific context modifiers (`placeholder`, `cancelOnDisappear`).
- Generic option modifiers (`memoryCacheExpiration`, `forceRefresh`).
- Callback modifiers (`onSuccess`, `onFailure`).

## Verification
- `xcodebuild test -scheme Kingfisher -destination platform=macOS -only-testing:KingfisherTests/KingfisherOptionsInfoTests/testKFImageContextModifiersDoNotMutateOriginalImage -only-testing:KingfisherTests/KingfisherOptionsInfoTests/testKFImageOptionModifiersDoNotMutateOriginalImage -only-testing:KingfisherTests/KingfisherOptionsInfoTests/testKFImageCallbackModifiersDoNotShareDelegateState`
- `xcodebuild build -scheme Kingfisher -configuration Release -destination platform=macOS`

### DIFF
--- a/Sources/General/KFOptionsSetter.swift
+++ b/Sources/General/KFOptionsSetter.swift
@@ -40,9 +40,17 @@ public protocol KFOptionSetter {
     var onFailureDelegate: Delegate<KingfisherError, Void> { get }
     var onSuccessDelegate: Delegate<RetrieveImageResult, Void> { get }
     var onProgressDelegate: Delegate<(Int64, Int64), Void> { get }
+    
+    func copyForMutation() -> Self
 }
 
 extension KF.Builder: KFOptionSetter { }
+
+extension KFOptionSetter {
+    public func copyForMutation() -> Self {
+        self
+    }
+}
 
 final actor KFDelegateObserver {
     static let `default` = KFDelegateObserver()
@@ -60,10 +68,11 @@ extension KFOptionSetter {
     /// - Returns: A `Self` value with changes applied.
     ///
     public func onProgress(_ block: DownloadProgressBlock?) -> Self {
-        onProgressDelegate.delegate(on: KFDelegateObserver.default) { (_, result) in
+        let result = copyForMutation()
+        result.onProgressDelegate.delegate(on: KFDelegateObserver.default) { (_, result) in
             block?(result.0, result.1)
         }
-        return self
+        return result
     }
 
     /// Sets the done block to current builder.
@@ -72,10 +81,11 @@ extension KFOptionSetter {
     /// - Returns: A `Self` with changes applied.
     ///
     public func onSuccess(_ block: ((RetrieveImageResult) -> Void)?) -> Self {
-        onSuccessDelegate.delegate(on: KFDelegateObserver.default) { (_, result) in
+        let result = copyForMutation()
+        result.onSuccessDelegate.delegate(on: KFDelegateObserver.default) { (_, result) in
             block?(result)
         }
-        return self
+        return result
     }
 
     /// Sets the catch block to current builder.
@@ -84,10 +94,11 @@ extension KFOptionSetter {
     /// - Returns: A `Self` with changes applied.
     ///
     public func onFailure(_ block: ((KingfisherError) -> Void)?) -> Self {
-        onFailureDelegate.delegate(on: KFDelegateObserver.default) { (_, error) in
+        let result = copyForMutation()
+        result.onFailureDelegate.delegate(on: KFDelegateObserver.default) { (_, error) in
             block?(error)
         }
-        return self
+        return result
     }
 }
 
@@ -103,8 +114,9 @@ extension KFOptionSetter {
     /// such as attempting to retrieve cached images and storing downloaded images within it.
     ///
     public func targetCache(_ cache: ImageCache) -> Self {
-        options.targetCache = cache
-        return self
+        let result = copyForMutation()
+        result.options.targetCache = cache
+        return result
     }
     
     /// Sets the target image cache to store the original downloaded image for this task.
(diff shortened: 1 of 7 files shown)


onevclaw:
Thanks for working on this. I checked the current diff against the context-sharing issue, and I think the fix is still incomplete.

The new copy path covers a few SwiftUI-specific modifiers, but the shared `KFOptionSetter` path still mutates `context` through `KFImageProtocol.options`:

```swift
public var options: KingfisherParsedOptionsInfo {
    get { context.options }
    nonmutating set { context.options = newValue }
}
```

All common option builders such as `memoryCacheExpiration`, `forceRefresh`, `setProcessor`, `requestModifier`, `retry`, `imageModifier`, `callbackQueue`, etc. still go through that nonmutating setter and mutate the same `Context` instance. For example:

```swift
let base = KFImage(url)
let a = base.memoryCacheExpiration(.seconds(60))
let b = base.forceRefresh(true)
```

These derived values can still affect each other because no context copy happens on that path.

I also think `onSuccess` / `onFailure` / `onProgress` still have the same issue. They mutate the delegates directly, and `Context.copy()` appears to preserve the same delegate instances, so deriving two images from a base image and attaching different callbacks can still overwrite or share callback state.

Could this PR either extend the fix to the `KFOptionSetter` and callback paths, or explicitly narrow its scope in the PR description? I would also suggest adding regression tests for at least one generic option modifier, for example `memoryCacheExpiration` or `forceRefresh`, and one callback modifier, in the same style as the new `placeholder` / `cancelOnDisappear` test.

— onevclaw, acting on behalf of @onevcat

EmanTahir888:
Thanks for the detailed review. You were right: the previous version only covered the SwiftUI-specific modifier path and missed the shared `KFOptionSetter` and callback paths.

I updated the PR to address those concerns:

- Generic `KFOptionSetter` modifiers now go through a new `copyForMutation()` hook.
- `KFImageProtocol` overrides `copyForMutation()` to copy its `Context` before option mutations, so modifiers such as `forceRefresh`, `memoryCacheExpiration`, `setProcessor`, `requestModifier`, `retry`, `imageModifier`, etc. no longer mutate the base/sibling `KFImage` context.
- Callback modifiers now also call `copyForMutation()` before setting delegates.
- `Context.copy()` now copies delegate instances instead of reusing the same delegate objects, so derived views no longer overwrite/share callback state.
- Added regression tests for:
  - SwiftUI-specific modifiers (`placeholder`, `cancelOnDisappear`)
  - Generic option modifiers (`memoryCacheExpiration`, `forceRefresh`)
  - Callback modifiers (`onSuccess`, `onFailure`)

Verification run locally:

```sh
xcodebuild test -scheme Kingfisher -destination platform=macOS \
  -only-testing:KingfisherTests/KingfisherOptionsInfoTests/testKFImageContextModifiersDoNotMutateOriginalImage \
  -only-testing:KingfisherTests/KingfisherOptionsInfoTests/testKFImageOptionModifiersDoNotMutateOriginalImage \
  -only-testing:KingfisherTests/KingfisherOptionsInfoTests/testKFImageCallbackModifiersDoNotShareDelegateState
```

```sh
xcodebuild build -scheme Kingfisher -configuration Release -destination platform=macOS
```

EmanTahir888:
Thanks for the review. I updated the PR to cover the missing paths.

The generic KFOptionSetter path now uses a copyForMutation hook, and
KFImageProtocol copies its Context before option mutations. This covers
modifiers like memoryCacheExpiration, forceRefresh, setProcessor,
requestModifier, retry, imageModifier, and similar option builders.

I also fixed the callback path. onSuccess, onFailure, and onProgress now
copy before mutation, and Context.copy() copies delegate instances instead
of sharing the same delegate objects.

I added regression tests for SwiftUI-specific modifiers, generic option
modifiers, and callback modifiers. I also ran the targeted tests and a
Release macOS build successfully.

onevcat:
@EmanTahir888 Thank you for the excellent work on this! The analysis of the shared `Context` mutation was spot on, and the `copyForMutation()` design — with the identity default keeping `KF.Builder` behavior intact — was a clean way to solve it.

Your commits were merged as part of #2554, which added a small follow-up on top (covering the deprecated `onFailureImage`, some deduplication, and extra regression tests). The fix will be included in the next release. Really appreciate the contribution! 🎉

PR #2556, opened 2026-07-17
Look up disk cache entries with a single metadata syscall

devzahirul:
### What

`DiskStorage.Backend.value(forKey:...)` used to touch the file system twice for every disk cache lookup:

1. `fileManager.fileExists(atPath:)` — a `stat` syscall
2. `FileMeta(fileURL:resourceKeys:)` (backed by `URL.resourceValues(forKeys:)`) — a `getattrlist` syscall

`URL.resourceValues(forKeys:)` already reports a nonexistent file (or a missing intermediate directory) with `NSFileReadNoSuchFileError`, so the `fileExists` gate only duplicates the disk touch. This PR reads the metadata directly and treats that error as a regular cache miss, halving the blocking syscalls of the lookup.

### Why it matters

With default options, `KingfisherManager.retrieveImage` probes cache existence **on the caller thread** — the main thread for `kf.setImage` / `KFImage` loads (`retrieveImageFromCache` → `imageCachedType` → `diskStorage.isCached` → this lookup). When the storage stack is under I/O load, each blocked syscall on that path shows up as dropped frames: #2217 reports exactly this at scale, and its trace points at the `fileExists(atPath:)` call this PR removes; #2323 profiles the same stall from `retrieveImage`. The probe also runs a second round for the original key when a non-default processor is used, so a scrolling list with e.g. downsampling pays up to 4 blocking syscalls per cell today, and 2 after this change.

The change composes with the opt-in `asyncCacheTypeCheck`: users who opt in move the remaining syscall off the caller thread entirely, while everyone on the default path now blocks half as often. The actual data loading path also drops from 3 file system touches to 2, which additionally helps `.loadDiskFileSynchronously` users.

Micro-benchmark of the probe sequence (release build, Apple Silicon, hot file cache; the relative win grows when the disk queue is contended, which is the reported scenario):

| Case | Before | After |
|---|---|---|
| Lookup of an existing entry | 3.70 µs (stat + getattrlist) | 1.11 µs (getattrlist) |
| Lookup of a missing entry that passed `maybeCached` | 1.50 µs | 2.79 µs |

The missing-entry case costs slightly more because constructing the thrown error is heavier than a plain `stat` returning false. That path is only reached for stale `maybeCached` entries (for example, the first probe after an expired file was swept); the common miss is still answered from the in-memory `maybeCached` set with no syscall at all, and the hit case dominates steady-state scrolling.

### Behavior compatibility

| Situation | Before | After |
|---|---|---|
| File missing / missing intermediate directory | `nil` (miss) | `nil` (miss), via `NSFileReadNoSuchFileError` |
| File present and valid | value | value |
| File present but expired | `nil` | `nil` |
| Metadata unreadable, file not confirmable on disk (e.g. no search permission on the folder) | `nil` | `nil`, via a cold-path `fileExists` fallback |
| Metadata unreadable on a file confirmed to exist | `invalidURLResource` | `invalidURLResource` |

The only observable difference is for foreign entries the storage never writes itself (for example, a dangling symlink manually placed in the cache folder under a hashed name): the existence probe now reports it as cached and the actual load fails, where it previously reported a miss. `store` only ever writes regular files, so this cannot arise from Kingfisher's own operation.

### Tests

`DiskStorageTests` gains seven cases pinning every branch of the changed lookup, verified at 100% line coverage of the changed function and the new error helper with `xccov`:

- File removed behind the storage (name still in `maybeCached`, so the lookup reaches the metadata reading and observes the missing file there) → miss, for both the existence probe and the value load.
- Unknown key answered by the `maybeCached` shortcut → miss without any disk touch.
- Unreadable containing directory (metadata fails, file not confirmable) → miss instead of an error.
- Unreadable metadata on a file confirmed to exist → `invalidURLResource`.
- Unreadable file content with readable metadata → the probe reports a hit, the load fails with `cannotLoadDataFromDisk`.
- Storage whose directory creation failed → `diskStorageIsNotReady`.
- Direct checks of the `isFileMissing` error classification (matching code and domain).

### How to verify

```
xcodebuild -project Kingfisher.xcodeproj -scheme Kingfisher -destination "platform=macOS" test
xcodebuild -project Kingfisher.xcodeproj -scheme Kingfisher -destination "platform=iOS Simulator,name=iPhone 17,OS=26.0" test
```

Both full suites pass locally on macOS 15 / Xcode 17.

### DIFF
--- a/Sources/Cache/DiskStorage.swift
+++ b/Sources/Cache/DiskStorage.swift
@@ -256,17 +256,12 @@ public enum DiskStorage {
                 throw KingfisherError.cacheError(reason: .diskStorageIsNotReady(cacheURL: directoryURL))
             }
 
-            let fileManager = config.fileManager
             let fileURL = cacheFileURL(forKey: key, forcedExtension: forcedExtension)
-            let filePath = fileURL.path
 
             let fileMaybeCached = maybeCachedCheckingQueue.sync {
                 return !maybeCachedSetupCompleted || (maybeCached?.contains(fileURL.lastPathComponent) ?? true)
             }
             guard fileMaybeCached else {
-                return nil
-            }
-            guard fileManager.fileExists(atPath: filePath) else {
                 return nil
             }
 
@@ -275,6 +270,17 @@ public enum DiskStorage {
                 let resourceKeys: Set<URLResourceKey> = [.contentModificationDateKey, .creationDateKey]
                 meta = try FileMeta(fileURL: fileURL, resourceKeys: resourceKeys)
             } catch {
+                // A missing file is already reported by the resource values reading, so a separate `fileExists`
+                // check before it is not needed. This runs on the caller thread when probing cache existence, so
+                // it should touch the disk as few times as possible.
+                if error.isFileMissing {
+                    return nil
+                }
+                // Other metadata failures keep the original semantics: a file that cannot be confirmed on disk
+                // is a cache miss, while an existing file with unreadable metadata is an error.
+                guard config.fileManager.fileExists(atPath: fileURL.path) else {
+                    return nil
+                }
                 throw KingfisherError.cacheError(
                     reason: .invalidURLResource(error: error, key: key, url: fileURL))
             }
@@ -670,6 +676,16 @@ extension DiskStorage {
             cacheName = "com.onevcat.Kingfisher.ImageCache.\(config.name)"
             directoryURL = config.cachePathBlock(url, cacheName)
         }
+    }
+}
+
+extension Error {
+    // `URL.resourceValues(forKeys:)` reports a nonexistent file (or a missing intermediate directory in its
+    // path) as `NSFileReadNoSuchFileError`. Treating it as a normal cache miss allows the disk lookup to skip
+    // a dedicated `fileExists` disk touch on the hot path.
+    var isFileMissing: Bool {
+        let nsError = self as NSError
+        return nsError.domain == NSCocoaErrorDomain && nsError.code == NSFileReadNoSuchFileError
     }
 }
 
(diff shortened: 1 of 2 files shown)


onevclaw (changes requested):
As @onevcat's assistant, I found a regression for cache lookups while the initial directory scan is still running or has failed.

`setupCacheChecking()` starts before `prepareDirectory()` completes. Until that scan finishes, `maybeCachedSetupCompleted` is false; after a scan failure, `maybeCached` remains unavailable. Both states now proceed directly to `URL.resourceValues`. For an unknown/missing key, absence is therefore detected by throwing `NSFileReadNoSuchFileError` instead of using the previous `fileExists` fast miss.

This matters because `ImageCache.imageCachedType` can synchronously touch disk. During cold start or when scanning is blocked, missing-key lookups take the exception-based metadata path repeatedly, increasing main-thread I/O overhead.

Please distinguish a confirmed completed-scan hit from an unknown scan state: only skip `fileExists` for the former, and retain the existence prefilter while scanning is pending or unavailable. Please also add regressions for a delayed scan and a failed scan that verify unknown keys still call `fileExists`. A dangling-symlink case would be useful as well, since the new metadata path can report it as `.disk` whereas the prior behavior treated it as a miss.

onevcat:
Thanks for the thorough investigation, benchmark, and compatibility analysis.

I re-evaluated the earlier requested change. The scan-state observation is correct, but keeping `fileExists` for an unknown scan state would trade a small error-construction cost on misses for an additional potentially blocking file-system call on hits. The reported problem is specifically an I/O-contended, hit-heavy path, so the single-metadata-call approach is a worthwhile improvement.

This is intentionally a tactical optimization rather than a complete fix for #2217/#2323. Fully removing caller-thread disk I/O requires making the probe/load path asynchronous by default, which has observable task and callback semantics and is better handled as a future major-version change.

I am going to merge this. I will follow up separately with a few non-blocking cleanups and investigate the directory-setup / initial-scan ordering. Thank you again for the careful work.

PR #2562, opened 2026-07-25
Release remaining image targets during in-flight downloads

onevcat:
## Summary
- release image-settable targets, buttons, CarPlay list items, text attachments, attributed views, and SwiftUI binders while their retrieval is in flight
- keep downloads and caching active, while forwarding terminal results after a target has been released
- add lifecycle regressions for every affected imperative path and `ImageBinder`

Follow-up to #2313 and #2561.

## Verification
- `bundle exec fastlane tests`
- targeted lifecycle regressions on iOS, macOS, and tvOS
- watchOS build

### DIFF
--- a/Sources/Extensions/CPListItem+Kingfisher.swift
+++ b/Sources/Extensions/CPListItem+Kingfisher.swift
@@ -109,11 +109,10 @@ extension KingfisherWrapper where Base: CPListItem {
         progressBlock: DownloadProgressBlock? = nil,
         completionHandler: (@MainActor @Sendable (Result<RetrieveImageResult, KingfisherError>) -> Void)? = nil) -> DownloadTask?
     {
-        var mutatingSelf = self
         return setImage(
             with: source,
             imageAccessor: ImagePropertyAccessor(
-                setImage: { image, _ in
+                setImage: { listItem, image, _ in
                     /**
                      * In iOS SDK 14.0-14.4 the image param was non-`nil`. The SDK changed in 14.5
                      * to allow `nil`. The compiler version 5.4 was introduced in this same SDK,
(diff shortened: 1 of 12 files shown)


onevclaw (changes requested):
`PHLivePhotoView.kf.setImage` still retains its target for the full request lifetime. Its `Task { @MainActor in ... }` captures `self`, and the subsequent `PHLivePhoto.request` result handler continues to access `self`/`base.livePhoto`. As a result, a `PHLivePhotoView` cannot be released while the download or PhotoKit request is pending.

Please make this path weak-target based as well, while continuing to forward the terminal completion when the target has already been released. A delayed Live Photo lifecycle regression test would also ensure this imperative image-target path is covered.

onevclaw:
All points from my earlier review have been addressed on this branch by f1c7f94 ("Release live photo views during in-flight retrieval"):

- The `Task { @MainActor in ... }` closure and the `PHLivePhoto.request` result handler now reference the view only through a `WeakBox`, so a `PHLivePhotoView` can be deallocated while the download or the PhotoKit request is still pending. The closure no longer captures `self`/`base`.
- `checkNotCurrentTask` became a static helper taking the weak-resolved view; when the view is already gone, the in-flight task is treated as current, so the terminal success/failure is still forwarded to the completion handler — consistent with the other imperative targets in this PR.
- The requested delayed lifecycle regression test is in place: `testLivePhotoViewNotRetainedByInFlightDownload` releases the view mid-download and asserts the terminal result is forwarded and both resources end up cached.

Verified locally by auditing the file for remaining strong-capture paths (none left) and running the `ImageViewExtensionTests` suite (38 tests, including the new regression) on iOS Simulator — all passed. All 24 CI build/test jobs are also green on the current head. Nothing from my review remains open; the requested changes are resolved.

PR #2568, opened 2026-08-21
Track download priority per shared-task consumer

### LINKED ISSUE #2567: `downloadPriority` is ignored when a request joins an in-flight download
When multiple requests share one in-flight download, the download priority is not handled per consumer. This surfaces in two ways.

### 1. A request joining a shared task has its `downloadPriority` silently dropped

In `ImageDownloader.addDownloadTask`, `downloadPriority` is applied only when a fresh `URLSessionDataTask` is created:

https://github.com/onevcat/Kingfisher/blob/6120d8fdca80a645076df0070640cc681abe33e4/Sources/Networking/ImageDownloader.swift#L419-L428

A request that coalesces onto an existing task via `sessionDelegate.append` (L421 above) never affects the task priority.

Typical impact: an `ImagePrefetcher` starts a download at low priority, the cell becomes visible and requests the same URL at high priority — but the download stays at low priority, and the visible image loads slowly.

### 2. `reducePriorityOnDisappear` can degrade other consumers of the shared task

The option added in #2211 is implemented as a raw write to the shared task:

https://github.com/onevcat/Kingfisher/blob/6120d8fdca80a645076df0070640cc681abe33e4/Sources/SwiftUI/ImageBinder.swift#L197-L200

Since the `URLSessionDataTask` may be shared (two `KFImage`s with the same URL, or a view plus a prefetcher), one view scrolling off-screen lowers the priority for every consumer still waiting on that URL — including a fully visible one. `restorePriorityOnAppear()` (L191) has the mirrored last-write-wins behavior.

## History

#73 (2015) reported the same "joining request never updates priority" scenario. 
It was addressed in c85a3fc46fb682b7c8640e62be1ce0b732328db3 by exposing a settable `priority` on the download task, shipped in 2.0.0.

The API survived through 4.x and was lost in the 5.0 downloader rewrite (5e2618062cd0b3da7915ee8d7b2e6a73076487b5). I couldn't find the removal mentioned in the changelog or migration notes, so it looks like an unintended casualty of the rewrite.

I'd be happy to work on a fix and submit a PR for this, if that sounds good to you.

onevtail:
## What

- Track the current download priority for each callback token in `SessionDataTask`.
- Set the underlying `URLSessionDataTask` priority to the maximum priority requested by active consumers.
- Recalculate the shared priority when a consumer joins, changes priority, or is cancelled.
- Route SwiftUI disappear/appear priority changes through the consumer token.
- Restore each SwiftUI request's original download priority instead of always restoring the default value.

## Why

Requests for the same final URL share one in-flight download task. Previously, a joining request could not raise that task's priority, and `reducePriorityOnDisappear` directly modified the shared task, potentially lowering the priority for other visible consumers.

## How tested

- Added regression coverage for a high-priority request joining an in-flight low-priority download.
- Covered lower-priority joins, cancellation fallback, and stale-token updates.
- Covered isolation between SwiftUI consumers and restoration of custom priorities.
- Extended the existing concurrent-access test with priority updates and resets.
- Ran the complete iOS test suite: 402 tests, 0 failures.

Fixes #2567

### DIFF
--- a/Sources/Networking/ImageDownloader.swift
+++ b/Sources/Networking/ImageDownloader.swift
@@ -144,6 +144,16 @@ public final class DownloadTask: @unchecked Sendable {
         guard let sessionTask, let cancelToken else { return }
         sessionTask.cancel(token: cancelToken)
     }
+
+    func setPriority(_ priority: Float) {
+        guard let sessionTask, let cancelToken else { return }
+        sessionTask.setPriority(priority, for: cancelToken)
+    }
+
+    func resetPriority() {
+        guard let sessionTask, let cancelToken else { return }
+        sessionTask.resetPriority(for: cancelToken)
+    }
     
     public var isInitialized: Bool {
         propertyQueue.sync {
@@ -423,7 +433,6 @@ open class ImageDownloader: @unchecked Sendable {
             downloadTask = existingDownloadTask
         } else {
             let sessionDataTask = session.dataTask(with: context.request)
-            sessionDataTask.priority = context.options.downloadPriority
             downloadTask = sessionDelegate.add(sessionDataTask, url: context.url, callback: callback)
         }
         return downloadTask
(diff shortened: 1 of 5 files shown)


onevpaw (approved):
Cross-review complete: I found no P0-P2 issue in this head, and the required checks passed.
This is a technical approval only; final merge remains with @onevcat.

Issue #73, opened 2015-07-01
Changing Priority of download-dataTask after it started


mheicke:
Hi,
since I prefetch a lot of images, but want to show a couple of images as fast as possible, I just looked into prioritization of Kingfisher. Looks quite good to me, besides one point: it's not possible to change the priority of a download, once it's started.

As an example: I start to download the images cat[1-100].png and dog[1-100].png. They are all priorized as low since its only prefetching. Now I open the cat view which has a tableView with cat images. So I want to set a couple of images (of the visible cells) via UIImageViews kf_setImageWithURL. Since there are 200 images currently downloading, I need a higher prioritization for the images cat1.png - cat5.png. 

But since they are part of the batch download of the prefetching, their url is already in self.fetchLoads in ImageDownloader and hence the started-closure is not called again in line 214 and "task.priority" is not changed to default, although the options dictionary in the downloadImageWithURL-call does not contain the lowPriority option.

A solution could be to attach the downloadTask to the ImageFetchLoad. In this case, it would be also possible to provide auto-priorisation (the more callbacks are attached to the ImageFetchLoad, the higher the priority could be. Some rewrite of ImageDownloader would be needed.

What do you think?

Best regards,
Matt

schickling:
This would be a huge improvement! :+1:

schickling:
Also is there a reason for just toggling between `NSURLSessionTaskPriorityLow` and `NSURLSessionTaskPriorityDefault`? The priority parameter is a `float` ranging from 0 to 1. This flexibility shouldn't be hidden.

schickling:
![screen shot 2016-01-02 at 12 25 26](https://cloud.githubusercontent.com/assets/1567498/12074319/f2f88d20-b14b-11e5-8a1f-47554fdac9ae.png)

ejoebstl:
Yes, I a agree. That feature would be very useful.

onevcat:
The original implementation hides the `Float` detail and tries to make things simpler. But I guess it is reasonable to give users more control on it by setting a value for the priority. However, changes on it might means API breaking. I will take a look to see if we could mark current `lowPriority` option as deprecated and introduce a new way to set initial priority for a download task soon.

And for changing the priority on air when downloading, the downloader APIs (as well as the image view extension APIs) are returning a image task. We could just expose the `priority` property of task object to support changing it after download task started. It will be implemented soon in next major version.

schickling:
Really looking forward to this improvement!

[...]

PR #2570, opened 2026-08-24
Handle download priority per consumer of a shared download task

### LINKED ISSUE #2567: `downloadPriority` is ignored when a request joins an in-flight download

mini-min:
### Motivation

Fixes #2567. When multiple requests share one in-flight download, the priority is not handled per consumer: a request joining a shared task has its `downloadPriority` dropped, and `reducePriorityOnDisappear` writes to the shared `URLSessionDataTask` directly, lowering the priority for every consumer still waiting on that URL. The issue has the details and the history.

### What this changes

`SessionDataTask` now tracks the priority each joined callback asks for, keyed by its cancel token, and keeps the underlying task at the maximum of those values. The value updates when a callback joins, is cancelled, or changes its own priority. So the task speeds up when a visible view joins a prefetch, and drops back once the high-priority consumers leave.

The settable `priority` that 2.0.0 exposed on the download task (c85a3fc, for #73), lost in the 5.0 downloader rewrite (5e26180), is restored on `DownloadTask`, adapted to the model where the session task can be shared: the setter changes only the priority this task asks for. A value set before the shell is linked to an actual task (async cache probe, async request modifier) is applied on linking, and re-applied when a retry links a new task.

`KFImage`'s appear/disappear handling goes through this instead of the raw writes, and `restorePriorityOnAppear` now restores the priority the view originally requested instead of the global default.

The work is split into four commits along these lines, so the API restoration and the behavior change can be reviewed or dropped separately.

### Tests

Added to `ImageDownloaderTests`, using `delayedStub` so two requests reliably coalesce onto one session task:

- a joining request raises the shared task's priority; a lower-priority joiner does not degrade it
- the priority is recalculated when a consumer cancels
- `DownloadTask.priority` changes only the caller's share, and a value set before linking is applied once linked

The full suite passes on macOS locally; other platforms are left to CI. A pre-existing `DataReceivingSideEffectTests` failure on the Xcode 27 beta reproduces on `master` and is unrelated.

`CHANGELOG.md` is untouched, since it looks like it is updated in the release commits.

### DIFF
--- a/Sources/Networking/ImageDownloader.swift
+++ b/Sources/Networking/ImageDownloader.swift
@@ -110,7 +110,7 @@ public final class DownloadTask: @unchecked Sendable {
     }
 
     private var _cancelToken: SessionDataTask.CancelToken? = nil
-    
+
     /// The cancellation token used to cancel the task.
     ///
     /// This is solely for identifying the task when it is cancelled. To cancel a ``DownloadTask``, call
@@ -118,6 +118,58 @@ public final class DownloadTask: @unchecked Sendable {
     public private(set) var cancelToken: SessionDataTask.CancelToken? {
         get { propertyQueue.sync { _cancelToken ?? _linkedTask?.cancelToken } }
         set { propertyQueue.sync { _cancelToken = newValue } }
+    }
+
+    // The last priority set through the `priority` setter.
+    // It is kept around so it can be applied when the task is linked, or re-applied when a retry links a new task.
+    private var _explicitPriority: Float? = nil
+
+    /// The relative priority of this download task.
+    ///
+    /// The value is between 0.0 and 1.0, with `URLSessionTask.defaultPriority` (0.5) as the default.
+    /// See the documentation of `URLSessionTask.priority` for more about how the value is used.
+    ///
+    /// In Kingfisher, the underlying session task may be shared by multiple `DownloadTask`s requesting the same URL.
+    /// Setting this value only changes the priority this task asks for.
+    /// The shared session task always runs at the highest priority among all `DownloadTask`s currently sharing it.
+    /// So lowering the value here never slows down another consumer of the same URL.
+    public var priority: Float {
+        get {
+            propertyQueue.sync {
+                if let linked = _linkedTask { return linked.priority }
+                if let sessionTask = _sessionTask, let token = _cancelToken,
+                   let value = sessionTask.priority(forToken: token)
+                {
+                    return value
+                }
+                return _explicitPriority ?? URLSessionTask.defaultPriority
+            }
+        }
+        set {
+            enum Target {
+                case linked(DownloadTask)
+                case session(SessionDataTask, SessionDataTask.CancelToken)
+                case notLinkedYet
+            }
+            // Decide the forwarding target under `propertyQueue`.
+            // The forwarding itself takes other locks and must happen outside of it.
+            let target: Target = propertyQueue.sync {
+                _explicitPriority = newValue
+                if let linked = _linkedTask { return .linked(linked) }
+                if let sessionTask = _sessionTask, let token = _cancelToken {
+                    return .session(sessionTask, token)
+                }
+                return .notLinkedYet
+            }
+            switch target {
+            case .linked(let task):
+                task.priority = newValue
+            case .session(let sessionTask, let token):
+                sessionTask.updatePriority(newValue, forToken: token)
+            case .notLinkedYet:
+                break // Applied in `linkToTask` once the actual task exists.
+            }
+        }
     }
 
     /// Cancel this single download task if it is running.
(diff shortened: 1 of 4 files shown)


onevtail (changes requested):
Thanks for addressing priority aggregation per consumer—the overall direction and the sequential behavior look good. However, I found a blocking concurrency issue on exact head `14a92315cdc976a627d6b21569e88bfa1063e5a4`.

In `SessionDataTask`, the aggregate priority is calculated while holding the lock, but `task.priority` is assigned after unlocking. This permits stale publication:

1. T1 updates a token to `0.75`, captures that aggregate, then pauses after unlocking.
2. T2 updates the same token to `0.25` and publishes `0.25`.
3. T1 resumes and publishes its stale `0.75`.

The stored per-token priority is then `0.25`, while the underlying task remains at `0.75`. A deterministic blocking mock reproduced this mismatch. The same pattern can affect add/cancel interleavings.

`DownloadTask.linkToTask` has an analogous race: it captures `_explicitPriority` on `propertyQueue` but forwards it afterward, allowing a concurrent setter to publish a newer value before the linking path overwrites it with the stale captured value.

Please establish one linearization mechanism for each state transition and its publication to the underlying task—for example, publishing under a proven-safe lock discipline or using a generation/version with serialized publication. Please also add deterministic concurrency tests covering update/update, update/cancel, and link/newer-setter interleavings.

The existing priority tests and full macOS suite pass locally, and the platform builds succeed, but those sequential tests do not exercise these races. Visible CI is also still pending.

mini-min:
Thanks for catching this. Fixed in eb357f2f.

`SessionDataTask` now assigns `task.priority` while still holding `lock`, through a shared `applyAggregatedPriority()` called from `addCallback` / `removeCallback` / `updatePriority`. A state change and its publication can no longer interleave. Setting `URLSessionTask.priority` doesn't call back into Kingfisher, so holding the lock across it is safe.

`DownloadTask.priority` and `linkToTask` both forward inside `propertyQueue` now, which serializes a concurrent setter against the linking step. The links only ever point towards the session task, so the nested locking can't cycle.

Added two tests. `testConcurrentPriorityUpdatesPublishConsistently` runs update/update, update/cancel and add/remove interleavings and asserts the published priority equals the stored aggregate after the threads join. `testLinkToTaskConcurrentWithPrioritySetterKeepsNewestValue` races `linkToTask` against a newer setter and asserts the newer value wins. The full macOS suite passes locally.

onevtail (commented):
I reviewed the exact head (`eb357f2f39dbe1f3dfdac8d3653ecdee776dc11a`). The previously identified stale priority publication race is resolved: state mutation and publication now share a clear synchronization point, and I found no remaining blocking correctness issue.

Local validation passed the full macOS suite, the relevant iOS tests, and builds for iOS, tvOS, watchOS, and visionOS. The per-consumer priority aggregation, cancellation recalculation, and delayed-link behavior also look correct.

I’m leaving this as a non-approving review because no remote checks are currently reported; approval should wait until the required CI completes successfully.

mini-min:
Closing this since the merged #2568 covers the same fix.

onevcat:
@mini-min Oh!
Anyway, thank you for the contribution!

PR #2571, opened 2026-08-25
Remove version guards below the Kingfisher 8 baseline

mini-min:
### Motivation

Kingfisher 8 requires iOS 13.0+ / tvOS 13.0+ / macOS 10.15+ / watchOS 6.0+ / visionOS 1.0+ and Swift 5.9+. Package.swift, Kingfisher.podspec, and the Xcode project all declare the same deployment targets. A few version checks below this baseline are still in the sources. They always take the same branch, so the other branch is dead code.

### Changes

- `String+SHA256`: the CryptoKit path is always available, so the CommonCrypto fallback and its import are removed. CommonCrypto is not used anywhere else.
- `Indicator`: picks the `.large` / `.medium` activity indicator styles directly. The `UIActivityIndicatorView.Style` compatibility extension only declared fallback constants for compilers older than 5.1, so nothing is left in it and the whole extension is removed.
- `AnimatedImageView`: reads `UITraitCollection.current.displayScale` directly instead of falling back to `UIScreen.main.scale`.
- `KingfisherError.isLowDataModeConstrained` and the low data mode request setup in `ImageDownloader`: only the `#available` clause is removed, the remaining conditions stay.
- `CPListItem`: every SDK that ships with Swift 5.4+ accepts `setImage(nil)`, so the image is set unconditionally.
- `DisplayLink`: the `swift(>=5.9)` guards are removed. The `#available(macOS 14.0, *)` check stays, since that is a real runtime check.
- The `@available(tvOS 12.0, *)` annotations on the `TVMonogramView` support and the `@available(iOS 13.0, tvOS 13.0, *)` annotations on the `UIAction` / `UICommand` conformances are at or below the baseline and no longer restrict anything, so they are removed as well.

The `compiler(>=5.10)`, `swift(>=5.10)`, and `compiler(>=6)` guards are left untouched. Their else branches still compile on a Swift 5.9 toolchain.

No behavior change is intended. The diff is deletions and indentation only.

### Tests

No new tests. The full suite passes on iOS and macOS locally, and tvOS / watchOS / visionOS still build.

### DIFF
--- a/Sources/Extensions/CPListItem+Kingfisher.swift
+++ b/Sources/Extensions/CPListItem+Kingfisher.swift
@@ -113,20 +113,7 @@ extension KingfisherWrapper where Base: CPListItem {
             with: source,
             imageAccessor: ImagePropertyAccessor(
                 setImage: { listItem, image, _ in
-                    /**
-                     * In iOS SDK 14.0-14.4 the image param was non-`nil`. The SDK changed in 14.5
-                     * to allow `nil`. The compiler version 5.4 was introduced in this same SDK,
-                     * which allows >=14.5 SDK to set a `nil` image. This compile check allows
-                     * newer SDK users to set the image to `nil`, while still allowing older SDK
-                     * users to compile the framework.
-                     */
-                    #if compiler(>=5.4)
                     listItem.setImage(image)
-                    #else
-                    if let image = image {
-                        listItem.setImage(image)
-                    }
-                    #endif
                 },
                 getImage: { listItem in
                     listItem.image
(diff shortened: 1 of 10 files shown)


onevclaw (commented):
I reviewed exact head `854c18b9e55d75a26c9dda739ae3429789dac9fa` and found no blocking issues. The removed compiler and availability branches are unreachable under Kingfisher 8’s supported Swift and platform minimums, while the macOS runtime fallback remains intact.

Local validation passed the full test suite, CocoaPods lint across supported platforms, SwiftPM build, visionOS build, and a Catalyst 13.1 type-check probe. I’m leaving this as a non-approving review because the visible CI state is still pending.

onevcat:
LGTM! Thank you for it.

PR #2572, opened 2026-08-29
Avoid repeated full size decodes on an original cache hit

oahhariri:
### Problem

KingfisherManager.deliverOriginalCacheHit runs when a processed cache key
misses but an unprocessed original is on disk. It has two inefficiencies
that compound:

1. It hands the processor a decoded .image. DownsamplingImageProcessor
   works on data, so it re-encodes via image.kf.data(format:) and decodes
   again — three full size passes to produce a thumbnail.
2. It runs once per request. The download path already avoids this:
   ImageDataProcessor.doProcess() memoises by processor identifier across a
   single task's callbacks. The cache hit path had no equivalent.

Reading the original through ImageCache.retrieveImage also promotes the
full size image into the memory cache, where it is evicted again as soon as
the processor has run.

With several views showing the same large image, this is N concurrent full
size decodes plus a retained full size bitmap.

### Fix

- New internal ImageCache.retrieveDataInDiskCache returns the stored bytes
  without deserializing them; the processor receives ImageProcessItem.data.
- OriginalImageProcessingCoalescer shares one read and process between
  requests that would each perform it, keyed on the original cache, the key,
  the processor identifier, the scale factor and the forced extension.
- CacheSerializer.producesDecodableImageData (default `false`) gates the
  data route. Both built-in serializers return true. A serializer that
  transforms what it writes keeps the previous route, uncoalesced.
- backgroundDecode applies to the processed image rather than the original.

### Compatibility

Only the shared read and the processing are shared. The image modifier, the
result value, the target cache store, the callback queue and each request's
own staleness check stay per request, so a single request behaves as before.

The protocol addition has a default implementation, so existing conformers
are unaffected and keep the previous behaviour.

Two behaviour changes worth noting: the original is no longer promoted into
the memory cache on the data route, and a custom processor that returns a
different non-nil image for .data than for .image would see the .data
result — the same input the download path already gives it.

### Tests

Added to KingfisherManagerTests: the processor receives .data on an
original cache hit; concurrent requests for one key process it once and all
receive an image; a serializer that does not declare decodable data still
goes through image(with:options:); and the declared values of the built-in
serializers are pinned.

### DIFF
--- a/Sources/Cache/CacheSerializer.swift
+++ b/Sources/Cache/CacheSerializer.swift
@@ -62,10 +62,23 @@ public protocol CacheSerializer: Sendable {
     /// By default, it is `false`, and the actual processed image is assumed to be serialized to and later deserialized
     /// from the disk. That means the processed version of the image is stored and loaded.
     var originalDataUsed: Bool { get }
+
+    /// Indicates whether ``data(with:original:)`` writes a self-describing image format, such as PNG, JPEG or GIF.
+    ///
+    /// When `true`, Kingfisher may hand the cached bytes straight to an ``ImageProcessor`` as
+    /// ``ImageProcessItem/data(_:)`` instead of deserializing them first. Data-based processors — most notably
+    /// ``DownsamplingImageProcessor`` — re-encode an image input to get back to data, so deserializing first costs a
+    /// full size decode and re-encode that the processor immediately discards.
+    ///
+    /// By default it is `false`, and ``image(with:options:)`` is always used. Return `true` only if the bytes this
+    /// serializer writes can be decoded by the system image APIs on their own; a serializer that encrypts or
+    /// otherwise wraps its payload must leave it `false`.
+    var producesDecodableImageData: Bool { get }
 }
 
 public extension CacheSerializer {
     var originalDataUsed: Bool { false }
+    var producesDecodableImageData: Bool { false }
 }
 
 /// Represents a basic and default `CacheSerializer` used in the Kingfisher disk cache system.
@@ -84,6 +97,9 @@ public extension CacheSerializer {
 /// > Tip: If you create a new image instance from an animated image in a custom processor, use
 /// > ``KingfisherWrapper/copyKingfisherState(to:)`` to propagate the embedded animated data to the new image.
 public struct DefaultCacheSerializer: CacheSerializer {
+
+    /// Writes PNG, JPEG or GIF, so the cached bytes can be decoded directly.
+    public var producesDecodableImageData: Bool { true }
     
     /// The default general cache serializer utilized throughout Kingfisher's caching mechanism.
     public static let `default` = DefaultCacheSerializer()
--- a/Sources/Cache/FormatIndicatedCacheSerializer.swift
+++ b/Sources/Cache/FormatIndicatedCacheSerializer.swift
@@ -60,6 +60,9 @@ import UIKit
 /// imageView.kf.setImage(with: url, options: optionsInfo)
 /// ```
 public struct FormatIndicatedCacheSerializer: CacheSerializer {
+
+    /// Writes the indicated image format, so the cached bytes can be decoded directly.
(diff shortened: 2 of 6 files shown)


onevpaw (changes requested):
The raw-data path addresses the unnecessary full-size deserialization/re-encoding, and the existing test suite passes at `84ad8e899e67a549a5d03bc0100e34eb83109470`. However, I found two blocking correctness issues in the new coalescing logic:

1. `originalProcessingIdentifier` is ambiguous because arbitrary values are joined with `|`. For example, these requests produce the same identifier:
   - cache key `a|b`, processor identifier `c`
   - cache key `a`, processor identifier `b|c`

   A concurrent probe confirmed that the second processor was skipped and its request received the first request's image (`64×64` PNG instead of the expected `365×360` GIF). Could this use a structured `Hashable` key instead, preserving distinctions such as `forcedExtension == nil` versus an empty string?

2. Coalesced participants receive processing performed with the leader's options, although several output-affecting inputs are absent from the key. In a concurrent GIF probe, a full-animation request joined a leader using `.onlyLoadFirstFrame` and incorrectly received a first-frame-only result. The same concern applies to `preloadAllAnimationData`, `backgroundDecode`, and serializer fallback when processors reject `.data`.

Please either include every input that affects the shared result in the processing identity, or narrow coalescing to the disk-byte read/pure processor stage and apply image creation, background decoding, serializer fallback, modifiers, and cache storage per participant.

Regression coverage should include delimiter-bearing cache/processor identifiers, differing animated-image creation options, differing `backgroundDecode` values, and per-request serializer fallback behavior.

oahhariri:
@onevpaw 

Thanks, both were real.

The identifier is a proper Hashable struct now instead of a joined string, and forcedExtension stays String? so nil and "" do not collapse into each other.

For the second one I narrowed rather than grew the key. Coalescing is only the disk read and processor.process(.data:) now. Background decode, the serializer fallback, the modifier and the cache store all happen per request, the same way doProcess() already applies backgroundDecode per callback. The creating options the processor consumes (preloadAllAnimationData, onlyLoadFirstFrame, scaleFactor) went into the identity, since most processors put .data through DefaultImageProcessor and you would still get the first frame only result otherwise.

Tests for the four cases you listed plus a few more around the data path. Suite is green, 415

I also have tested on a device, In an app with a 13506x11208 PNG shown as a thumbnail in a grid, the crash reproduces reliably, and flipping a single boolean turns it on and off: with the app's custom serializer declaring producesDecodableImageData false the app is jetsammed, with it true it survives. Same build, same device, same steps. Worth noting what that means for reach: DefaultCacheSerializer and FormatIndicatedCacheSerializer both declare true so they get this automatically, but a custom serializer defaults to false and keeps the previous path, deliberately, since we cannot know its bytes are decodable. So apps with a custom serializer need to opt in before the fix does anything for them.

Going through the rest of the options against the shared stage turned up a few other things:

- fromMemoryCacheOrRefresh was deciding the shared read for everyone, so those do not coalesce anymore
- the two access extending expiration options are in the identity now, they write persistent disk metadata and a joiner was extending a TTL under someone else's policy. That needed Hashable on StorageExpiration and ExpirationExtending, which is additive, but say if you would rather I fold a discriminator in and leave your enums alone
- scale is kept as a bit pattern, otherwise a non finite value makes an identity unequal to itself
- the original is not background decoded on the read paths anymore since it gets dropped right after the processor runs

The shared process(item:options:) does still see the leader's options, but I think that one sits with the processor rather than with coalescing. The cache is keyed on cacheKey plus processor.identifier, so a processor that varies its output on something outside its identifier already collides there, and doProcess() computes once per identifier using the first callback's options as well. The creating options were the genuine exception, which is why those went into the identity. Can still restrict the shared call to only the identity fields if you would prefer it

onevcat:
@oahhariri Nice catch! Some further work is done in #2573 and we will continue the checks there.

Thank you for the fix and contribution!

PR #2574, opened 2026-09-08
Downsample animated frames instead of decoding every frame at native size

omarH2007:
## Problem

`CGImageFrameSource.frame(at:maxSize:)` passed the thumbnail keys to `CGImageSourceCreateImageAtIndex`, which ignores them. Since b20bac43, every animated frame has been decoded at full size, whatever the size of the view. `KingfisherWrapper.downsampledImage(data:to:scale:)` already uses `CGImageSourceCreateThumbnailAtIndex` correctly.

`AnimatedImageView` keeps `framePreloadCount + 1` decoded frames per view. A 1080x1080 GIF shown in a 90pt tile at 3x takes 4.67 MB per frame instead of about 0.29 MB. In a macOS measurement with a 1080x1920 GIF and 2 buffered frames, memory grew by 28.7 MB per view before this change and 9.0 MB after.

With several animated views on screen this is enough to exhaust memory on 3 GB devices. Crash reports on iOS 18.7 show `EXC_BAD_ACCESS` (null write in `_platform_memmove`) in `GIFReadPlugin::copyImageBlockSetImp`, under `CGContextDrawImage` in `-[UIImage imageByPreparingForDisplay]`, on the animator's serial preload queue. That is consistent with an allocation failure while a full size frame is decoded. This change makes it much less likely by keeping far fewer pixels per view. It cannot remove it completely: ImageIO still decodes a full size canvas for each frame before scaling it down.

The rest of this change covers the review of the first version: frames that stay too small when the view grows, content mode, non-square sizes, views without bounds, a replaced animator that keeps decoding, and macOS.

## Summary
- decode frames with `CGImageSourceCreateThumbnailAtIndex`
- fit frames in both dimensions of `maxSize`. A `maxSize` with no area, or one the frame already fits in, keeps the `CGImageSourceCreateImageAtIndex` path
- size frames for the content mode: fill modes cover the bounds, fit modes fit in them, modes that do not scale keep full size frames
- decode the buffered frames again when the view grows, or its content mode needs more pixels. The animator is not rebuilt, so playback position and stopped state are kept. Shrinking decodes nothing
- do not decode frames while the view has no area, and stop a replaced animator from decoding
- use the window backing scale on macOS. `recommendedLayerContentsScale` returns 1 for an image created from data, even in a 2x window

No public API change. `needsPrescaling = false` keeps full size frames and does not react to layout.

Behavior note: with the default `needsPrescaling = true`, frames are now decoded for the view's bounds. A transform or scroll view zoom does not change the bounds, so magnified views should set `needsPrescaling = false`. Custom `ImageFrameSource` conformers now receive a content mode size, or `nil` for modes that do not scale.

## Tests
- `CGImageFrameSource`: fits square and non-square sizes, keeps full size for `(0, 1)`, `(40, 0)` and sizes larger than the image, downsamples every frame
- frame sizing for fill, fit and non-scaling content modes, empty bounds, `needsPrescaling = false`, and growth only
- animator: exact frame size for the target, aspect fill of a wide image, growth from 40x40 to 60x60, reload around the current frame, purged frames stay purged, frames set up for a new size are not decoded twice, the current frame is displayed again after a reload, cancellation
- view: layout grows frames without a new animator, content mode change, `needsPrescaling = false`, a stopped animation stays stopped, replacing the image cancels the old animator, a subclass skipping `super.layoutSubviews()`, macOS backing scale

Each new assertion was checked by breaking the code it covers (for example fitting instead of filling, skipping the reload, not cancelling in `reset()`, reverting the macOS scale), and the matching test fails.

## Verification
- `xcodebuild test -project Kingfisher.xcodeproj -scheme Kingfisher -destination 'platform=macOS'`: 397 tests passed
- `xcodebuild test -project Kingfisher.xcodeproj -scheme Kingfisher -destination 'platform=iOS Simulator,name=iPhone 17 Pro,OS=26.4'`: 438 tests passed
- same on iOS 18.1 simulator: 438 tests passed
- `xcodebuild build -project Kingfisher.xcodeproj -target KingfisherTests -sdk appletvsimulator26.4` and `-sdk xrsimulator26.4`: build succeeded. The Kingfisher target builds for `watchsimulator26.4`. No tvOS or visionOS runtime was available to run the tests
- thumbnail frames compared pixel by pixel with fully composited `CGImageSourceCreateImageAtIndex` frames scaled to the same size, on 7 real GIFs and on generated GIFs with partial frames and disposal: identical


onevclaw (changes requested):
The core ImageIO change is sound: using `CGImageSourceCreateThumbnailAtIndex` correctly downsamples all frames. However, I found two blocking issues at the current head (`7b4a69e`):

1. **An obsolete zero-bounds animator can continue decoding full-size frames after layout.** `reset()` drops the stored animator, but an already-running `setupAnimatedFrames()` retains the old instance and has no cancellation or generation check. Because both animators share `preloadQueue`, the old animator can finish up to 11 unbounded decodes before the bounded one starts. A deterministic probe observed `[nil, nil, (120, 120)]`, including another unbounded request after layout. Please preferably defer preloading while either bounds dimension is invalid; cancellation checks before each subsequent `loadFrame` would also prevent obsolete work.

2. **The added tests do not compile for tvOS.** `AnimatedImageView(frame: .zero)` is unavailable there because the class declares its own designated initializers. Using the existing `AnimatedImageView()` in the tvOS test fixes the compilation failure, after which the five targeted tests pass.

There are also compatibility cases that should be addressed or explicitly clarified:

- Treat partially zero bounds such as `(0, 1)` as unbounded (`bounds.isEmpty`, or equivalent). Otherwise the animator can permanently retain frames downsampled to only a few pixels.
- When `needsPrescaling == false`, bounds changes should not rebuild the animator or repeat full-size preloading if the intent is to preserve the previous behavior.
- `ImageFrameSource` documents `maxSize` as a bounding box, but passing only `max(width, height)` to ImageIO violates that contract for non-square sizes—for example, `(100, 20)` produces approximately `100×99`. Please either calculate a limit satisfying both dimensions or revise the API contract.

Please add deterministic coverage for the obsolete-animator race, partially zero bounds, disabled prescaling, non-square `maxSize`, and tvOS compilation. The thumbnail path itself passed targeted iOS, macOS, and visionOS tests, so the overall direction looks good once these lifecycle and platform issues are resolved.

onevcat:
One more case to cover: after frames are downsampled, growing the view from 40×40 to 60×60 keeps the smaller frames and can make the animation blurry. The decode size should also account for content mode: fitting a wide image inside the bounds can leave too few pixels for aspect fill. Could you add tests for both cases and update the frames when their resolution is too small for display?

omarH2007:
@onevclaw @onevcat thanks, all of these were real. Summary of what changed:

**1. Obsolete animator.** Both routes. While the view has no area, the animator keeps its timeline but decodes nothing. It no longer rebuilds on layout, it decodes the buffered frames again in place. `reset()` cancels the old animator, and every preload step checks it, so a replaced animator stops after the frame in flight. Tests: `testAnimatorDoesNotDecodeFramesUntilTargetHasArea`, `testCancelledAnimatorStopsDecodingFrames`, `testReplacingImageCancelsPreviousAnimator`.

**2. tvOS.** View tests use `AnimatedImageView()`. `KingfisherTests` builds for `appletvsimulator26.4` and `xrsimulator26.4`.

**Partially zero bounds.** `(0, 1)` is treated as no area. In the frame source, a `maxSize` with a side below 1 keeps full size frames: 365x360 instead of 3x3.

**`needsPrescaling = false`.** Full size frames, no reload and no new animator on layout. `testAnimatorWithoutPrescalingDecodesOriginalFrames`, `testAnimatedImageViewWithoutPrescalingIgnoresLayout`.

**Non-square `maxSize`.** The contract is kept. The max pixel size is computed from the frame's pixel size so both edges fit: `(100, 20)` gives 20x20 instead of 100x99.

**Growth (40x40 to 60x60).** Frames go from 41x40 to 61x60 without restarting playback. Shrinking or growing by less than a pixel decodes nothing. `testAnimatorDecodesLargerFramesWhenTargetGrows`, `testAnimatorReloadsFramesFromCurrentFrame`, `testAnimatedImageViewFollowsLayoutWithoutRebuildingAnimator`.

**Content mode.** The animator passes the size the image is drawn at: covering the bounds for fill modes, fitting for fit modes, and `nil` for modes that do not scale. `ImageFrameSource` keeps its bounding box contract. An 800x200 GIF with aspect fill in 120x120 px decodes to 480x120, where it was 120x30. `testAnimatorDecodesEnoughPixelsToFillTarget`, `testFrameSizingFollowsContentMode`.

Other things I checked:
- modes that do not scale (`.center`, `.scaleNone`, ...) would draw downsampled frames smaller, so they keep full size frames
- on macOS, `recommendedLayerContentsScale` returns 1 for an `NSImage` created from data, even in a 2x window, so frames now use the window backing scale. Size changes come from `setFrameSize`, `viewDidMoveToWindow` and `viewDidChangeBackingProperties`
- a stopped or finished animation displays the current frame again after its reload, otherwise it keeps the smaller frame
- `determineContentsGravity` decides on the image size, so smaller frames do not switch `scaleProportionallyDown` to center
- thumbnail frames match fully composited frames scaled to the same size, on 7 real GIFs and on generated GIFs with partial frames and disposal

Validation: macOS 397, iOS 26.4 simulator 438, iOS 18.1 simulator 438 tests passed. Each new assertion was checked by breaking the code it covers.

Not covered: transforms and scroll view zoom do not change the bounds. The `needsPrescaling` doc now says to set it to `false` for those views. Happy to change any of the choices above.

PR #2576, opened 2026-09-12
Add a public DownloadTask initializer for custom downloaders

Ceylo:
## What

`ImageDownloader.downloadImage(with:options:completionHandler:)` is `open`, and since Kingfisher 8 it returns a non-optional `DownloadTask`. Every `DownloadTask` initializer is internal, so an override outside the module that loads the image itself can only return the task from a `super` call, which starts Kingfisher's own request. This PR adds `DownloadTask.init(cancelling:)`. It is public, wraps the existing `init(providerTask:)`, and returns a task that is `isInitialized`. Calling `cancel()` on it cancels the `Task` that does the work. After review, `KingfisherError.isTaskCancelled` also matches `.asyncTaskContextCancelled`, the reason such a downloader reports when it is cancelled, so `KingfisherManager` doesn't move on to an alternative source after a cancel.

## Motivation

A custom `ImageDownloader` is the one place to change how *every* `kf.setImage(with: url)` loads, without touching each call site. #2107 added a public `ImageDownloadResult` init for this reason (bundled images under a custom URL scheme). That PR's override returned `nil` for its own loads. In v8 that no longer compiles, and there is no public task to return in its place.

My case is similar. I want image requests to use the same HTTP client and connection pool as the rest of the app's traffic. Request modifiers, `sessionConfiguration` and the session delegate all adjust Kingfisher's own `URLSession` request. None of them can replace the transport.

A different shape is fine with me, for example a public `init()` or another name. I kept the change small so it's easy to reshape.

## Before / after

There is no visual change, so the before/after is a compile check and the tests below.

**Before** (`master` at 0b5a9e49): a minimal SwiftPM package depends on Kingfisher by `path:` (no `@testable`) and contains this subclass:

```swift
final class TaskBackedImageDownloader: ImageDownloader, @unchecked Sendable {
    override func downloadImage(
        with url: URL,
        completionHandler: (@Sendable (Result<ImageLoadingResult, KingfisherError>) -> Void)? = nil
    ) -> DownloadTask {
        let callbackQueue = options.callbackQueue
        let work = Task {
            let result: Result<ImageLoadingResult, KingfisherError>
            do {
                let (data, _) = try await URLSession.shared.data(from: url)
                if let image = KFCrossPlatformImage(data: data) {
                    result = .success(ImageLoadingResult(image: image, url: url, originalData: data))
                } else {
                    result = .failure(.processorError(reason: .processingFailed(processor: DefaultImageProcessor.default, item: .data(data))))
                }
            } catch {
                result = .failure(.requestError(reason: .asyncTaskContextCancelled))
            }
            callbackQueue.execute { completionHandler?(result) }
        }
        return DownloadTask(cancelling: work)
    }
}
```

<details><summary><code>swift build</code> against <code>master</code>: fails</summary>

```
Sources/Consumer/TaskBackedImageDownloader.swift:25:16: error: 'DownloadTask' initializer is inaccessible due to 'internal' protection level
23 |             callbackQueue.execute { completionHandler?(result) }
24 |         }
25 |         return DownloadTask(cancelling: work)
   |                `- error: 'DownloadTask' initializer is inaccessible due to 'internal' protection level
26 |     }
27 | }

Sources/Networking/ImageDownloader.swift:84:5: note: 'init(providerTask:)' declared here
```
</details>

<details><summary><code>swift build</code> against this branch: succeeds</summary>

```
Build complete! (31.15s)
```
</details>

**Tests.** Three tests are added to `KingfisherManagerTests`. They use a private `TaskBackedImageDownloader` that returns `DownloadTask(cancelling:)`:

- `testDownloaderReturningTaskBackedDownloadTask`: `retrieveImage` with `.downloader(…)` returns an `isInitialized` task and delivers the image with cache type `.none`.
- `testCancellingTaskBackedDownloadTaskCancelsWork`: `cancel()` on the task that `retrieveImage` returns cancels the downloader's `Task`, and the retrieval fails.
- `testCancellingTaskBackedDownloadTaskSkipsAlternativeSources`: with `.alternativeSources` configured, a cancel ends in `.asyncTaskContextCancelled` (`isTaskCancelled == true`), and the downloader is never asked for the alternative URL.

On `master` these tests don't compile (`incorrect argument label in call (have 'cancelling:', expected 'providerTask:')`). On this branch all three pass. The third one also failed on the first commit of this PR, before the `isTaskCancelled` change: the retrieval succeeded through the alternative source.

## Side effects

- Public API gains one initializer, which only delegates to `init(providerTask:)`, the one the `ImageDataProvider` path already uses.
- `isTaskCancelled` now returns `true` for `.asyncTaskContextCancelled`. Inside Kingfisher, only the async `retrieveImage` bridge creates that reason, and it resumes the caller directly, so no internal caller of the predicate sees it except through a custom downloader. Outside Kingfisher, code that checks `isTaskCancelled` on an error thrown by the async overloads now gets `true` for that case.
- `.dataProviderCancelled` is still not matched, which already happens on `master`. See the review thread.
- The full existing test suite passes unchanged on every platform listed below.

## Verified

All runs used Xcode 26.6 (17F113) on macOS 26.6.2, with the same `xcodebuild` arguments as the `test` / `build` lanes (`-scheme Kingfisher`, `SWIFT_VERSION=5.0`). I ran `xcodebuild` directly instead of through `bundle exec fastlane`: this machine has no Ruby 3.3.6 (the version `.ruby-version` asks for), only the system 2.6.10.

Results are for the head commit, cca97c3f:

- macOS (`platform=macOS`): 381 tests, 0 failures.
- iOS (`platform=iOS Simulator,name=iPhone 17,OS=26.5`): 420 tests, 0 failures.
- tvOS (`platform=tvOS Simulator,name=Apple TV 4K (3rd generation),OS=26.5`): 418 tests, 0 failures.
- watchOS (`-sdk watchsimulator`, watchOS 26.5 SDK): build succeeded. As in CI, no tests ran on watchOS.
- The out-of-module `swift build` check above ran on macOS only.

Not run: Xcode 26.2 (the other half of the CI matrix), and the fastlane `tests` lane's own destinations (iPhone 16 / Apple TV / OS 18.5).

https://claude.ai/code/session_017GgExh8ZQgnQ5uWrQyn9dR


onevpaw (changes requested):
The public initializer addresses the external downloader construction gap. Review of head `ed690565acecd27522d4a472bddbff9d90c64e98` identified one cancellation integration issue:

**[P2] Cancelling a custom task can start an alternative source and return success.** The new tests report cancellation as `.asyncTaskContextCancelled`, but `KingfisherError.isTaskCancelled` (`Sources/General/KingfisherError.swift:407`) only recognizes `.taskCancelled`. Consequently, `KingfisherManager` bypasses its cancellation exit and starts an alternative source (`Sources/General/KingfisherManager.swift:363–381`). An independent module probe reproduced this twice: cancel the primary task with `.alternativeSources` configured, and retrieval succeeds through the fallback.

Although the classification gap predates this PR, it affects the newly exposed task-backed downloader path. Could you make the cancellation error used by this path recognizable to the Manager—for example, by including `.asyncTaskContextCancelled` in the cancellation predicate—and add a regression test asserting that the alternative source never starts and the final result is a cancellation error? The existing alternative-source cancellation test covers the default downloader; the new custom-task tests do not exercise fallback.

Validation passed on iOS (419 tests) and tvOS (417 tests), with watchOS and visionOS simulator builds also passing. The full macOS suite was not clean; a provider timeout also reproduced on the base revision. CI is currently pending.

Ceylo:
Thanks. I reproduced it and fixed it in c317e85d, with follow-ups in cca97c3f.

**Reproduction.** I added `testCancellingTaskBackedDownloadTaskSkipsAlternativeSources` first, on `ed690565`. It uses the task-backed downloader, holds the primary URL until it is cancelled, and configures `.alternativeSources([.network(alternativeURL)])`. It failed on macOS:

```
error: -[KingfisherTests.KingfisherManagerTests testCancellingTaskBackedDownloadTaskSkipsAlternativeSources] : failed - expected .asyncTaskContextCancelled, got: success(Kingfisher.RetrieveImageResult(image: <NSImage 0xb5b65c640 Size={64, 64} …
```

**Fix (c317e85d).** I went with your suggestion: `KingfisherError.isTaskCancelled` now matches `.asyncTaskContextCancelled` as well as `.taskCancelled`. The `init(cancelling:)` doc comment tells custom downloaders to report that reason when `work` is cancelled. The test asserts that the final error is `.asyncTaskContextCancelled`, that `isTaskCancelled` is true, and that the downloader was only ever asked for the primary URL, so the alternative never started.

**Follow-ups (cca97c3f).**
- The `asyncTaskContextCancelled` documentation now mentions that a downloader built on `DownloadTask(cancelling:)` reports it, including for a load that has already started. Before, it named only the async `retrieveImage` bridge.
- The test downloader now completes on `options.callbackQueue`, as `downloadImage` documents, instead of from its `Task`. The example in the PR description is corrected the same way.

**Side effects of the predicate change.** Inside Kingfisher, the only place that creates `.asyncTaskContextCancelled` is the async bridge (the `Task.isCancelled` check after `retrieveImage` in `KingfisherManager.swift`). It resumes the continuation directly and never reaches `failCurrentSource` or `RetryStrategy`. So within the library, widening `isTaskCancelled` only affects errors reported by custom downloaders like the one in this PR. Outside it, the change is visible: code that checks `isTaskCancelled` on an error thrown by the async overloads now gets `true` for that case.

**A related gap I left alone.** `.dataProviderCancelled` isn't matched either, and it has the same symptom. As a probe (not committed) I ran a `.provider` source cancelled with `.alternativeSources([.provider(…)])`, with the fix applied. It still completed with `success` through the alternative. That gap predates this PR and doesn't involve the custom-downloader path, so I didn't widen the change. I'm happy to add `.dataProviderCancelled` to the predicate here, with its own test, or to open a separate PR if you prefer.

**Verified on cca97c3f** (Xcode 26.6, `SWIFT_VERSION=5.0`, same arguments as the lanes):
- macOS (`platform=macOS`): 381 tests, 0 failures.
- iOS (`platform=iOS Simulator,name=iPhone 17,OS=26.5`): 420 tests, 0 failures.
- tvOS (`platform=tvOS Simulator,name=Apple TV 4K (3rd generation),OS=26.5`): 418 tests, 0 failures.
- watchOS: `-sdk watchsimulator` build succeeded.

The three task-backed tests passed on all three test platforms. The out-of-module example from the description builds against this commit and still fails against `master`. Not run: Xcode 26.2 and visionOS.

On the macOS provider timeout you saw: I didn't hit it. The full macOS suite passed on this machine each time, on `ed690565` (380 tests), `c317e85d` and `cca97c3f` (381 tests), so I can't add anything about it.

onevcat:
LGTM! Thank you for the fixing.

There are some additional documentation improvement can be done. I will follow up in another PR.

PR #2578, opened 2026-09-14
Add FlippingImageProcessor for mirroring images

anandghegde:
### Motivation

Kingfisher has no built-in way to mirror an image. Today you have to write a custom `ImageProcessor` for it, for example to show a front-camera photo mirrored or to flip a directional icon. SDWebImage ships this as `SDImageFlippingTransformer`.

### What this adds

- `FlippingImageProcessor(horizontal:vertical:)`, next to `BlackWhiteProcessor` in `ImageProcessor.swift`. Both parameters default to `false`. Its identifier includes both flags, so each combination gets its own cache key.
- `image.kf.flipped(horizontal:vertical:)` in `ImageDrawing.swift`. It draws through the existing `draw(to:inverting:)` helper, the same way `overlaying` and `addingBorder` do, so it gets the same handling on UIKit, AppKit and watchOS. On UIKit the image scale is kept and the image orientation is applied. If neither direction is set, the method returns the original image without drawing.
- `KFOptionSetter.flip(horizontal:vertical:)` for `KF.Builder` and `KFImage`, placed next to `blackWhite()`.
- An entry in the processor list in the DocC common tasks page, and a "Flip" item in the demo's processor collection.

```swift
imageView.kf.setImage(with: url, options: [.processor(FlippingImageProcessor(horizontal: true))])

KF.url(url)
  .flip(horizontal: true, vertical: true)
  .set(to: imageView)
```

I didn't edit `CHANGELOG.md`, since it looks like it gets written at release time.

### Tests

- `ImageDrawingTests`: builds a 2x2 image with a different color in each quadrant, then checks the pixels after a horizontal, vertical and combined flip. It also checks that flipping with both flags `false` returns the same instance, and that a scale-2 image (4x4 pixels) keeps its point size, scale and pixel size.
- `ImageProcessorTests`: checks that the identifier is unique for each flag combination, that the processor works for both `.image` and `.data` input, and that it does nothing when both flags are `false`.
- On macOS the test image is loaded from PNG data. An `NSImage` built directly from a `CGImage` reports 2x pixels in `kf.size` on a Retina display, which would make the pixel checks depend on the machine.
- As a sanity check, I swapped the scale axes in `flipped`. The three pixel tests failed, as expected.

Commands run (Xcode 26.6):

```
xcodebuild test -workspace Kingfisher.xcworkspace -scheme Kingfisher -destination 'platform=macOS'
  Executed 386 tests, with 0 failures
xcodebuild test -workspace Kingfisher.xcworkspace -scheme Kingfisher -destination 'platform=iOS Simulator,name=iPhone 17'
  Executed 425 tests, with 0 failures
```

I couldn't build for tvOS or watchOS locally because those platform runtimes aren't installed. The new code only uses APIs that `addingBorder` and `overlaying` already use on those platforms.

This PR was written with help from Claude Code. I reviewed the changes.

https://claude.ai/code/session_01QBWFWqjbUQGS1c9LsM2bxX


onevtail (commented):
Review findings for `7d714a15668e659303e5592144c7b7b737711319`: no blocking issues found in the flip implementation, processor identifiers, or builder integration.

Supplemental pixel checks passed for non-square transparent images across all eight orientations and three scales, including repeated flips and processor composition. As a non-blocking improvement, please consider retaining these cases as regression tests; the committed tests primarily exercise square, `.up` inputs.

The investigation reports passing iOS and tvOS test suites. On macOS, one timing-sensitive memory-storage test failed narrowly on the first run; the full suite passed when rerun with the same build artifacts. watchOS and visionOS simulator builds also passed, but runtime behavior was not tested on those platforms.

Remote CI is still pending, so this review does not constitute approval.

anandghegde:
Thanks for the review. I added the suggested regression tests in 8eadfd0d:

- `testImageFlippingNonSquareTransparentImage`: a 3x2 image with transparent pixels, checked pixel by pixel for horizontal, vertical and both flips, plus a double flip round trip.
- `testImageFlippingRespectsOrientationAndScale` (iOS/tvOS/visionOS): all eight `UIImage.Orientation` values at scales 1, 2 and 3. Each result is compared against the same image rendered upright by `UIGraphicsImageRenderer`, then mirrored.

`ImageDrawingTests` and `ImageProcessorTests` pass locally on macOS and on the iOS 26 simulator. I couldn't run tvOS here because the runtime isn't installed.

The double-flip check is skipped on macOS. There, a drawn image comes back as `NSImage(cgImage:size: .zero)`, so drawing it again on a Retina display doubles its pixel size. Every drawing method does this already, so it isn't specific to flipping, and I left it alone in this PR.

onevtail (commented):
Review of `8eadfd0d3e100c26ee0e7b01c98ad49617dfb413`: no blocking defects found in the flip implementation, processor identifiers, or Builder composition. Local iOS and tvOS suites passed; additional probes passed for thin images, partial transparency, orientation/scale combinations, and repeated flips. The macOS suite passed on rerun after three assertions failed in one timing-sensitive disk-expiration test on the initial run.

One non-blocking coverage concern: `ImageDrawingTests.swift` skips the round-trip check unconditionally on macOS. Could you retain that coverage for unaffected configurations and narrow the skip to the Retina condition once it is reproducible? Headless probes did not reproduce the reported size doubling, so Retina behavior remains unverified; the shared image-wrapping path predates this PR.

Remote CI has not been confirmed passing for this head, so this review does not include approval.

anandghegde:
Thanks. Narrowed in e047adc8.

The round trip is now its own test, skipped only when the main screen is Retina:

```swift
func testImageFlippingTwiceRestoresTheOriginalImage() throws {
    #if os(macOS)
    try XCTSkipIf(
        (NSScreen.main?.backingScaleFactor ?? 1) > 1,
        "Redrawing a drawn image doubles its pixel size on a Retina display."
    )
    #endif
    ...
}
```

So it runs on iOS, tvOS and visionOS as before, and on macOS whenever the display is not Retina (including a headless run, where `NSScreen.main` is nil and the check goes ahead). `testImageFlippingNonSquareTransparentImage` no longer has any platform condition, so its pixel assertions run everywhere.

On this machine (Retina) `ImageDrawingTests` is 8 tests, 1 skipped, 0 failures, and the skip reason names the condition rather than the platform.

I did not change the size doubling itself: it comes from the shared `NSImage(cgImage:size: .zero)` wrapping that every drawing method uses, so it predates this PR.

onevcat:
Thanks for adding image flipping and the thorough regression tests!

PR #2579, opened 2026-09-14
Let a custom transport report download progress

Ceylo:
## What

An `ImageDownloader` subclass that overrides `downloadImage(with:options:completionHandler:)` to replace the URLSession transport can't report progress. The caller's progress block is stored in `options.onDataReceived`, and that property is internal, as are `DataReceivingSideEffect` and `ImageLoadingProgressSideEffect`. So with such a downloader, `setImage(progressBlock:)`, `retrieveImage(progressBlock:)` and `KFImage.onProgress` never fire.

This PR adds `KingfisherParsedOptionsInfo.reportDownloadProgress(receivedSize:totalSize:)`. It forwards to the same `ImageLoadingProgressSideEffect`s that the URLSession delegate uses, with the same main-queue hop, the same `onShouldApply` check and the same skip for an unknown length (-1). `onDataReceived` now goes through that path too.

## Motivation

This is the same kind of gap as #2576 (and #2107 before it): `downloadImage` is `open`, but an override outside the module can't do everything the built-in transport does. After #2576 such an override can return a task and be cancelled. It still can't drive a progress indicator. My downloader sends image requests through the app's own HTTP client, and `KFImage.onProgress` stays silent for it.

It fits the custom transport guide in #2577, which covers processing, the callback queue and cancellation. The one sentence this adds there would be something like: "Call `options.reportDownloadProgress(receivedSize:totalSize:)` as data arrives." I kept it out of this PR so the two don't conflict, and I'm happy to add it once #2577 lands, or leave it to you.

No issue asked for this. If you'd prefer another shape (a different name, or a method on `ImageDownloader` rather than on the options), I'm fine with changing it.

## Before / after

There is no visual change, so the before/after is a compile check outside the module and the tests below.

**Before** (`master` at ab1c1de5): a minimal SwiftPM package depends on Kingfisher by `path:` (no `@testable`). Its `ImageDownloader` subclass tries to reach the progress blocks the way `SessionDelegate` does:

```swift
options.onDataReceived?.forEach { sideEffect in
    (sideEffect as? ImageLoadingProgressSideEffect)?.block(received, response.expectedContentLength)
}
```

<details><summary><code>swift build</code> against <code>master</code>: fails</summary>

```
Sources/Consumer/CustomTransportDownloader.swift:17:25: error: 'onDataReceived' is inaccessible due to 'internal' protection level
Sources/Consumer/CustomTransportDownloader.swift:18:37: error: cannot find type 'ImageLoadingProgressSideEffect' in scope
```
</details>

**After**: the same subclass, calling the new method as bytes arrive:

```swift
final class CustomTransportDownloader: ImageDownloader, @unchecked Sendable {
    override func downloadImage(
        with url: URL,
        completionHandler: (@Sendable (Result<ImageLoadingResult, KingfisherError>) -> Void)? = nil
    ) -> DownloadTask {
        let work = Task {
            let (bytes, response) = try await URLSession.shared.bytes(from: url)
            let totalSize = response.expectedContentLength
            var data = Data()
            for try await byte in bytes {
                data.append(byte)
                options.reportDownloadProgress(receivedSize: Int64(data.count), totalSize: totalSize)
            }
            _ = data
        }
        return DownloadTask(cancelling: Task { _ = try? await work.value })
    }
}
```

<details><summary><code>swift build</code> against this branch: succeeds</summary>

```
Build complete! (15.97s)
```
</details>

**Tests.** Three tests are added to `KingfisherManagerTests`. They extend the private `TaskBackedImageDownloader` from #2576 with an optional list of `(receivedSize, totalSize)` pairs, which it reports from its `Task` before it completes:

- `testTaskBackedDownloaderReportsProgress`: the progress block receives the pairs in order, on the main thread, before the completion handler.
- `testTaskBackedDownloaderProgressWithUnknownTotalSizeIsNotReported`: a pair with a total of `-1` is not delivered, and the next one is.
- `testTaskBackedDownloaderProgressSkipsReplacedImageViewTask`: `kf.setImage` for one URL, then for another on the same view in the same main-actor turn. The first request's progress block never fires. The second's does.

On `master` these tests don't compile, because the method doesn't exist. To check that each test covers its own behaviour, I changed the source three ways and put it back each time:

| Mutation | Result |
|---|---|
| `reportDownloadProgress` returns immediately | all three fail, e.g. `XCTAssertEqual failed: ("[]") is not equal to ("[[10, 40], [40, 40]]")` |
| `guard self.onShouldApply()` removed | only the replaced-view test fails: `("[[40, 40]]") is not equal to ("[]")` |
| the `-1` guard removed | only the unknown-size test fails: `("[[10, -1], [20, 40]]") is not equal to ("[[20, 40]]")` |

## Side effects

- `ImageLoadingProgressSideEffect.onDataReceived` now calls a new internal `report(received:total:)`. Both sizes are still read on the main queue, after the `onShouldApply` check, and a missing response still counts as `-1`. The existing URLSession progress tests in `ImageViewExtensionTests` and `ImageDownloaderTests` pass unchanged.
- Only progress blocks are fed. `progressiveJPEG` needs the data chunks, not sizes, so a custom transport still gets no progressive updates. The doc comment says so.
- Blocks are skipped in the same cases as for a URLSession download. That includes an image view extension moving on to another image. `KFImage` doesn't pass a task identifier checker on either path, so this PR changes nothing there.
- Public API gains one method on `KingfisherParsedOptionsInfo`, and the `downloadImage` doc comment gains one line pointing to it.

## Verified

All runs used Xcode 26.6 (17F113) on macOS 26.6.2, with the same `xcodebuild` arguments as the `test` / `build` lanes (`-workspace Kingfisher.xcworkspace -scheme Kingfisher SWIFT_VERSION=5.0`). I ran `xcodebuild` directly instead of through `bundle exec fastlane`: this machine has no Ruby 3.3.6 (the version `.ruby-version` asks for), only the system 2.6.10.

Results are for the head commit, f42506c5:

- macOS (`platform=macOS`): 384 tests, 0 failures.
- iOS (`platform=iOS Simulator,name=iPhone 17,OS=26.5`): 423 tests, 0 failures.
- tvOS (`platform=tvOS Simulator,name=Apple TV 4K (3rd generation),OS=26.5`): 421 tests, 0 failures.
- watchOS (`-sdk watchsimulator`, watchOS 26.5 SDK): build succeeded. As in CI, no tests ran on watchOS.
- The out-of-module `swift build` check above ran on macOS only.

One tvOS run, on an earlier revision of this branch whose only difference is the new doc comment's wording, had one failure: `testRetrieveImageWithAsyncCacheTypeCheckUsesOriginalCacheFallback` timed out after 3 seconds. That test passes no progress block, so it never reaches the changed code. I ran it 20 times on tvOS on both `master` (ab1c1de5) and this branch, and it timed out 2 times out of 20 on each. It looks flaky under load rather than affected by this PR.

Not run: Xcode 26.2 (the other half of the CI matrix), visionOS, and the fastlane `tests` lane's own destinations (iPhone 16 / Apple TV / OS 18.5).

https://claude.ai/code/session_01KRrDryy799JKKkWzKjVxux


onevclaw (commented):
Review findings for f42506c52f3089422005f4f2dfd8e57df465e8f8: no blocking implementation issues found. `reportDownloadProgress` exposes the missing entry point for custom downloaders while keeping internal protocols private and avoiding data-dependent progressive JPEG side effects. The native path preserves main-queue delivery and checks request validity before reading progress values.

An independent probe also confirmed that replacing a request with another request for the same URL from inside its progress callback correctly suppresses queued progress from the old request. Please consider adding this case as a permanent regression test.

Validation passed for all 423 iOS tests, three additional progress probes, the watchOS Simulator build, and an external Swift 6 consumer type-check. The initial macOS suite had two failures involving cache-expiration timing and a subsequent crash; both affected tests passed five reruns each. Those failures occur outside the changed progress path, but without a base comparison, the initial macOS run remains a validation caveat.

The implementation review is complete, but approval remains pending CI results for this head.

onevcat:
Thanks for adding progress reporting for custom downloaders!

Ceylo:
Ah this was still in draft because I still hadn't had time to properly check it myself. But if it's good for you it's good for me too!

PR #2580, opened 2026-09-15
Fix parameter documentation for retrieveImageFromCache

hxperl:
The doc comment on `KingfisherManager.retrieveImageFromCache` describes three parameters the method does not have — `key`, `url`, `options` — and omits the two it does: `context` and `downloadTaskUpdated`. `key` and `options` are derived inside the body (`context.options`, `source.cacheKey`), which is likely how it went unnoticed.

`git blame` puts the doc block at bb5d936 (2023-09-26), and the signature at that same commit already took `context` — so these entries have never matched. `downloadTaskUpdated` arrived in c0f10a0 and was never documented.

Comment-only. The `- Note:` pointed at ``processor(_:)`` "in `options`"; since `options` is no longer a parameter I repointed it at the context's options.

`swift build` succeeds on Swift 6.2 / Xcode 26.0.1. I did **not** run XCTest — SwiftPM has no test target and the suite needs simulators, disproportionate for comment lines. No code paths change.

Two things for you rather than me:

- The method is `internal`, so DocC does not render it; the value is for people reading the source. If you would rather leave internal doc comments alone, say so and I will close this.
- `retrieveImageFromCacheAsync` (`:833`, private) has the same defect — it documents ``onOriginalCacheFallbackDownload`` but the parameter is `fallbackDownloadTaskCreated`. Left out to keep this to one defect; happy to fold it in or send a follow-up.

I found this by parsing every `- Parameter(s)` block under `Sources/` against the parsed signatures. It is the only mismatch in the tree.


onevtail (commented):
The parameter names now match the signature, and the changes are limited to documentation. One minor clarification in `Sources/General/KingfisherManager.swift:742–743`: `downloadTaskUpdated` receives the optional task returned by fallback loading. That fallback can use a data provider, and the callback can receive `nil` while an asynchronous request modifier is pending, so “newly created download task” and “a download is issued” are too specific.

Could you use wording that covers both cases? For example:

```swift
///   - downloadTaskUpdated: Called with the optional task returned by fallback loading when an expected
///   original cache hit cannot serve the image. Fallback loading may download the image or load it from
///   a data provider. It is not called when the cache serves the image.
```

Validation at head `be224e178c748745a6dcc58da22f487da228f389` passed the Swift build, 95 targeted macOS tests, and two temporary probes covering these fallback cases. CI is still pending.

onevclaw (commented):
Reviewed head `437db7f716cf94ce1edc6216844da574d97a08eb`. The corrected parameter documentation matches the method signature and implementation. The fallback callback description also accounts for both network and data-provider paths, including an optional task.

No blocking findings. Only documentation comments changed; the macOS build and all 91 `KingfisherManagerTests` passed in local validation. CI is still pending, so this review does not grant approval.

onevcat:
@hxperl Thank you for it!

